PR #133, opened 2022-09-23
Add download commands for mac and linux

cagiti:
## Type of change
docs

### What should this PR do?
Adds chainctl binary retrieval commands to the guide

Missing information

Readers can download the chainctl binary from commands specified in the guide.

Use the commands added to retrieve the chainctl binary for each platform type

image


ltagliaferri:
Thank you for the PR, the `curl` line was deleted in error. Please file an issue if you notice something awry in the future, thanks!

ltagliaferri:
Fixed with #136

PR #136, opened 2022-09-23
Reinstate curl

ltagliaferri:
Add `curl` command back in that was deleted in error


cagiti:
@ltagliaferri: How come you don't sign your commits?

PR #150, opened 2022-10-12
Add terminal back to melange getting started guide

jamonation:
## Type of change
Bug/Documentation

### What should this PR do?
This PR adds the interactive terminal back to the Getting Started with Melange tutorial.

With the 0.2.2 release of [minicli/curly](https://github.com/minicli/curly/releases/tag/0.2.2) the `hello-minicli:test` image at the end of the tutorial now runs as expected. 

Test the tutorial and look for output like the following:

```
docker run --rm hello-minicli:test


Work is never as important as you think it is.
```

### DIFF
--- a/content/open-source/melange/getting-started-with-melange.md
+++ b/content/open-source/melange/getting-started-with-melange.md
@@ -13,6 +13,7 @@ menu:
     parent: "melange"
 weight: 100
 toc: true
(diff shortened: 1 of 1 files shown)


ltagliaferri (approved):
Hooray, it works 🎉

PR #152, opened 2022-10-13
Fixing Netlify Deploy Preview

erikaheidi:
### What should this PR do?
This PR sets additional parameters to the Netlify configuration file to fix the deploy preview. It was not necessary to change the configuration of the website.

We should check the Netlify preview to see if it was fixed, and make sure the production site keeps working as expected once changes are incorporated.

### DIFF
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -1,4 +1,4 @@
-baseUrl = ""
+baseUrl = "https://edu.chainguard.dev/"
 canonifyURLs = false
 disableAliases = true
 disableHugoGeneratorInject = true
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -1,4 +1,4 @@
-baseurl = "https://edu.chainguard.dev/"
+baseUrl = ""
 canonifyURLs = false
 disableAliases = true
 disableHugoGeneratorInject = true
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -1,4 +1,4 @@
-baseUrl = "https://edu.chainguard.dev/"
+baseURL = "https://edu.chainguard.dev/"
 canonifyURLs = false
 disableAliases = true
 disableHugoGeneratorInject = true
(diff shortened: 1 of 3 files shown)


netlify.toml
```diff
@@ -6,13 +21,13 @@
     X-Content-Type-Options = "nosniff"
     Content-Security-Policy = '''
        default-src 'self';
-       frame-src 'self' https://player.vimeo.com https://www.youtube.com www.youtube-nocookie.com https://platform.twitter.com https://syndication.twitter.com https://visualization-ui.chainguard.app;
-       style-src 'self' 'unsafe-inline' cdn.jsdelivr.net https://fonts.googleapis.com *.netlify.app;
+       frame-src 'self' *.chainguard.dev https://player.vimeo.com https://www.youtube.com www.youtube-nocookie.com https://platform.twitter.com https://syndication.twitter.com https://visualization-ui.chainguard.app;
```
ltagliaferri, line 24: Can we replace wildcard `*` with `edu` throughout?

config/_default/config.toml
```diff
@@ -1,4 +1,4 @@
-baseurl = "https://edu.chainguard.dev/"
+baseUrl = "https://edu.chainguard.dev/"
```
ltagliaferri, line 1:
```suggestion
baseURL = "https://edu.chainguard.dev/"
```

On Hugo config I am seeing this

config/_default/config.toml
```diff
@@ -1,4 +1,4 @@
...
 canonifyURLs = false
```
ltagliaferri, line 2: I am working on analytics related to canonifying URLs, but we may want to revisit documentation here

https://gohugo.io/content-management/urls/#canonicalization

erikaheidi: yeah I thought this should be set to true only for production, however I saw this on the docs: `In the May 2014 release of Hugo v0.11, the default value of canonifyURLs was switched from true to false, which we think is the better default and should continue to be the case going forward`

so I don't know anymore. In any case if we decide to change it I think it should go on the config/production folder

jamonation (approved):
Looks good to me on the preview site, all the assets load now right from netlify.

ltagliaferri (approved):

PR #175, opened 2022-11-04
add documentation for additional Chainguard Enforce Agent install methods

joshrwolf:
### What should this PR do?

Adds documentation for new declarative install features in Enforce Agent.

Address users who wish to install Enforce Agent without `chainctl`.

### DIFF
--- a/content/chainguard/chainguard-enforce/chainguard-enforce-kubernetes/alternative-installation-methods.md
+++ b/content/chainguard/chainguard-enforce/chainguard-enforce-kubernetes/alternative-installation-methods.md
@@ -1,0 +1,192 @@
+---
+title: "Alternative Enforce Agent Installation Methods"
+type: "article"
+lead: ""
+description: "Alternative installation methods and instructions."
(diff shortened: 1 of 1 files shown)


ltagliaferri (commented):
Thank you @joshrwolf — would you be opposed to the team breaking this down into two separate docs (one each for installing with raw YAML and one for installing with Helm), and I think we can also pull out the installation with chainctl as well to be a separate doc for onboarding. We can then link these resources as alternatives within the onboarding guide.

My thinking here is that a user will be looking to install Enforce with a particular approach and will use search to find that approach, and that will keep each of the guides short and easier to parse. We can also have a paragraph at the top explaining that they have a choice of methods with links out.

joshrwolf:
I don't feel strongly either way so I defer to the experts 😅 let me know how I can help!

ltagliaferri (approved):
Thank you @joshrwolf , I'll merge in and we'll make updates from there. This is great!

Issue #178, opened 2022-11-08
Add Enforce OpenAPI Specification to site


jamonation:
There's already an `openapi` shortcode in the template for the site, so this issue is hopefully just a matter of putting an `index.md` file in place and pointing it to the Enforce OpenAPI `api.json` file.

PR #179, opened 2022-11-08
Add Enforce OpenAPI Specification to the site (#178)

jamonation:
## Type of change
Documentation

### What should this PR do?
Add Enforce OpenAPI Specification to the site based on #178 

Preview should show the alpha version of the OpenAPI spec under the path `/chainguard/chainguard-enforce/enforce-api/`


jamonation:
netlify doesn't like the iframe from unpkg.dev, but the shortcode is getting rendered and the element is on the page:

Screenshot 2022-11-08 at 4 31 01 PM

The `api.json` file is also there, so it should work on the live site.

ltagliaferri (approved):

PR #182, opened 2022-11-14
Adding global CSP rules

erikaheidi:
### What should this PR do?
This PR adds CSP rules to protect against XSS attacks.

We had a report from a user that the search box would allow for inline script execution.

Type the following in the search box and you should see a prompt/dialog come up in the current site, while with the changes (check preview site) the inline script execution is blocked by the browser.

```
"><img src=x onerror=prompt(document.domain);>
```

### DIFF
--- a/layouts/partials/head/script-header.html
+++ b/layouts/partials/head/script-header.html
@@ -1,3 +1,16 @@
+<meta http-equiv="Content-Security-Policy"
+      content="
+       default-src 'self';
+       frame-src 'self' edu.chainguard.dev https://player.vimeo.com https://www.youtube.com https://www.youtube-nocookie.com https://platform.twitter.com https://syndication.twitter.com https://visualization-ui.chainguard.app;
+       style-src 'self' edu.chainguard.dev 'unsafe-inline' cdn.jsdelivr.net https://fonts.googleapis.com;
+       form-action 'self';
+       font-src 'self' edu.chainguard.dev https://fonts.googleapis.com https://fonts.gstatic.com https://cdn.jsdelivr.net;
+       script-src 'self' edu.chainguard.dev *.googleapis.com cdn.jsdelivr.net *.googletagmanager.com 'sha256-vOgyKS2vkH4n5TxBJpeh9SgzrE6LVGsAeOAvEST6oCc=';
+       connect-src 'self' edu.google-analytics.com;
+       img-src 'self' edu.chainguard.dev;
+       base-uri 'self';
+     ">
+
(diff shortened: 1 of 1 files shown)


layouts/partials/head/script-header.html
```diff
@@ -1,3 +1,16 @@
+<meta http-equiv="Content-Security-Policy"
+      content="
+       default-src 'self';
+       frame-src 'self' edu.chainguard.dev https://player.vimeo.com https://www.youtube.com https://www.youtube-nocookie.com https://platform.twitter.com https://syndication.twitter.com https://visualization-ui.chainguard.app;
+       style-src 'self' edu.chainguard.dev 'unsafe-inline' cdn.jsdelivr.net https://fonts.googleapis.com;
+       form-action 'self';
+       font-src 'self' edu.chainguard.dev https://fonts.googleapis.com https://fonts.gstatic.com https://cdn.jsdelivr.net;
+       script-src 'self' edu.chainguard.dev *.googleapis.com cdn.jsdelivr.net *.googletagmanager.com 'sha256-vOgyKS2vkH4n5TxBJpeh9SgzrE6LVGsAeOAvEST6oCc=';
```
jamonation, line 8: Curious what this `sha256-vOgyKS2vkH4n5TxBJpeh9SgzrE6LVGsAeOAvEST6oCc=` is - does it need to be changed on any deploy where the JS files change, or is it that CSP nonce thing?

jamonation (commented):
Something is generating some inline `data:image/svg+xml` element in the site.

I spotted this in Firefox's console with the new policies in place:

```
Content Security Policy: The page’s settings blocked the loading of a resource at data:image/svg+xml,%3csvg xmlns='http://… (“img-src”).
```

But I can't figure out what is making those images. Is there a CSP policy to allow those somehow?

layouts/partials/head/script-header.html
```diff
@@ -1,3 +1,16 @@
...
+       connect-src 'self' edu.google-analytics.com;
+       img-src 'self' edu.chainguard.dev;
```
jamonation, line 10: Adding `data:` at the end of the `img-src` line seems to resolve the `data:image/svg...` error on the console.

ltagliaferri: Can you check this out and push that into the PR @jamonation ?

jamonation: Sure, I've added and pushed to the PR branch @ltagliaferri

ltagliaferri (approved):

PR #183, opened 2022-11-14
Update frame-src CSP to allow terminal.inky.wtf

jamonation:
This PR adds `https://terminal.inky.wtf` to the `frame-src` portion of the Content Security Policy.


ltagliaferri (approved):
looks good, thank you

PR #185, opened 2022-11-16
Add unpkg.com to style-src and script-src CSPs

jamonation:
The [OpenAPI page](https://edu.chainguard.dev/chainguard/chainguard-enforce/chainguard-enforce-kubernetes/api/) doesn't load because it calls JS/CSS from unpkg.com. This PR adds them, along with what I think is the SHA hash of the generated inline javascript for the page.


jamonation:
Checks all passed, going to merge this to get the page fixed.

Issue #163, opened 2022-10-17
Migrate site deployment from from Google App Engine to Cloud Run


jamonation:
Deploying the site with App Engine is a bit lacking in terms of control over included files and runtime environment. Migrating the site to Cloud Run will enable running in a signed nginx container like [cgr.dev/chainguard/nginx:latest](https://github.com/chainguard-images/nginx).

Using nginx will enable things like path based authentication, manipulation of HTTP headers, and even proxying to other services if we ever need that functionality.

There is a [Cloud Run GitHub Action](https://github.com/marketplace/actions/deploy-to-cloud-run) that supports workload identity federation, so the final deploy pipeline should be pretty similar to what we're currently using.

This isn't a priority, but is definitely a nice to have improvement.

PR #191, opened 2022-11-21
Create cloud-run.yaml

jamonation:
Add a `cloud-run.yaml` file to deploy to Cloud Run using GitHub Actions

### What should this PR do?
Fixes: #163 

This PR adds a GitHub action that builds an nginx container and deploys it to Cloud Run.

New content or code should be visible at https://academy-aze2ubqi3q-uc.a.run.app

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM cgr.dev/chainguard/nginx:1
+FROM cgr.dev/chainguard/nginx:latest
 
 COPY public/ /var/lib/nginx/html/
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM cgr.dev/chainguard/nginx:1
```
ltagliaferri, line 1: Just a question about the `:1` and if there is a concern about using `:latest` instead?

ltagliaferri (approved):
LGTM

PR #202, opened 2022-12-02
Fixing broken links in alternative installation guide

SharpRake:
## Type of change

This change is just fixing some broken links in the Alternative Enforce Agent Installation Methods guide.

I was clicking around on this doc and noticed that the internal links weren't entered correctly. I also made a few minor tweaks to the copy that I thought made sense.


ltagliaferri (approved):
Thanks, I'm going to merge it

PR #226, opened 2022-12-16
Add one-liner showing how to verify chainctl binary with cosign

jamonation:
## Type of change
Documentation

### What should this PR do?
This adds a one line command users can run to verify their `chainctl` binary with Cosign.

Run the command in the PR to verify a recent version of `chainctl`.


ltagliaferri (approved):
Looks good, I like the abstracting away of the `chainctl version` output too 👍

PR #235, opened 2022-12-21
Added workflow for autodocs - images reference

erikaheidi:
This PR creates a workflow to run autodocs via manual trigger, from the GitHub Actions tab, so whenever there's an update to the source images READMEs, a new PR is created here bringing the changes.

We can later on add a schedule to run this periodically. To try it out first, I would prefer to run manually, that's why I didn't add the schedule yet.

Example of PR that will be created: https://github.com/chainguard-dev/deved-autodocs/pull/3

### DIFF
--- a/.github/workflows/autodocs-images.yaml
+++ b/.github/workflows/autodocs-images.yaml
@@ -47,8 +47,3 @@ jobs:
             images
             automated
           assignees: erikaheidi
-          reviewers: |
-            erikaheidi
-            jamonation
-
-
--- a/.github/workflows/autodocs-images.yaml
+++ b/.github/workflows/autodocs-images.yaml
@@ -13,7 +13,7 @@ jobs:
           path: edu
 
       - name: "Check out Images Monorepo"
-        uses: actions/checkout@v2
+        uses: actions/checkout@@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
         with:
           repository: chainguard-images/images
           path: images
--- a/.github/workflows/autodocs-images.yaml
+++ b/.github/workflows/autodocs-images.yaml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Check out Destination Repo"
-        uses: actions/checkout@v2
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
         with:
           path: edu
 
--- a/.github/workflows/autodocs-images.yaml
+++ b/.github/workflows/autodocs-images.yaml
@@ -22,7 +22,7 @@ jobs:
         run: mkdir -p "${{ github.workspace }}/${{ env.WORKDIR }}" && chmod 777 "${{ github.workspace }}/${{ env.WORKDIR }}"
 
       - name: "Update the reference docs for Chainguard Images"
-        uses: chainguard-dev/deved-autodocs@1.0.2
+        uses: chainguard-dev/deved-autodocs@84badec14541d344c47d1a8435a8a214e72dbe95 # v1.0.2
         with:
           command: build images
         env:
--- a/.github/workflows/autodocs-images.yaml
+++ b/.github/workflows/autodocs-images.yaml
@@ -1,0 +1,54 @@
+name: Build Images Reference Docs
+on:
+  workflow_dispatch:
+env:
+  WORKDIR: images
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Check out Destination Repo"
+        uses: actions/checkout@v2
+        with:
+          path: edu
+
+      - name: "Check out Images Monorepo"
+        uses: actions/checkout@v2
+        with:
+          repository: chainguard-images/images
+          path: images
+
+      - name: "Set up workdir"
+        run: mkdir -p "${{ github.workspace }}/${{ env.WORKDIR }}" && chmod 777 "${{ github.workspace }}/${{ env.WORKDIR }}"
+
+      - name: "Update the reference docs for Chainguard Images"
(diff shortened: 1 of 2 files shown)


.github/workflows/autodocs-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  workflow_dispatch:
+env:
+  WORKDIR: images
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Check out Destination Repo"
+        uses: actions/checkout@v2
```
jamonation, line 11:
```suggestion
        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
```

jamonation: I think there's a preference for using commit hashes with actions, since tags can change invisibly from underneath us. Any reason to use v2 over v3 of this action?

.github/workflows/autodocs-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          path: edu
+
+      - name: "Check out Images Monorepo"
+        uses: actions/checkout@v2
```
jamonation, line 16:
```suggestion
        uses: actions/checkout@@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
```

.github/workflows/autodocs-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+          commit-message: Update Images Reference
+          title: "[AutoDocs] Update Images Reference Docs"
+          signoff: true
+          labels: |
+            documentation
+            images
+            automated
+          assignees: erikaheidi
+          reviewers: |
+            erikaheidi
```
ltagliaferri, line 51: Did you want to add the code owners so this could be a variable?

ltagliaferri (approved):

.github/workflows/autodocs-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - name: "Check out Images Monorepo"
+        uses: actions/checkout@v2
+        with:
+          repository: chainguard-images/images
+          path: images
+
+      - name: "Set up workdir"
+        run: mkdir -p "${{ github.workspace }}/${{ env.WORKDIR }}" && chmod 777 "${{ github.workspace }}/${{ env.WORKDIR }}"
+
+      - name: "Update the reference docs for Chainguard Images"
```
jamonation, line 24: More curious about this than anything, but could these files get generated in the `deved-autodocs` repo with a GitHub action and uploaded to a cloud storage bucket there, then downloaded as part of this workflow here?

.github/workflows/autodocs-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+        uses: chainguard-dev/deved-autodocs@1.0.2
```
jamonation, line 25:
```suggestion
        uses: chainguard-dev/deved-autodocs@84badec14541d344c47d1a8435a8a214e72dbe95 # v1.0.2
```

.github/workflows/autodocs-images.yaml
```diff
@@ -0,0 +1,54 @@
...
+          YAMLDOCS_OUTPUT: "${{ github.workspace }}/${{ env.WORKDIR }}"
+
+      - name: "Copy updates to main repo"
+        run: |
+          echo "Copying files..." && \
+          cp -R "${{ github.workspace }}/${{ env.WORKDIR }}" "${{ github.workspace }}/edu/content/chainguard/chainguard-images" && \
+          echo "Finished copy"
+
+      - name: Create a PR
+        uses: peter-evans/create-pull-request@v4
```
jamonation, line 39:
```suggestion
        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
```

jamonation: This action looks like magic! Does it need a github token anywhere to create the PR or does it just figure it out for public repos?

jamonation (approved):
LGTM! A couple suggestions for versioning things, but they're mostly cosmetic changes.

.github/workflows/autodocs-images.yaml
erikaheidi: Ah, thanks! I  have been following mostly the examples they have and I didn't know about using hashes, makes sense. And no reason to use v2, it was just because I copied from another action I had :) thanks!

.github/workflows/autodocs-images.yaml
erikaheidi: I will update the version after merging these.

.github/workflows/autodocs-images.yaml
erikaheidi: there are some permission challenges for generating the files inside the repo directly, because of the user running in the container image (it usesone of our distroless php images). Uploading would be possible in a separate step since the files are shared

.github/workflows/autodocs-images.yaml
erikaheidi: You can set the token, but by default it uses the credentials from the GitHub bot, that's why it shows up as PR author even thou the commit comes from me. One issue we'll have is with the Enforce app, because it won't be gitsigned, but I think there is an action for gitsign that we could use.

.github/workflows/autodocs-images.yaml
erikaheidi: Good point, let me try adding that too.

erikaheidi: Ok, added the CODEOWNERS file, and removed the reviewers portion since that is not necessary. When this is merged, all PRs will automatically tag @chainguard-dev/deved-team for a review. To enforce the review, I think we'll need to set up branch protection thou.

PR #248, opened 2023-01-04
Change menu link

ltagliaferri:
I updated the menu bar so that it points to "Chainguard" and the www site rather than Academy. Do we think this looks ok following discussions from yesterday cc @sheesh @SharpRake ?

Also updated the year in the footer

Signed-off-by: ltagliaferri


SharpRake (approved):
I think this works overall. I still find myself expecting to go back to the Academy home page when I click the link at the top, but I don't think it's confusing or unnatural that it goes to www now. Ultimately, I think a dual-nav type of thing would be the best of both worlds but this works for now.

PR #272, opened 2023-01-13
Set up OCI directory

sheesh:


content/open-source/oci/what-is-the-oci.md
```diff
@@ -0,0 +1,111 @@
+---
+title: "What is the Open Container Initiative?"
+type: "article"
+description: " ... (fill out)"
```
erikaheidi, line 4: I think you forgot to set the description here! this shows up in the list page of articles under the "open container" menu.

config/_default/menus/menus.en.toml
```diff
@@ -69,11 +69,17 @@
   parent = "menu-oss"
   weight = 30
 
+[[main]]
+  name = "Open Containers"
+  url = "/open-source/oci/"
+  parent = "menu-oss"
+  weight = 35  
+
```
erikaheidi, line 77: I have mixed feelings about this being in the top-level menu. I see the "sbom" article was added here too, but it feels like they should be together in a section that has more conceptual articles grouped together. Do you plan on having more OCI articles?

The way it looks right now, "Open Container Initiative" is being listed on the left menu together with our open source projects, so it can be misleading. You can see what I mean here: https://deploy-preview-272--ornate-narwhal-088216.netlify.app/open-source/oci/what-is-the-oci/

erikaheidi (changes requested):

erikaheidi:
hey Erin! The article itself is great, there's only the description to fill in. The thing I am a bit unsure about is the menu entry, because the topic "OCI" is not one of our open source projects and may not get additional articles in the future. What do you think? I wonder if we could come up with a section / menu to group more conceptual articles together so there is a clear separation between our open source projects and other topics.

PR #276, opened 2023-01-17
Add terminal to how-to-use-chainguard-images page

jamonation:
### What should this PR do?
This PR adds the terminal to the [How To Use Chainguard Images](https://edu.chainguard.dev/chainguard/chainguard-images/how-to-use-chainguard-images/) page. It reworks the text a little bit because the `Dockerfile` syntax to include heredocs is not widely supported across most of the older versions of Docker out there.

Adding the terminal to images and showing how they work in the browser is a nice easy way to introduce users to building with wolfi.

The netlify preview won't work, but tutorial should work in production.

Test the steps using a terminal directly: https://terminal.inky.wtf/?image=academy/chainguard-images:latest

### DIFF
--- a/content/chainguard/chainguard-images/how-to-use-chainguard-images.md
+++ b/content/chainguard/chainguard-images/how-to-use-chainguard-images.md
@@ -11,6 +11,7 @@ menu:
   docs:
     parent: "chainguard-images"
 weight: 300
+terminalImage: academy/chainguard-images:latest
 toc: true
 ---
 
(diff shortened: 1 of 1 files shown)


jamonation:
@erikaheidi when you have a moment could you take a look at this PR? I think splitting out there heredoc is a good idea, and doing it that way enables adding the interactive terminal. Thanks in advance! 🙇

ltagliaferri (approved):

erikaheidi (approved):

erikaheidi:
Looks good! I think this page has been edited by someone in the images team, it doesn't look like the initial example I had there, so I wasn't aware of the heredoc thing. I noticed a couple links that need updating but I will create a separate issue for that. This one is good to merge 👍🏼

jamonation:
Thanks for the review @erikaheidi! - I see another example that uses the heredoc format in https://github.com/chainguard-images/.github/blob/main/profile/README.md so I'll update the instructions there too.

PR #295, opened 2023-01-20
Update images reference workflow - fixes #274

### LINKED ISSUE #274: Fix README import / workflow  after SVG issue with Academy theme
There is an issue with how the theme renders images, it calls a .Resize method that doesn't work for SVG files. With the latest update on image READMEs, which includes SVG badges, the entire site breaks as Hugo is unable to build. Needs to overwrite the layout that renders images and check if the image is an svg to render it differently.

erikaheidi:
Updating the images reference action version. This update fixes #274 

The new overview page has a table exhibiting the image tags with information obtained by a JSON file. Example: https://github.com/chainguard-dev/deved-autodocs/pull/10/files#diff-0b359d83768563ba64efc5d825da82b824c1860372d9e4465a3cbb19b0ad3e00

I am changing this workflow to use a tag instead of a commit because this Github Action comes from a trusted source: it's repository we own, inside our org, it has CODEOWNERS in place, and I'll only use tagged releases for it. Using a specific commit hash is a lot harder specially because this action is under active development.

I will still need to update this workflow when there's a new version, but it will be easier to update.

### DIFF
--- a/.github/workflows/autodocs-images.yaml
+++ b/.github/workflows/autodocs-images.yaml
@@ -25,7 +25,7 @@ jobs:
         run: mkdir -p "${{ github.workspace }}/${{ env.WORKDIR }}" && chmod 777 "${{ github.workspace }}/${{ env.WORKDIR }}"
 
       - name: "Update the reference docs for Chainguard Images"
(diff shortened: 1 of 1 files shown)


jamonation (approved):
Makes sense to me, I trust us 😄

PR #308, opened 2023-01-27
fix chainctl curl command

k4leung4:
## Type of change

old path points to dated 0.1.44, point to latest

matches the path here, https://github.com/chainguard-dev/edu/blob/b5b223538769c5624fbbb233c0e113e888e9d831/content/chainguard/chainguard-enforce/how-to-install-chainctl.md?plain=1#L52

fix: https://github.com/chainguard-dev/mono/issues/7664

### What should this PR do?

fix incorrect download command


SharpRake (approved):

SharpRake:
Great catch! Thanks Kenny.

Issue #286, opened 2023-01-19
Resource: How to use cloud discovery in Chainguard Enforce


SharpRake:
**What topic are you requesting a resource about?**
* **Chainguard product**
* Open source related
* Conceptual security related
* Other (please describe)

**Proposed title:**

How to use cloud discovery in Chainguard Enforce

**Description:**

A standalone guide outlining how to use Enforce's new `discover` feature

ltagliaferri:
Does this issue refer to https://github.com/chainguard-dev/edu/pull/309 or something else?

SharpRake:
> Does this issue refer to #309 or something else?

Yes it does, this can be closed.

PR #309, opened 2023-01-27
adding Getting Started with Enforce Discovery doc, also playing with …

SharpRake:
…ordering slightly

## Type of change

This adds an onboarding doc for Discovery.

### What should this PR do?

it's one component of a larger issue I've been assigned on documenting Discovery. It specifically resolves issue #286 

It's a new cool feature we want folks to try out.

Copy edit and tech test. This came from a doc drafted by Kim and Matt Moore so it all works pretty smoothly, but if anything feels underexplained please let me know.

### DIFF
--- a/content/chainguard/chainguard-enforce/chainguard-enforce-kubernetes/chaingaurd-enforce-discovery-onboarding.md
+++ b/content/chainguard/chainguard-enforce/chainguard-enforce-kubernetes/chaingaurd-enforce-discovery-onboarding.md
@@ -1,0 +1,136 @@
+---
+title: "Getting Started with Chainguard Enforce Discovery"
+type: "article"
+description: "Use Chainguard Enforce's Discovery feature to discover workloads and resources from multiple cloud providers."
+date: 2023-26-01T11:07:52+02:00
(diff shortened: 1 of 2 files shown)


content/chainguard/chainguard-enforce/chainguard-enforce-kubernetes/chaingaurd-enforce-discovery-onboarding.md
```diff
@@ -0,0 +1,136 @@
...
+	"CLOUD_RUN",
+	"ECS",
+	"EKS",
+	"GKE",
+  ]
+  profiles = ["enforcer"]
+}
+```
+
+This will return the Discovery results, which can be piped into other resources (possibly as an output). The main use case for this data, though, will be to feed it into a “chainguard_cluster” resource:
```
ltagliaferri, line 108:
```suggestion
This will return the Discovery results, which can be piped into other resources (possibly as an output). The main use case for this data, though, will be to feed it into a `chainguard_cluster` resource:
```

ltagliaferri (approved):

PR #321, opened 2023-01-30
Disable terminal while backend is being repaired

jamonation:
The cluster that hosts the terminal is unhealthy, so this PR turns off the terminal on all pages by disabling the Hugo partial template.


jamonation:
Confirmed that this works - pages with the terminal still render correctly, but on the live site they won't load the iframe: https://deploy-preview-321--ornate-narwhal-088216.netlify.app/open-source/apko/getting-started-with-apko/

ltagliaferri (approved):

PR #339, opened 2023-02-04
Fixing typo in Discovery URL

SharpRake:
## Type of change

This is another small fix, this time to fix the URL of the DIscovery onboarding.


ltagliaferri (changes requested):
Good catch, can you add the redirect to the [NGINX config](https://github.com/chainguard-dev/edu/blob/a8028da52e93ac3f7d0c879f67b11ab92eae7b3a/nginx.conf#L30), in this block:

```
    # add URLs after the `default` line that are moved and aren't redirecting via Hugo aliases
    map $request_uri $redirect_url {
        default "";
        "~^/chainguard/chainguard-enforce/chainctl-docs/how-to-install-chainctl(.+)?$" /chainguard/chainguard-enforce/how-to-install-chainctl/;
    }
```

ltagliaferri (approved):

PR #355, opened 2023-02-08
Fix: Invalid policy

pdeslaur:
## Type of change
Bug fix

### What should this PR do?
Updates a policy to be valid.

Policy is broken


jamonation (approved):
+1, works for me!

PR #363, opened 2023-02-09
Fix enforce signing typos

found-it:
## Type of change

Just simple typo fixes in the enforce signing docs

### What should this PR do?

Typo fixes

Just making documentation more clear


jamonation (approved):
+1!

PR #400, opened 2023-02-21
Add network requirements section about image registry access

jamonation:
### What should this PR do?

This PR adds a section about image registry access for the Enforce agent.

Some environments may be blocking Enforce from accessing image registries, so this change makes that requirement explicit.

The new section will show up on the netlify preview page.

Check the network requirements page on netlify.


ltagliaferri (approved):
LGTM, thank you!

PR #442, opened 2023-03-07
Enforce signing: recommend using cosign v2

hectorj2f:
## Type of change

Update our scripts to recommend using cosign v2 instead of a release candidate

### What should this PR do?

Use cosign v2 instead of a release candidate

Using cosign v2 release candidates should not be recommended, instead we should document the usage of cosign v2.

related to https://github.com/chainguard-dev/mono/issues/8949

Perhaps we should wait until the PR is merged in mon.


ltagliaferri (approved):

ltagliaferri:
Thank you @hectorj2f , I'm going to close the other PR as the policy controller file snuck into this PR too 😄

PR #465, opened 2023-03-12
Add github workflow to build terminal-images

jamonation:
### What should this PR do?
This PR adds a github action to build terminal images using `melange` and `apko` exclusively.

Images have only been built on a best-effort basis. With this in place, all the images for academy can be built automatically using a matrix job strategy.

The action should not fail (it probably will at first).

Run the action with `academy-base` as the image to build. The resulting image will be pushed to the registry that hosts images for the terminal, but it will be called `academy-base:testing` so it will not overwrite any existing images. It will not be usable either since it is not included in the allow list.


jamonation:
Here's the updated workflow to build all the images for Academy. I ran it yesterday on my fork and it worked:

Screenshot 2023-03-19 at 8 31 49 AM

Images and sboms and signatures show up in the registry e.g. here's one for rekor:
Screenshot 2023-03-19 at 8 33 47 AM

jamonation:
I think this is ready to go @ltagliaferri. The signature stuff I can sort out later - right now it can at least be downloaded with `cosign download signature <image ref>`.

@erikaheidi, could I get your 👀 on this too please since it is building images with melange/apko/github actions? Maybe just a test that you can `melange build` and `apko build` one of the images, like `apko` itself, or `rekor`?

ltagliaferri (approved):
🎉

erikaheidi (approved):
It builds fine here! I tried the apko image, both melange and apko worked.

PR #478, opened 2023-03-15
Update autodocs-images.yaml

erikaheidi:
This PR updates the autodocs action version so that we're now able to generate a page with details about image variants. 

Once this PR is accepted, I will run the automation to generate the first batch of specs pages, and this will open a new pull request that we'll be able to review and share to others using the Netlify preview.


jamonation (approved):
+1 here!

PR #512, opened 2023-03-24
Bumping autodocs and creating workflow for slack notifications

erikaheidi:
This PR bumps the autodocs action to the newest version that fixes a typo in the image provenance template. It also creates a new workflow that is a starting point for the autodocs slack bot.


jamonation (approved):
Well that's cool!

PR #523, opened 2023-03-27
Add partial to render data/notices.yaml for any page that needs a notice

jamonation:
## Type of change
platform

### What should this PR do?
This PR adds a partial that will pick up on any note in `data/notices.yaml` and render it at the top of a page. Notes can be written in Markdown and specify a type of `NOTICE, WARNING, DEPRECATION` to get a coloured background.

Sometimes pages need a banner notice at the top for ephemeral changes and the like. This PR is intended to be used for one-off notices, warnings, and deprecation notes on pages.

Pages with notes will have a block at the top, those with no notes will have none.

Test locally using the included JRE note. Run `hugo serve` and visit http://localhost:1313/chainguard/chainguard-images/reference/jre/overview/

You should see the following: 

Screenshot 2023-03-27 at 3 06 15 PM

### DIFF
--- a/data/notices.yaml
+++ b/data/notices.yaml
@@ -12,8 +12,8 @@
 # The 'type' field will render the div with the selfsame class applied.
 
 - title: "Image Overview: jre"
-  note: |
-    This image will switch from including a full compliment of JDK binaries to a minimally required set in a forthcoming release. The work is being tracked in [this GitHub issue](https://github.com/chainguard-images/images/issues/297).
-
-    If you rely on any binaries in this image that are not a part of the following list, you should switch to using the [JDK Image](/chainguard/chainguard-images/reference/jdk/overview/).
   type: WARNING
+  note: |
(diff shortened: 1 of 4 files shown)


data/notices.yaml
```diff
@@ -0,0 +1,19 @@
...
+#     A line, or multi-line message with details.
+#
+#     The note can be plain text, or use Markdown.
+#   type: One of NOTICE, WARNING, or DEPRECATED
+#
+# The 'type' field will render the div with the selfsame class applied.
+
+- title: "Image Overview: jre"
+  note: |
+    This image will switch from including a full compliment of JDK binaries to a minimally required set in a forthcoming release. The work is being tracked in [this GitHub issue](https://github.com/chainguard-images/images/issues/297).
```
ltagliaferri, line 16: Should it be "complement" instead of compliment?

ltagliaferri (approved):
Just one q about the word spelling

data/notices.yaml
jamonation: Thanks for catching that!

PR #529, opened 2023-03-28
Update UX docs

julienv3:
## Type of change

This PR updates UI docs to reflect the latest changes in https://console.enforce.dev.

This update wording that mentions obsolete concepts in the console such as the top-right filter button.


jamonation (approved):
LGTM!

PR #530, opened 2023-03-28
Fix notice partial and css rules with working class names

jamonation:
## Type of change
Bug

### What should this PR do?
Fixes missing CSS styles on notice partial blocks.

CSS classes weren't being rendered to the final main.css file

A warning block should be coloured coral red.

Check the netlify preview for the [JRE images overview page](https://deploy-preview-530--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/reference/jre/overview/)

### DIFF
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -29,9 +29,11 @@ blockquote {
   margin: 1.5em 10px;
   padding: 0.5em 10px;
 }
+
 blockquote > p {
-  margin-bottom: 0px;
+  margin-bottom: 0;
 }
+
 .featured-icon {
   font-size: 4rem;
 }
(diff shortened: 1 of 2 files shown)


assets/scss/common/_custom.scss
```diff
@@ -72,14 +75,26 @@ h4 code {
...
+[data-dark-mode] #note.deprecated {
+  background: yellowgreen;
+}
+
+[data-dark-mode] #note.notice {
+  background: darkcyan;
+}
+
+[data-dark-mode] #note.warning {
+  background: darkred;
```
ltagliaferri, line 99: Would you be against making it more yellow / orange in dark mode?

jamonation: Hah I'm probably least qualified to pick colours, so anything you suggest I'll go with!

jamonation: Those look good to me, thanks for updating!

ltagliaferri (approved):

Issue #516, opened 2023-03-27
Move Images autodocs templates outside of GH Actions


ltagliaferri:
The current workflow is not sustainable when there are changes

PR #539, opened 2023-03-29
Bringing autodocs templates to facilitate updating - #516

erikaheidi:
This PR addresses #516 and updates autodocs to the latest version, which supports using a custom location for page templates. With this update, we won't need to bump the action version every time we need a small template change, and it will be easier for other people to update these templates when needed. The updated workflow also includes the slack notification step, replacing the old `autodocs-notify` workflow.

A new `autodocs` folder is added, where we'll keep the templates and the changelog for latest updates. For now, the changelog only tells if there was a new image added, but it can be improved in the future to show more useful information.


autodocs/templates/image_provenance_page.tpl
```diff
@@ -0,0 +1,65 @@
...
+```
+
+By default, this command will fetch signatures for the `latest` tag. You can also specify the tag you want to fetch signatures for.
+
+## Downloading and Verifying SBOMs
+
+All Chainguard Images come with a high-quality Software Bill Of Materials (SBOM) generated at build-time. The SBOM can be downloaded using the cosign tool:
+
+```shell
+cosign download sbom --platform linux/amd64 cgr.dev/chainguard/{{ title }} | jq
```
jamonation, line 35: Should this `--platform linux/amd64` be there? We have linux/arm64 images too right?

.github/workflows/autodocs-images.yaml
```diff
@@ -25,31 +25,66 @@ jobs:
...
 
       - name: "Copy updates to main repo"
         run: |
           echo "Copying files..." && \
           cp -R "${{ github.workspace }}/${{ env.WORKDIR }}" "${{ github.workspace }}/edu/content/chainguard/chainguard-images" && \
           echo "Finished copy"
 
+      - name: "Get Changelog"
+        run: |
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
```
jamonation, line 57: Clever!

jamonation (approved):
Looks good, just the one question about handling other `--platform` architectures.

autodocs/templates/image_provenance_page.tpl
erikaheidi: Yeah this was from the first time I build these provenance pages, and we only had amd64 at that time... I think it's safe to remove it.

PR #544, opened 2023-03-30
Updating autodocs workflow to fix image docs changelog

erikaheidi:
This PR updates the image autodocs workflow to fix an issue with the changelog generation. Now we can set the source dir for the "diff" so it works correctly when comparing the generated folders with existing content.


jamonation (approved):
👍

PR #548, opened 2023-03-30
Updating Wolfi with Dockerfiles guide

erikaheidi:
This PR brings some updates to the new Wolfi with Dockerfiles guides, based on feedback from @amouat and @ltagliaferri .

It also fixes a typo in the melange tutorial tags definition.

Preview: https://deploy-preview-548--ornate-narwhal-088216.netlify.app/open-source/wolfi/wolfi-with-dockerfiles/

### DIFF
--- a/content/open-source/wolfi/wolfi-with-dockerfiles.md
+++ b/content/open-source/wolfi/wolfi-with-dockerfiles.md
@@ -15,7 +15,7 @@ toc: true
 
 ## Introduction
 
-[Wolfi](https://edu.chainguard.dev/open-source/wolfi/overview/) is a minimal open source Linux distribution created specifically for cloud workloads, with an emphasis on software supply chain security. Using [apk](https://wiki.alpinelinux.org/wiki/Package_management) for package management, Wolfi differs from Alpine in a few important aspects, most notably the use of glib-c instead of Musl and the fact that Wolfi doesn't have a kernel of its own since it is built to run on containers. The minimal footprint makes Wolfi an ideal base for both _distroless_ runtimes and fully-featured builder images.
+[Wolfi](https://edu.chainguard.dev/open-source/wolfi/overview/) is a minimal open source Linux distribution created specifically for cloud workloads, with an emphasis on software supply chain security. Using [apk](https://wiki.alpinelinux.org/wiki/Package_management) for package management, Wolfi differs from Alpine in a few important aspects, most notably the use of glibc instead of musl and the fact that Wolfi doesn't have a kernel of its own since it is built to run on containers. The minimal footprint makes Wolfi an ideal base for both _distroless_ runtimes and fully-featured builder images.
 
 A _distroless_ image is typically a very minimal container image that doesn't have shells or package managers. The extra tightness improves security in several aspects, but it requires a more sophisticated strategy for image composition since you can't install packages so easily.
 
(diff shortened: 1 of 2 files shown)


content/open-source/wolfi/wolfi-with-dockerfiles.md
```diff
@@ -110,10 +97,12 @@ Copy the following content to it:
 ```
 FROM cgr.dev/chainguard/wolfi-base
```
amouat, line 98: I'm fine with this, I just wonder if using python:$version-dev would be simpler?

content/open-source/wolfi/wolfi-with-dockerfiles.md
```diff
@@ -175,10 +167,10 @@ RUN  pip3 install -r requirements.txt --user
 
 FROM cgr.dev/chainguard/python:latest
```
amouat, line 168: think you can do `python:$version` to avoid compatibility problems.

amouat (approved):

content/open-source/wolfi/wolfi-with-dockerfiles.md
```diff
@@ -110,10 +97,12 @@ Copy the following content to it:
...
 FROM cgr.dev/chainguard/wolfi-base
 
+# Use this variable to set up Python version
+ARG version=3.11
+
 RUN adduser -D nonroot
 WORKDIR /app
-
-RUN apk add python-3.11 py3.11-pip && \
+RUN apk add python-$version py$version-pip && \
```
amouat, line 105: you should be able to do `py${version}-pip` which is a bit easier to read.

ltagliaferri (approved):
LGTM after you address @amouat's edits, thank you @erikaheidi!

erikaheidi:
> Great!
> 
> Small things:
> 
>     * You can use `${version}` to make it clearer where the variable starts and stops
> 
>     * There's still going to be a problem when `latest` moves version I _think_
>       
>       * can't we use `python:3.11` instead? You can even do this I think:
> 
> 
> ```
> ARG version=3.11
> FROM cgr.dev/chainguard/python:$version
> ```

I tried the `FROM` line with a variable but it doesn't seem to work, maybe the variable is not in that context? 

```
Step 9/14 : FROM cgr.dev/chainguard/python:$version
invalid reference format

```

amouat:
Interesting, let me have a look...

amouat:
OK, it's an interesting one. If you move the `ARG version=3.11` line to be the first line in the file (before the first FROM) it should work.

erikaheidi:
> 

That didn't work here - now the variable is not resolved:
```
fetch https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  py-pip (no such package):
    required by: world[py-pip]
  python- (no such package):
    required by: world[python-]
```

I have changed it to use the 3.11 tag instead of latest, let me know if that works.

amouat:
Sure, that's fine.

Can you send me the Dockerfile that isn't working for you?

erikaheidi:
> Sure, that's fine.
> 
> Can you send me the Dockerfile that isn't working for you?

The Dockerfile only worked with the VAR added to three locations (at the top, and on each stage). I am not sure this is the best way because of the duplication. So I'll be merging these updates now to incorporate them asap on the live piece, but we can always update the Dockerfile.

PR #551, opened 2023-03-31
Update autodocs-images.yaml

erikaheidi:
This PR bumps autodocs to the latest version, which fixes the changelog generation and the slackbot notification when a PR is open.


.github/workflows/autodocs-images.yaml
```diff
@@ -63,6 +63,7 @@ jobs:
 
       - name: Create a PR
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
+        id: cpr
```
jamonation, line 66: Ah yeah these id versus name things are so annoying. I wish github would just make both mandatory.

jamonation (approved):

PR #554, opened 2023-03-31
Updating workflow to build persistent changelog

erikaheidi:
Another try at fixing the changelog persistence for autodocs images. The problem is that the application don't have access to the current changelog while building the new batch of automated docs, so now I'm using a step in the workflow to append the last update into the changelog file. Hoping this works now :)

### DIFF
--- a/.github/workflows/autodocs-images.yaml
+++ b/.github/workflows/autodocs-images.yaml
@@ -37,14 +37,12 @@ jobs:
           YAMLDOCS_SOURCE: "${{ github.workspace }}/images/images"
           YAMLDOCS_OUTPUT: "${{ github.workspace }}/${{ env.WORKDIR }}"
           YAMLDOCS_TEMPLATES: "${{ github.workspace }}/edu/autodocs/templates"
-          YAMLDOCS_CHANGELOG: "${{ github.workspace }}/${{ env.WORKDIR }}/changelog.md"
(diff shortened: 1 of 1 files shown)


.github/workflows/autodocs-images.yaml
```diff
@@ -53,14 +51,19 @@ jobs:
...
           EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
           echo "CHANGELOG<<$EOF" >> $GITHUB_ENV
           echo "`cat ${{ github.workspace }}/edu/autodocs/last-update.md`" >> $GITHUB_ENV
           echo "$EOF" >> $GITHUB_ENV
         id: images-build-output
 
+      - name: "Append Update to Changelog"
+        run: |
+          [ ! -f "${{ github.workspace }}/edu/autodocs/changelog.md" ] && touch ${{ github.workspace }}/edu/autodocs/changelog.md
+          sed -i '1s/^/${{ env.CHANGELOG }}\n\n/' ${{ github.workspace }}/edu/autodocs/changelog.md
```
jamonation, line 65: It always comes back to sed grep and awk doesn't it 🤣

jamonation (approved):

.github/workflows/autodocs-images.yaml
erikaheidi: YESSSSS! gosh.

PR #559, opened 2023-04-03
Updating image autodocs workflow to keep changelog history

erikaheidi:
This PR is another try at fixing the changelog persistence issue I was having in the previous version. The current workflow should work without permission issues.


jamonation (approved):
+1

erikaheidi:
This didn't work unfortunately, so I'm reverting to un-break the automation.

PR #566, opened 2023-04-04
Add gitsign to enforce github actions

jamonation:
### What should this PR do?
This PR adds gitsign support to the Enforce autodocs workflow.

Commits ought to be signed when run via Github actions.

N/A

Only way to test here easily is to merge this, run the action, and check for a signed commit.


SharpRake (approved):
:+1: I trust you! Let's merge this and see how it works

PR #582, opened 2023-04-10
trying to move chainctl docs - do not merge

SharpRake:
## Type of change

just testing, will edit this if it works


jamonation:
This one is a bit tricky to test @SharpRake - it has to be merged into the `main` branch for the change to the Github workflow to get picked up. If you want to test the changed rsync command I'll get you some credentials so that you can run the `gcloud` commands locally - that ought to give you a better idea of where the files will end up on the rendered site.

PR #584, opened 2023-04-11
Sign images autodocs

ltagliaferri:
I looked at your autodocs @jamonation but I know this is set up a little differently. Just want some signing here too 😄


jamonation (approved):
Just like that, magic signing

PR #596, opened 2023-04-14
I messed everything up, here's a new PR

SharpRake:
## Type of change

This PR is to replace one that I accidentally closed/broke. It will move the chainctl docs to their own top level folder under the `content` directory.

### What should this PR do?

This is the first step of the Enforce docs IA overhaul.

None of the automation should break, the chainctl docs and navigation should all be moved appropriately.


jamonation (approved):
Looks good in advance of regenerating Enforce chainctl docs. Let's go!

PR #598, opened 2023-04-14
Tidy integrate enforce docs github action

jamonation:
## Type of change
bug

### What should this PR do?
Fix integrate enforce docs action

To make it work

Action should work

Run the action


SharpRake (approved):
looks great!

PR #600, opened 2023-04-14
deleting original chainctl docs folder, fixing new index file

SharpRake:
## Type of change

This PR deletes the original chainctl-docs directory and makes a minor change to the new chainctl directory's `_index.md` file.
### What should this PR do?

No

Cleaning up chainctl docs after move.

Everything looks right, the chainctl docs are still accessible but they are in the correct location.


jamonation (approved):
+1!

PR #601, opened 2023-04-14
change location of why apk

ltagliaferri:
Changing location (and menu weight) of "why apk" based on discussion here: https://github.com/chainguard-dev/internal/issues/1665#issuecomment-1508948376

Also updated the nginx config


jamonation (approved):
Makes sense to move this around +1!

PR #604, opened 2023-04-15
Update url to upstream project

bengtfredh:
* fix typo in upstream doc link


ltagliaferri (commented):
Thanks for sending the typofix, but these docs are coming from the upstream Images repo, you can find the image here if you would like to fix it there: https://github.com/chainguard-images/images/tree/main/images/k8s-sidecar

PR #614, opened 2023-04-19
Run Enforce docs on a schedule, only when there is a release

jamonation:
## Type of change
automated enhancement

### What should this PR do?
This PR runs the Enforce docs workflow on an hourly schedule. It also only runs if there is a detected difference between upstream releases and published releases in the changelog. 

It isn't much, but this cuts down the toil of monitoring when Enforce releases are published and running the GitHub action here manually.

The schedule should run hourly, and only open a PR if there's a delta between current and latest docs in the changelog.

Here is an actions run showing a no-op when both released and published docs match: https://github.com/chainguard-dev/edu/actions/runs/4738648929

And here is one when the versions are different:
https://github.com/chainguard-dev/edu/actions/runs/4738609362

### DIFF
--- a/.github/workflows/autodocs-enforce.yaml
+++ b/.github/workflows/autodocs-enforce.yaml
@@ -2,6 +2,9 @@ name: Build Enforce Docs
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - "update-enforce-workflow"
 
 env:
   PROJECT_ID: "${{ secrets.PROJECT_ID }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/autodocs-enforce.yaml
```diff
@@ -1,7 +1,12 @@
 name: Build Enforce Docs
 
 on:
+  schedule:
+    - cron: "0 * * * *"
```
ltagliaferri, line 5: 👍

ltagliaferri (approved):

PR #640, opened 2023-04-26
Add nginx redirects for stale incoming urls

jamonation:
## Type of change
bug

### What should this PR do?
This PR adds nginx redirects for incoming URLs that land on 404 pages. 

Search engines have some stale results that point at 404 pages. They should eventually get updated when the site is crawled again but in the meantime these redirects will get users to the page they are after.

nginx redirects should work

```
docker build . -t academy-nginx:latest
docker run --rm -t -p 8080:8080 academy-nginx:latest
curl -sIL localhost:8080/chainguard/chainguard-enforce/chainguard-enforce-kubernetes/chainguard-enforce-policy-examples/ |egrep '(HTTP|Location)'
```

Example curl commands ought to return a 301, then 200 response like this:

```
HTTP/1.1 301 Moved Permanently
Location: http://localhost:8080/chainguard/chainguard-enforce/policies/chainguard-enforce-policy-examples/
HTTP/1.1 200 OK
```


ltagliaferri (approved):

jamonation:
A couple more to add here before merging, working on them now.

ltagliaferri (approved):

PR #646, opened 2023-04-28
Disable terminal until k8s cluster is fixed

jamonation:
GKE network endpoint is broken, so requests can't reach the terminal environment. Disabling until the terminal environment is reachable again.


SharpRake (approved):
:+1:

PR #713, opened 2023-05-19
Fixing description in debugging distroless article

erikaheidi:
Small fix to the Debugging Distroless article, just now realized the description was not correct.


jamonation (approved):
Nice catch!

PR #720, opened 2023-05-23
ArgoCD overview: fix link

axelsimon:
## Type of change
documentation fix

### What should this PR do?
fix a link in docs

### DIFF
--- a/content/chainguard/chainguard-images/reference/argocd/overview.md
+++ b/content/chainguard/chainguard-images/reference/argocd/overview.md
@@ -78,7 +78,7 @@ dex:
 
 ### ArgoCD Components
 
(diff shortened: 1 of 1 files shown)


ltagliaferri (commented):
Thanks for sending this fix! These docs are autogenerated from the Images upstream repo, so the file that would need to be updated is in the ArgoCD Images Readme, and this would pull in changes automatically on a timer: https://github.com/chainguard-images/images/tree/main/images/argocd

axelsimon:
@ltagliaferri got it, will try to make the change there!

ltagliaferri:
Ah sorry about that @axelsimon , it looks like someone else had the same idea as you https://github.com/chainguard-images/images/pull/609

There probably are more broken links, and easy docs wins in the readmes if you are interested in helping out 😄

PR #734, opened 2023-05-26
Adding guide on how to use tag history api

erikaheidi:
### What should this PR do?
This PR adds a guide on how to use the new tag history API for Chainguard Images.

We need to let users know how they can "lock" a build to specific image digests, which may become more necessary as we'll make only `latest` and `latest:dev` public.

Preview:
https://deploy-preview-734--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/using-the-tag-history-api/

### DIFF
--- a/content/chainguard/chainguard-images/using-the-tag-history-api.md
+++ b/content/chainguard/chainguard-images/using-the-tag-history-api.md
@@ -40,13 +40,17 @@ For public images (tagged as `latest` or `latest-dev`), you can request a regist
 The following command will obtain a token for the **Python** image and register a variable called `tok` with the resulting value, which you can use in a subsequent command to obtain the tag history:
 
 ```shell
-tok=$(curl "https://cgr.dev/token?scope=repository:chainguard/python:pull" | jq -r .token)
+tok=$(curl "https://cgr.dev/token?scope=repository:chainguard/python:pull" \
+  | jq -r .token)
 ```
 
 For images that are not part of the public catalog, you'll need to exchange your Chainguard token for a registry token. This assumes you've set up auth with [chainctl auth configure-docker](https://edu.chainguard.dev/chainguard/chainguard-images/registry/authenticating/):
 
 ```shell
-tok=$(curl -H "Authorization: Bearer $(echo 'cgr.dev' | docker-credential-cgr get)" -v "https://cgr.dev/token?scope=repository:chainguard/python:pull" | jq -r .token)
+tok=$(curl -H "Authorization: Bearer \
+  $(echo 'cgr.dev' | docker-credential-cgr get)" \
+  -v "https://cgr.dev/token?scope=repository:chainguard/python:pull" \
+  | jq -r .token)
 ```
 
 To make sure your token is set, you can run the following command:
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-images/using-the-tag-history-api.md
```diff
@@ -0,0 +1,111 @@
...
+The following command will obtain a token for the **Python** image and register a variable called `tok` with the resulting value, which you can use in a subsequent command to obtain the tag history:
+
+```shell
+tok=$(curl "https://cgr.dev/token?scope=repository:chainguard/python:pull" | jq -r .token)
+```
+
+For images that are not part of the public catalog, you'll need to exchange your Chainguard token for a registry token. This assumes you've set up auth with [chainctl auth configure-docker](https://edu.chainguard.dev/chainguard/chainguard-images/registry/authenticating/):
+
+```shell
+tok=$(curl -H "Authorization: Bearer $(echo 'cgr.dev' | docker-credential-cgr get)" -v "https://cgr.dev/token?scope=repository:chainguard/python:pull" | jq -r .token)
```
ltagliaferri, line 49: What do you think about breaking this command up so that it's easier to read?

content/chainguard/chainguard-images/using-the-tag-history-api.md
```diff
@@ -0,0 +1,111 @@
...
+
+To make sure your token is set, you can run the following command:
+
+```shell
+echo $tok
+```
+
+And you should get a long string token as output.
+
+You should now be ready for calling the API, either manually or programmatically.
```
ltagliaferri, line 60:
```suggestion
You should now be ready to call the API, either manually or programmatically.
```

content/chainguard/chainguard-images/using-the-tag-history-api.md
```diff
@@ -0,0 +1,111 @@
...
+    },
+    ...
+
+  ]
+}
+```
+
+## Using Image Digests within a Dockerfile
+
+Setting up your Dockerfile to use a stale build is a matter of modifying your `FROM` line to use an image digest instead of a tag. For instance, let's say you want to make sure you keep using the current latest build of the Python image. In a previous section of this page we obtained the tag history of the Python image, and the most recent build digest is listed as `sha256:81c334de6dd4583897f9e8d0691cbb75ad41613474360740824d8a7fa6a8fecb`. With that information, you can edit your Dockerfile and replace:
```
ltagliaferri, line 99:
```suggestion
Setting up your Dockerfile to use an older build is a matter of modifying your `FROM` line to use an image digest instead of a tag. For instance, let's say you want to make sure you keep using the current latest build of the Python image. In a previous section of this page we obtained the tag history of the Python image, and the most recent build digest is listed as `sha256:81c334de6dd4583897f9e8d0691cbb75ad41613474360740824d8a7fa6a8fecb`. With that information, you can edit your Dockerfile and replace:
```

ltagliaferri (approved):
Awesome and quick work here, Erika! Everything worked very well, I made a few suggestions. I know we are probably using the term "stale builds" when we talk about this, but I think for user facing being more neutral would be nice as they probably have a good reason for pinning to an older version of an image.

One thing I would add, and I am happy to do this if you would like to merge this in now, would be a "Learn more" kind of section at the end where there is a brief statement on the Images catalogs and that the latest is free to use but to pin to an older version will require the premium catalog. We can point to the FAQs which I updated with that info:

```
/chainguard/chainguard-images/faq/#what-are-the-different-catalog-tiers-of-chainguard-images
```

erikaheidi:
Alright, I added the closing session. Merging now!

PR #784, opened 2023-06-16
suggest an alternative tool for installing qemu emulators

developer-guy:

### DIFF
--- a/content/open-source/melange/troubleshooting.md
+++ b/content/open-source/melange/troubleshooting.md
@@ -46,9 +46,14 @@ To enable additional architectures, you'll need to enable them within your kerne
 docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
 ```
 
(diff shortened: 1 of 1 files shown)


content/open-source/melange/troubleshooting.md
```diff
@@ -46,9 +46,14 @@ To enable additional architectures, you'll need to enable them within your kerne
 docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
 ```
 
+or you can use the following as an alternative approach, they both will do the same thing:
```
ltagliaferri, line 49:
```suggestion
An alternate approach to achieve the same result is to run the following command:
```

ltagliaferri (changes requested):
Thanks so much for adding this! I rephrased the sentence a little if you think that sounds good 😄

ltagliaferri (approved):

PR #795, opened 2023-06-22
Fix card and footer link colours

jamonation:
## Type of change
platform development

### What should this PR do?
This fixes the card link and footer link hover colour differences from the Figma design that @traversevans spotted the other day. 

Consistency!

Link colours should be light/dark to match regular header colours on list pages. Footer hover links should be Chainguard blue.

Check netlify

### DIFF
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -11,7 +11,7 @@
   padding-bottom: 0;
 }
 
-.docs-content > h1:first-child {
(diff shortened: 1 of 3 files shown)


jamonation:
I also made each of the cards clickable in their entirety based on feedback from @traversevans 

Screenshot 2023-06-22 at 1 49 55 PM

ltagliaferri (approved):

PR #802, opened 2023-06-27
fix: use the right tool

hectorj2f:
## Type of change

Change the tool name used to generate SBOMs.

### What should this PR do?

Fix: we are using Syft from anchore to generate sboms, we aren't using Grype for sbom generation.


SharpRake (approved):
Thanks for this catch, Hector!

PR #818, opened 2023-07-05
Sm article updates

mcaveety:
## Type of change
Minor updates to CVE Intro article and SLSA index

### What should this PR do?
Provide minor fixes to the aforementioned article / index

There was some missing punctuation and inaccurate descriptions in these two files

Changes should make sense and improve the articles ever so slightly

Do a quick passover of the changes :)


content/software-security/cves/cve-intro.md
mmcaveety, line 1: I specifically changed the description because that is surfacing in the articles as opposed to the lead. Thought it would be the other way around, so this fixes my confusion and shows what I originally intended

ltagliaferri (approved):

PR #823, opened 2023-07-06
Add rumble-cve-data.yaml workflow to parse CVE data into CSVs

jamonation:
## Type of change
platform development

### What should this PR do?
This adds a GitHub action that extracts CVE data from rumble and uploads it to GCS as CSV files.

This action is needed to get data flowing into GCS so that it can be displayed on image overview pages. These files can then be parsed and used on image overview pages to show tabular information about CVEs.

The action ought to work.

Run it once this is merged with fingers 🤞

### DIFF
--- a/.github/workflows/rumble-cve-data.yaml
+++ b/.github/workflows/rumble-cve-data.yaml
@@ -1,0 +1,90 @@
+name: Rumble CVE Files
+on:
+  schedule:
+    - cron: "1 5 * * *"
+  workflow_dispatch:
+  push:
+    branches: [auto-rumble]
+
+env:
+  PROJECT_ID: "${{ secrets.PROJECT_ID }}"
+  STORAGE_BUCKET: "${{ secrets.STORAGE_BUCKET }}"
+  WORKLOAD_IDENTITY_PROVIDER: "${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}"
+  SERVICE_ACCOUNT: "${{ secrets.GH_ACTION_SERVICE_ACCOUNT }}"
+  GH_TOKEN: ${{ github.token }}
+
+jobs:
+  generate-cve-files:
+    runs-on: ubuntu-latest
+
+    strategy: 
+      matrix:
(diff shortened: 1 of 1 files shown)


jamonation:
Here's what it looked like in my forked repo running on a local GitHub actions runner:

Screenshot 2023-07-06 at 2 16 26 PM

.github/workflows/rumble-cve-data.yaml
```diff
@@ -0,0 +1,89 @@
...
+  SERVICE_ACCOUNT: "${{ secrets.GH_ACTION_SERVICE_ACCOUNT }}"
+  GH_TOKEN: ${{ github.token }}
+
+jobs:
+  generate-csvs:
+    runs-on: ubuntu-latest
+
+    strategy: 
+      matrix:
+        image: ["bash", "busybox", "deno", "git", "go", "kube-state-metrics", "mariadb", "maven", "minio", "minio-client", "nginx", "node", "php", "python", "rabbitmq", "ruby", "rust", "wait-for-it", "wolfi-base"]
```
ltagliaferri, line 20: nit but what do you think about breaking this up on different lines? if we expand the number of image data we ingest it will get unwieldy

ltagliaferri (approved):
Just one nitpick here

.github/workflows/rumble-cve-data.yaml
jamonation: Yeah it is something strange with GitHub actions - I tried breaking it up but it complains about an "unexpected end of the stream within a flow collection". I think we could move it into an environment variable though, and store the set of images as a variable in the repository settings here on GitHub?

ltagliaferri: oh weird! because it's a data structure?

jamonation: Figured it out - apparently GitHub actions uses different kinds of YAML notation within a single file. The flow collection error message means that GitHub expects the matrix line to use that flow style. So then the question, what is flow style? TIL, it is a compact YAML notation that puts everything on a single line: https://www.yaml.info/learn/flowstyle.html

PR #832, opened 2023-07-11
docs: update chainctl to account-associations

denhamparry:
- update chainctl cloud account association command

## Tired

```sh
chainctl iam group set-aws
```

## Inspired

```sh
chainctl iam account-associations set aws
```

Signed-off-by: Lewis Denham-Parry <lewis@chainguard.dev>


SharpRake (approved):
Thanks (again) Lewis!

PR #833, opened 2023-07-12
docs: fix chainctl cluster install

denhamparry:
Found a reference to `chainctl install cluster`, and it should be `chainctl cluster install`.

I appreciate that there is some linting going on in this PR, I can look to remove if required.

- correct order of chainctl cluster install


SharpRake (approved):
:+1: Thanks for catching this Lewis!

PR #836, opened 2023-07-13
Fix copy around scanner used by rumble

erikaheidi:
Quick fix to copy used around rumble, used to be Trivy but it changed recently to Grype.


SharpRake (approved):
lgtm!

PR #849, opened 2023-07-18
Update Autodocs Workflow

erikaheidi:
This PR updates Autodocs to the latest version:

- New tags history page, now grouping tags with same digest
- Updated workflow to use repo secrets when setting up chainctl
- moved env variables for better readability

TODO: 
- Image specs page, will require a lot more work and several `crane` calls to obtain image metadata, so I will evaluate if this is viable or not.
- Use rsync instead of copy to remove retired images: blocked by the image specs change update, since this would delete these pages that are momentarily not being generated by the automation.


jamonation (approved):
+1 from me!

PR #854, opened 2023-07-18
Fixing tags table in image overview page

erikaheidi:
This PR updates autodocs with a fix to the tags table in the image overview page - only realized the issue after merging the previous PR.


SharpRake (approved):
:tada:

Issue #841, opened 2023-07-14
Broken links in the How to Use Chainguard Images article


stormqueen1990:
**Describe the bug**
In the ["How to Use Chainguard Images" article](https://edu.chainguard.dev/chainguard/chainguard-images/how-to-use-chainguard-images/), the following sentence contains two links that point to repositories that don't exist in the chainguard-images GitHub org:
> If your binary is dependent on glibc or musl, take a look at the [glibc-dynamic](https://github.com/chainguard-images/glibc-dynamic) and [musl-dynamic](https://github.com/chainguard-images/musl-dynamic) images.

**To Reproduce**
1. Navigate to https://edu.chainguard.dev/chainguard/chainguard-images/how-to-use-chainguard-images/
2. Scroll down to the end of the "Using the Distroless Base Images" section.
3. Notice that the links to glibc-dynamic and musl-dynamic are both pointing to a non-existent page.

**Expected behavior**
Links point to an existing page or repository.

stormqueen1990:
I was able to find [glibc-dynamic on chainguard-images/images](https://github.com/chainguard-images/images/tree/main/images/glibc-dynamic), and there is also a broken link to musl-dynamic in the glibc-dynamic README.

ltagliaferri:
Hi @stormqueen1990 thank you so much for reporting this issue! We updated the link, and the Dockerfile in the tutorial will also need to be updated, which we will do promptly. The team is working on the best catalog fit for each image and we are discussing this feedback to continue to iterate. Again, thank you for bringing this to our attention.

PR #857, opened 2023-07-19
Removing retired images from reference

erikaheidi:
This PR removes retired images from reference pages, namely:

- sdk
- musl-dynamic

It also adds redirect rules to redirect requests to these pages to the main reference list page, and updates the "Getting started with Chainguard Images" guide to remove reference to musl-dynamic [#841]. Autodocs was also updated to ignore the musl-dynamic image since it is still listed by chainctl, but not being updated anymore.


jamonation (approved):
Looks good, I like the strategy of redirecting +1

PR #879, opened 2023-07-27
Update image_specs.md pages for images

jamonation:
## Type of change
documentation

### What should this PR do?
Provides an interim update to images `image_specs.md` files for our various images. 

Some data will be incorrect for images that have switched to componentized variants like say `kyverno-{compoent}-controller`, but those pages are also incorrect on the site today.

So we should be no worse off with this, and the majority of updates are relevant and bring things up to date.

Image specs reference pages should show correct data for most non-componentized images.

Netlify should show image specs pages with up to date data.


jamonation:
I'm not sure why all the image names are getting Uppercased - nothing that I can see in the PHP code that does that.

ltagliaferri (approved):

PR #885, opened 2023-08-03
Update Github rolebinding terraform tutorial

nsmith5:
Folks don't need to log into the platform before getting a role binding anymore so remove those aspects of the tutorial. Also makes it less Enforce-y by calling it the Chainguard platform where possible.

### DIFF
--- a/content/chainguard/chainguard-enforce/iam-groups/rolebinding-terraform-gh/index.md
+++ b/content/chainguard/chainguard-enforce/iam-groups/rolebinding-terraform-gh/index.md
@@ -200,7 +200,7 @@ data "chainguard_identity" "team_ids" {
 
 This block's `for_each` meta-argument iterates through each member of the team. For each iteration, it retrieves that user's GitHub ID and then retrieves a Chainguard identity that it derives using that GitHub ID.
 
-If there are members of the Github team who have not yet registered with Chainguard, this method will still assign them the correct permissions when they log in for the first time.
+If there are members of the GitHub team who have not yet registered with Chainguard, this method will still assign them the correct permissions when they log in for the first time.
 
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-enforce/iam-groups/rolebinding-terraform-gh/index.md
```diff
@@ -201,22 +198,9 @@ data "chainguard_identity" "team_ids" {
...
-  identities_with_ids = [ 
-    for y in data.chainguard_identity.team_ids : try(y.id, null)
-  ]
-  filtered_identities = compact(local.identities_with_ids)
-}
-```
+This block's `for_each` meta-argument iterates through each member of the team. For each iteration, it retrieves that user's GitHub ID and then retrieves a Chainguard identity that it derives using that GitHub ID.
 
-This block creates a variable, `filtered_identities`, by taking the `team_ids` data source and removing any GitHub team members that do not have a Chainguard identity. The final block in this file will use this filtered list of identities to create rolebindings for those team members with existing Chainguard identities tied to their GitHub accounts.
+If there are members of the Github team who have not yet registered with Chainguard, this method will still assign them the correct permissions when they log in for the first time.
```
ltagliaferri, line 203:
```suggestion
If there are members of the GitHub team who have not yet registered with Chainguard, this method will still assign them the correct permissions when they log in for the first time.
```

ltagliaferri (approved):
Thanks, @nsmith5 ! I have some maybe tangential questions...

Should it be "Chainguard Platform" or stay as "Chainguard platform"? 

Do you think we should move this into a "Chainguard" category outside of Enforce and Images? cc @SharpRake

PR #898, opened 2023-08-09
Update rumble-cve-data.yaml

jamonation:
## Type of change
platform

### What should this PR do?
Fixes a commit I somehow reverted, which ordered vulnerability data by severity.

Vulnerabilities should be sorted from critical -> negligible

Github action should run without issues.

A file like https://storage.googleapis.com/chainguard-academy/cve-data/nginx.csv should show vulnerabilities in the correct order.

### DIFF
--- a/.github/workflows/rumble-cve-data.yaml
+++ b/.github/workflows/rumble-cve-data.yaml
@@ -75,7 +75,17 @@ jobs:
             )
             SELECT image, package, vulnerability, version, type, severity FROM ruuuumble
             WHERE DATE(t) BETWEEN DATE_SUB(CURRENT_DATE(), INTERVAL 30 DAY) AND CURRENT_DATE()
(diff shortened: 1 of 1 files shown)


.github/workflows/rumble-cve-data.yaml
```diff
@@ -75,7 +75,17 @@ jobs:
             )
             SELECT image, package, vulnerability, version, type, severity FROM ruuuumble
```
jspeed-meyers, line 76: `ruuuumble` makes me laugh :)

jspeed-meyers (approved):
LGTM.

.github/workflows/rumble-cve-data.yaml
jamonation: Hah, I thought I should put an Easter egg in there for future travelers, makes me smile every time too 😆

PR #900, opened 2023-08-09
Update IDP language

ltagliaferri:
This is a first pass on https://github.com/chainguard-dev/internal/issues/2152, however we are waiting for some more guidance for additional updates


SharpRake (approved):
:+1:

PR #925, opened 2023-08-17
Removing retired images and updating specs pages

erikaheidi:
This PR removes from reference docs three images that have been retired a while back, but are still being listed by `chainctl`:

- `nri-kube-events`
- `nri-kubernetes`
- `nri-prometheus`

In addition to that, the PR also brings a manual update to variants pages for some images that were recently changed.


jamonation (approved):
+1

PR #934, opened 2023-08-21
Updating autodocs workflow

erikaheidi:
Update images autodocs workflow to latest version to fix build issue introduced in the previous release. Now setting docs pages to build via ENV variable, so that the variants page can be generated in a separate workflow.


jamonation (approved):
+1!

PR #949, opened 2023-08-24
Fix overly broad nginx rewrite for /vulnerabilities

jamonation:
## Type of change
bug

### What should this PR do?
Fixes nginx location that was capturing any URL with `/vulnerabilities` in it.


SharpRake (approved):
:+1:

PR #993, opened 2023-09-11
Move Getting Started guides + reorganize menu

erikaheidi:
### What should this PR do?
This PR:

- Moves "Getting Started" image guides to their own section under "Chainguard Images", something we need to do in order to change the image reference pages to use tabs.
- Moves "Registry" section to the top-level "Chainguard Products" menu
- Reorganizes menu for better readability and easy access to content
- Adds redirects to "registry" section and individual tutorials that were relocated
- Adds a missing redirect to the "Network requirements" section that has been recently moved

Preview: https://deploy-preview-993--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/getting-started/getting-started-postgres/

This is how the new menu should look like after the changes:
Screenshot from 2023-09-11 18-18-14


ltagliaferri (approved):
Thanks, Erika! I don't think you need to do this in this PR, but what do you think if we changed the `linkTitle` to be the name of the image rather than the whole long "Getting Started with ...." title?

Screenshot 2023-09-11 at 12 58 08 PM

erikaheidi:
> Thanks, Erika! I don't think you need to do this in this PR, but what do you think if we changed the `linkTitle` to be the name of the image rather than the whole long "Getting Started with ...." title?
> 
> Screenshot 2023-09-11 at 12 58 08 PM

That makes sense! I was unsure about this one. Let me do it now already :)

erikaheidi:
Updated, It's much better now! 
Screenshot from 2023-09-11 19-16-02

PR #995, opened 2023-09-12
Updating image reference docs with new menu organization

erikaheidi:
### What should this PR do?
This manual PR uses content generated from the new autodocs, and updates the image reference docs to the new menu structure. The change has a dramatic impact on build time and memory consumption when building the website. Images now use tabs to organize the contents.

Includes an Nginx redirect (🤞🏼):

```
"~^/chainguard/chainguard-images/reference/(.+)?/overview/(.+)?$" /chainguard/chainguard-images/reference/$1/;
```

Following, I will disable the current workflow running in this repository and enable the new autodocs workflow in the [images-autodocs](https://github.com/chainguard-dev/images-autodocs) repository, which will continue creating pull requests here on a daily basis.

Use the [preview link](https://deploy-preview-995--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/reference/apko/) to check the left menu and image reference pages.


jamonation (approved):
This is amazing @erikaheidi 😲  - so much cleaner, faster, smaller! 50+s down to 10s to build the site from scratch locally for me.

My only suggestion with the `images-autodocs` generation side of things would be to add an h2 or h3 heading to the tags page. It is a minor thing, but without a heading, the `toc: true` setting doesn't get triggered, so there's no right-nav. In itself, not a big deal, but visually when flipping between tabs it makes the content move around. Maybe another PR after this one?

erikaheidi:
Thanks @jamonation ! That makes total sense, I was also annoyed by the "jump" of content when moving into that page. I will investigate adding an additional tags table for arm architecture tags, and add the headers. Merging this now as a starting point.

PR #1003, opened 2023-09-13
Adding redirect for gcc-musl

erikaheidi:
### What should this PR do?
Adds an Ngnix redirect rule to the retired `gcc-musl` image.


jamonation (approved):
+1

PR #1030, opened 2023-10-01
Update rumble chart theme with light/dark colours

jamonation:
## Type of change
platform

### What should this PR do?
This updates the theme colours for Rumble vulnerability comparison charts depending on light/dark mode.

There's also a function to fetch wolfi advisory data mixed in here, but it is not used yet.

Visual consistency with the rest of our site theme.

The screenshots here show light and dark modes.
Screenshot 2023-10-01 at 12 24 39 PM
Screenshot 2023-10-01 at 12 24 45 PM

Testing locally should work, netlify won't. Toggling light/dark needs a page reload to get the chart to pick up on the change.


static/rumble/bar/index.html
```diff
@@ -5,9 +5,9 @@
     <link rel="icon" type="image/svg+xml" href="/inky-logo.svg" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Visualization</title>
-    <script type="module" crossorigin src="../assets/bar.ab469926.js"></script>
-    <link rel="modulepreload" crossorigin href="../assets/RumbleLoader.c6d6c914.js">
-    <link rel="modulepreload" crossorigin href="../assets/Grid.be76391d.js">
+    <script type="module" crossorigin src="../assets/bar.ae60c53a.js"></script>
```
ltagliaferri, line 8: I know we had relative links before, but can these be made to be absolute links, like in `static/rumble/blog/index.html` below?

ltagliaferri (approved):
Checked out locally and looks good, just one nit

static/rumble/bar/index.html
jamonation: That's a good catch, I updated it to be `/rumble/assets/...` for these. I still don't quite know how to make the upstream `visualization-ui` React app use a subdirectory, but when I figure that out it will be easier to make these urls consistent.

PR #1044, opened 2023-10-06
Fix github actions email.

wlynch:
Yay opaque magic values. ✨

See https://github.com/wlynch/test/commit/7f6aed848e3be2e7a6362b0f8bb623408ae860ed for an example.

## Type of change

Automation

### What should this PR do?

Fixes GHA email so it doesn't say invalid-email-address

image


jamonation (approved):
How you found that, I'm in awe. Thanks for spotting the issue!

PR #1046, opened 2023-10-06
Add missing release notes from v0.1.179

wlynch:
## Type of change

Docs

### What should this PR do?

Updates missing release notes.


jamonation (approved):
Something is up with the github action to autogenerate these, so thanks for making this PR. I'll keep investigating the issue, but I'm pretty sure the next release should bring things back in line.

PR #1060, opened 2023-10-11
Moving 'Considerations for Keeping Images Up to Date' to Images bucket

SharpRake:
## Type of change

This is just a quick PR to move the conceptual article I published last week from the "software security" bucket to "images".

I didn't make any significant changes to the doc so just an eye that it's in the right place would be good.


ltagliaferri (changes requested):
Can you update the nginx config for the redirect too, please?

ltagliaferri (approved):

PR #1078, opened 2023-10-18
[AutoDocs] Update Images Reference Docs

autodocsbot:
Updated image reference docs.


erikaheidi:
This didn't bring the updated provenance page, so I'll bump the action and the nightly build will pick up the change.

PR #1087, opened 2023-10-24
Platform bucket

### LINKED ISSUE #1033: Plan for moving + updating IAM docs so they're less focused on Enforce
**What topic are you requesting a resource about?**
* **Chainguard product**
* Open source related
* Conceptual security related
* Other (please describe)

**Proposed title:** 
n/a

**Description:**
IAM management is one of Chainguard's secret sauces. Right now our IAM docs are all housed under Enforce, but IAM is increasingly relevant to the Images product. 

Questions to answer:
* How hard would it be to update our IAM docs so they're less focused on Enforce
* Would it make sense to move these docs to under Images? If not, how can we separate them out from Enforce?

SharpRake:
## Type of change

This change pulls some resources out from the Enforce bucket and moves them to a new Administration bucket. When possible, I removed references to Enforce to make these docs more useful to Images users. Note that this also removes "Chainguard" from the sidebar menu options (i.e., "Chainguard Images" ==> "Images")

### What should this PR do?

Resolves #1033 

To make these resources more discoverable and useful outside of Enforce-specific contexts.

@amdawson I'd appreciate a :+1:  or :-1:  from you on all this. 

I'm still open to any general feedback, particularly with how things are organized.


SharpRake:
One weird quirk in all this is the CloudEvents docs. These docs are pretty strictly tied to Enforce - to the point that they can't really be genericized - but I understand there's a use case for people to use CloudEvents with Images. 

I think one option would be to move the CloudEvents docs back to the Enforce bucket, then create a couple equivalent ones that will live under Images. Maybe then we could have like a CloudEvents overview doc that lives under the Administration bucket? But not sure if that's overcomplicating things.

ltagliaferri (approved):
This looks great, thank you!

PR #1101, opened 2023-10-31
More nginx redirect tweaks

SharpRake:
## Type of change

Having trouble getting the nginx redirects just right, I think this fixes what I'm most concerned about.


SharpRake:
I just tested this locally and I'm sure these redirects work as expected.

ltagliaferri (approved):

PR #1110, opened 2023-11-03
Removing files not used anymore by autodocs

erikaheidi:
This PR removes old templates that are not used anymore by autodocs, to avoid confusion - the templates being used now live in the same repository as the autodocs app: [chainguard-dev/images-autodocs](https://github.com/chainguard-dev/images-autodocs).  The `last-update.md` file is also not used anymore to generate the changelog.


jamonation (approved):
makes sense to me, +1

PR #1115, opened 2023-11-07
moving Events Reference doc to Administration bucket from Enforce

SharpRake:
## Type of change

Moving the Events Reference doc to the new Admin bucket, and removing its current Enforce references. 

### What should this PR do?

should correctly minimize Enforce references without losing meaning, and move the Events article to the Admin bucket while creating appropriate redirect.

So this doc resides closer to our other events docs and is more easily discoverable.

The content should still make sense as written. To be clear, I took a quick and dirty approach to removing Enforce references, often just being changing something like "the Enforce event" to "the Chainguard event." There are a few cases where the article is directly referring to Enforce via an example so I couldn't avoid those cases, but if those could be easily switched to using an example revolving around Images I'd be all ears. 

Please review the updated content (https://deploy-preview-1115--ornate-narwhal-088216.netlify.app/chainguard/administration/cloudevents/events/) and let me know if any Images-focused examples could be added. 

Thanks @imjasonh for agreeing to take a gander at this!


SharpRake:
Pausing this as a reviewable PR - @jamonation let me know that this doc is built from a template file that I need to update. In the meantime, I'm leaving this PR up to use the deploy preview as an example of an updated template that folks can :+1: or :-1:

PR #1123, opened 2023-11-08
Add Makefile and apko config for a local dev-container environment

jamonation:
This is a bit of an experiment that I hope we can turn into a daily tool for Academy work with various supporting tools. To start it adds a Makefile with two targets, `apko-build` and `dev-container`.

1. `apko-build`: creates a local image for Academy development, using our `cgr.dev/chainguard/apko` image and the included `apko.yaml` file in this PR.
2. `dev-container`: runs the image with various volumes mounted into it, and ports forwarded with a copy of Hugo and nginx to start with.

Future versions can add things like autodocs, rumble code, chainlink and so on.

To test, run:
```
make apko-build
make dev-container
```

Once it builds and runs, use `hugo serve --bind 0.0.0.0` and then visit http://localhost:1313 when it is ready.

To build and preview HTML files with the included nginx, run `hugo -e local-dev`, then visit http://localhost:8080.

Docker and `make` are the only prerequisites to run this I think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,6 @@ arch := $(shell uname -m)
 .PHONY: apko-build dev-container
 apko-build:
 	docker run --rm -v $$PWD:/work -it cgr.dev/chainguard/apko:latest build \
-	--repository-append https://packages.wolfi.dev/os \
-	--keyring-append    https://packages.wolfi.dev/os/wolfi-signing.rsa.pub \
-	--package-append    wolfi-baselayout \
 	--arch              $(arch) \
 	/work/$(cfg) academy.local /work/academy.tar
 	docker load < academy.tar
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ apko-build:
 	rm academy.tar
 
 dev-container:
+ifeq ("$(arch)","x86_64")
 	docker run --rm \
 	-v $$PWD/config/entrypoint.sh:/entrypoint.sh \
 	-v $$PWD/public:/usr/share/nginx/html \
@@ -17,4 +18,14 @@ dev-container:
 	-v $$PWD:/home/inky/ \
 	-p 8080:8080 -p 1313:1313 \
 	-it --user root \
-	academy.local:latest-$(arch)
+	academy.local:latest-amd64
+else
+	docker run --rm \
+	-v $$PWD/config/entrypoint.sh:/entrypoint.sh \
+	-v $$PWD/public:/usr/share/nginx/html \
+	-v $$PWD/nginx.conf:/etc/nginx/nginx.conf \
+	-v $$PWD:/home/inky/ \
+	-p 8080:8080 -p 1313:1313 \
+	-it --user root \
+	academy.local:latest-$arch
+endif
--- a/Makefile
+++ b/Makefile
@@ -13,4 +13,11 @@ apko-build:
 	rm academy.tar
 
 dev-container:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,16 @@
...
+	--repository-append https://packages.wolfi.dev/os \
+	--keyring-append    https://packages.wolfi.dev/os/wolfi-signing.rsa.pub \
+	--package-append    wolfi-baselayout \
+	--arch              $(arch) \
+	/work/$(cfg) academy.local /work/academy.tar
+	docker load < academy.tar
+	rm academy.tar
+
+dev-container:
+	docker run --rm -v $$PWD/config/entrypoint.sh:/entrypoint.sh -v $$PWD/public:/usr/share/nginx/html -v $$PWD/nginx.conf:/etc/nginx/nginx.conf -v $$PWD:/home/inky/ -p 8080:8080 -p 1313:1313 -it --user root apko.local:latest-$(arch)
```
jamonation, line 16: Split this up

jamonation, line 16:
```suggestion
	docker run --rm \
	-v $$PWD/config/entrypoint.sh:/entrypoint.sh \
	-v $$PWD/public:/usr/share/nginx/html \
	-v $$PWD/nginx.conf:/etc/nginx/nginx.conf \
	-v $$PWD:/home/inky/ \
	-p 8080:8080 -p 1313:1313 \
	-it --user root \
	apko.local:latest-$(arch)
```

apko.yaml
```diff
@@ -0,0 +1,52 @@
+contents:
+  repositories:
+    - https://packages.wolfi.dev/os
```
imjasonh, line 3: nit: you don't need this if you pass it in `--repository-append`

imjasonh (approved):

apko.yaml
jamonation: Oh good catch, will remove

jamonation: So I flipped this around and added the key and repo to `apko.yaml` - this way building directly with `apko` also works:
```
apko build --arch host apko.yaml academy.local academy.tar
```

erikaheidi:
I got an error when trying to run the `make dev-container` command. It seems there is some platform mismatch:

```
INFO[0007] Final index tgz at: /work/academy.tar        
docker load < academy.tar
aa6782e0dc57: Loading layer [==================================================>]  86.15MB/86.15MB
Loaded image: academy.local:latest-amd64
rm academy.tar
rm: remove write-protected regular file 'academy.tar'? y
❯ make dev-container
docker run --rm \
-v $PWD/config/entrypoint.sh:/entrypoint.sh \
-v $PWD/public:/usr/share/nginx/html \
-v $PWD/nginx.conf:/etc/nginx/nginx.conf \
-v $PWD:/home/inky/ \
-p 8080:8080 -p 1313:1313 \
-it --user root \
academy.local:latest-x86_64
Unable to find image 'academy.local:latest-x86_64' locally
docker: Error response from daemon: pull access denied for academy.local, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
make: *** [Makefile:13: dev-container] Error 125
```

jamonation:
Ahh thanks for checking @erikaheidi - seems like the `uname -m` output is doing something strange with both `amd64` and `x86_64`. I'll try to figure out how to set that properly.

erikaheidi (approved):
it works now! great work Jamon 👏🏼

PR #1136, opened 2023-11-14
fixing nginx redirect for network reqs

SharpRake:
## Type of change

Forgot to add a redirect for the network requirements doc when I moved it to the Administration bucket. This PR corrects that.


SharpRake:
redirect was actually added and works correctly, closing this PR

PR #1140, opened 2023-11-15
fixing actions doc to finalize events reference move

SharpRake:
## Type of change

There was a block I missed in the `actions.yaml` file that was causing some issues with the autodocs generation. This PR fixes that block.

I also merged the last Enforce autodocs PR which used the updated Events reference template, but that had some residual Git data in the doc. I submitted a PR (approved and merged) to fix the template, but this PR also edits that file just to clean it up until the next autodocs workflow run.

Cleaning up autodocs


jamonation (approved):
LGTM, I think you've found all of the dusty undocumented corners now!

PR #1183, opened 2023-11-30
removing outdated Image Reference docs

SharpRake:
## Type of change

I noticed that the `external-attacher` image has [been withdrawn](https://chainguard-dev.slack.com/archives/C05GYUBM07Q/p1697246551578519) and replaced by `kubernetes-csi-external-attacher`. I also noticed that the `external-resizer` image is no longer in the Images repo.

I'm guessing that these docs have been hanging around for a little while. @erikaheidi is there a step in the Autodocs setup that removes Reference docs after the Image has been withdrawn? 

We don't want folks to think we're offering Images we don't really have, eh?

I tagged Erika as a reviewer specifically, because if I'm wrong here and these Reference docs need to stay up, I think she'd be the one to know.


SharpRake:
also noticed that `oidc-discovery-provider` has been replaced by `spire-oidc-discovery-provider`

erikaheidi (changes requested):
Hi Mark! Thank you for this find, currently Autodocs doesn't have a mechanism for detecting and auto-removing this, also because some retired images remain being listed with chainctl and we need to add them to an ignore list. I have [an issue](https://github.com/chainguard-dev/internal/issues/3329) to implement a command to facilitate doing that in autodocs but it's not yet in progress.

Everything looks good, the only thing missing is that we need to also add a redirect in Nginx (that's how we did it in the past) to make sure we don't get 404s. This for reference: https://github.com/chainguard-dev/edu/blob/main/nginx.conf#L61

Thank you again! I am going to add these to the ignore list in autodocs to make sure it doesn't bring it all back again.

erikaheidi (approved):

PR #1191, opened 2023-12-05
update gitlab identity docs to list image repos

imjasonh:
## Type of change

Refocus docs toward images

### What should this PR do?

Refocus GitLab assumable identity doc toward an Images use case, instead of a policy-based use case.

We expect users to be more interested in examples focused on images

The example code works as expected

By creating the resources and running the GitLab workflow to show expected output.

### DIFF
--- a/content/chainguard/chainguard-enforce/authentication/identity-examples/enforce-gitlab-identity.md
+++ b/content/chainguard/chainguard-enforce/authentication/identity-examples/enforce-gitlab-identity.md
@@ -45,38 +45,19 @@ The first file, which we will call `main.tf`, will serve as the scaffolding for 
 
 The file will consist of the following content.
 
-```
+```hcl
 terraform {
   required_providers {
     chainguard = {
-      source = "chainguard/chainguard"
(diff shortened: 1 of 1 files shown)


ltagliaferri:
Hi @imjasonh , where should we move this doc to, considering the focus on Images?

imjasonh:
> Hi @imjasonh , where should we move this doc to, considering the focus on Images?

I think we should move it to some general-Admin-oriented docs area. Like docs about groups or IAM, which is shared by Enforce and Images.

ltagliaferri:
> I think we should move it to some general-Admin-oriented docs area. Like docs about groups or IAM, which is shared by Enforce and Images.

They're currently under Enforce > Authentication > Identity Examples

We have a top-level Administration > IAM & Groups --- should we wait until all the identity examples are updated to Images and then move them up to the Admin > IAM level? Or move this for now as a standalone?

ltagliaferri (approved):
If it's all right, I'll merge this for now and then batch move them over when we have critical mass

PR #1196, opened 2023-12-06
update jenkins identity docs to list image repos

imjasonh:
Previously, https://github.com/chainguard-dev/edu/pull/1195


jamonation (approved):
+1, thanks for powering through these

PR #1206, opened 2023-12-08
moving the assumable identity docs to admin bucket

### LINKED ISSUE #1203: Move Assumable Identity Examples
Now that we've refocused the Assumable Identity example docs more towards Images we can move them over to the `Administration` bucket.

SharpRake:
## Type of change

this update removes any final references to Enforce from the assumable identity docs. It also moves these docs to the IAM & Groups folder in the Administration bucket. I did some minor rearranging in these folders and set up the necessary redirects (to the best of my knowledge). 

### What should this PR do?

Resolves #1203 

Enforce is winding down but these docs are relevant for Images. 

Ensure that changes make sense. I did some wordsmithing here and there to clarify some points and minimize Enforce concepts.

Review copy changes and please double check that the redirects are correct.


ltagliaferri (approved):
Working well locally, thank you!

PR #1207, opened 2023-12-08
Update python getting started guide.

amouat:
## Type of change
Update to Python Getting Started

Closes https://github.com/chainguard-dev/internal/issues/3425

### What should this PR do?
Main change is to use venv to install and copy dependencies, which frees us from specifying the python version in the path.

To improve the experience for users and make it less likely they will run into problems

Ideally test the Dockerfile and check the english.

As above, if you could try running through the steps and flag any issues or wording problems.

### DIFF
--- a/content/chainguard/chainguard-images/getting-started/getting-started-python.md
+++ b/content/chainguard/chainguard-images/getting-started/getting-started-python.md
@@ -148,7 +148,9 @@ You have successfully completed the single-stage Python Chainguard Image. At thi
 
 ## Example 2 — Multi-Stage Build for Python Chainguard Image
 
-In this example, we'll build and run a multi-stage Python Chainguard Image to demonstrate the process that includes a build image with pip and a shell, and a final distroless image.
+In this example, we'll build and run a multi-stage Python Chainguard Image. We'll have a build image
+that includes pip and a shell before creating a final distroless image without these development
+tools for production.
 
 ### Step 1: Setting up a Demo Application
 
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-images/getting-started/getting-started-python.md
```diff
@@ -148,7 +148,9 @@ You have successfully completed the single-stage Python Chainguard Image. At thi
 
 ## Example 2 — Multi-Stage Build for Python Chainguard Image
 
-In this example, we'll build and run a multi-stage Python Chainguard Image to demonstrate the process that includes a build image with pip and a shell, and a final distroless image.
+In this example, we'll build and run a multi-stage Python Chainguard Image. We'll have a build image
+that includes pip and a shell before creating a final distroless image without these development
```
SharpRake, line 152: sometimes paragraphs get split onto multiple lines like this, but it doesn't seem to matter on the front end? If I'm being a stickler I'd say remove the line breaks but nbd.

SharpRake (approved):
Worked like a charm!

content/chainguard/chainguard-images/getting-started/getting-started-python.md
amouat: I'd say everything should be formatted to 80-100 columns or so :)

Having paragraphs all on one line makes it difficult to edit especially in line-based editors like Vim.

That being said, I realise I'm the only one using Vim so we should go with whatever the majority wants.

amouat: If all the other docs use long-lines, let's change this one.

ltagliaferri: I don't think it's a big deal either way, we can merge it

PR #1210, opened 2023-12-09
Updates to Custom IDP docs + more!

### LINKED ISSUE #1057: Update docs for configuring custom IDP and IAM and Security
Custom identity providers now require users configure a default role to grant to users when first logging in if they have not yet registered through that IDP. 

The user is bound to the role at the group where the custom IDP is defined. Typically this is the organization/root level. The default role can be a built in role (e.g. `viewer`) or a custom role defined in the same group or in an ancestor group.

A new flag, `--default-role` has been added to `chainctl iam identity-provider create` and `update`. For `create`, if the user omits the flag they will be asked to select a role interactively. Roles can be referenced by either ID or name. 

Users cannot configure a default role that has more capabilities than they do.

It also may be worth adding that users can set their verified org name with a single command, similar to setting `default.identity-provider`:

`chainctl config set default.org-name <org name>`

### LINKED ISSUE #1209: Fix advanced usage shortcode
**Describe the bug**
The `images-advanced.html` shortcode has some bad formatting in it. It doesn't render standaloen Markdown, 

**To Reproduce**
1. Go to  https://edu.chainguard.dev/chainguard/chainguard-images/getting-started/getting-started-postgres/#advanced-usage
2. Note blcok isn't rendering, is it

**Expected behavior**
It should render!

**Screenshots**
Screenshot from 2023-12-08 15-40-10

 - OS: [e.g. iOS] Ubuntu
 - Browser [e.g. chrome, safari] Firefox
 - Version [e.g. 22] 119.0

**Additional context**
I've got a fix coming!

SharpRake:
## Type of change

This PR adds some details around the now-required `--default-role` option. It also makes the command at the end of the Custom IDP example docs fully declarative and cleans up the Introduction to Custom IDPs doc a bit.

As a bonus this also includes some general improvement fixes, including
* a shortcode that I noticed isn't rendering properly in the Getting started Guides
* some misconfigured front matter in the getting started guides

### What should this PR do?

Resolves #1057
Resolves #1209 

I felt like I was going a little crazy trying to copy changes across the three Custom IDP examples. I would say these don't need a full tech test, as the only change is some details around the final command. Mostly just an eye that the changes across those three are consistent, and that the new copy overall makes sense / is clear.

### DIFF
--- a/content/chainguard/administration/custom-idps/azure-ad/index.md
+++ b/content/chainguard/administration/custom-idps/azure-ad/index.md
@@ -65,7 +65,7 @@ Finally, take note of the client secret “Value” that is created. You'll need
 
 Now that your Azure Active Directory application is ready, you can create the custom identity provider.
 
-First, log in to Chainguard with `chainctl`, using an OIDC provider like Google, Github, or GitLab to bootstrap your account.
+First, log in to Chainguard with `chainctl`, using an OIDC provider like Google, GitHub, or GitLab to bootstrap your account.
 
 ```sh
 chainctl auth login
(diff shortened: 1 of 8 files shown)


ltagliaferri (approved):
Just little nits to fix

PR #1241, opened 2023-12-13
Remove true colour setting.

amouat:
Don't use true colour in Python example as it breaks on some terminals.

## Type of change
Minor fix to python code example.

### What should this PR do?
Simplifies python code and ensures compatibility.

Make sure our docs works for everyone!

New code works for everyone!

If you can run the python code and make sure the output looks ok, that would be appreciated.


jamonation (approved):
Looks good, works for me without those extra modules.

PR #1249, opened 2023-12-18
Add images logos

erikaheidi:
This PR adds the image logos to the Images Reference Page. The content is now centralized.

For this to work as-is for all images, there is an assumption that all images have a logo.svg file available in our drive, at the location `https://storage.googleapis.com/chainguard-academy/logos/{{ .Params.linktitle }}/logo.svg`.

Preview: https://deploy-preview-1249--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/reference/

The caveat is that the logos must be synced to google cloud automatically to avoid broken images, as this is infused directly into layouts. I have tried several different tricks to load a default image when the logo is broken, but the ones that worked made the page extremely slow. I will leave this optimization for the experts :)

### DIFF
--- a/layouts/article/list-images.html
+++ b/layouts/article/list-images.html
@@ -23,7 +23,7 @@
                                         <div class="card flex-fill">
                                             <div class="card-body d-flex flex-column">
                                                 <h5 class="card-title mt-0">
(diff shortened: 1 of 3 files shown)


SharpRake (approved):
Looks good to me!

PR #1273, opened 2024-01-04
Federate GCP auth right before it's used.

mattmoor:
This reduces the window where the GCP token is exposed.


mattmoor:
cc @jamonation

jamonation (approved):

PR #1276, opened 2024-01-05
Fix minor path issue in command.

amouat:
## Type of change
Fix command in tutorial.

### What should this PR do?
Fixes apko command so it builds.

Tutorials should change

Please check the new command works

Run through the tutorial steps


amouat:
Note this was found by a user: https://github.com/chainguard-dev/melange/issues/902

jamonation (approved):
Nice catch!

amouat:
I missed the path to the key, fixed that as well now

amouat:
Actually there's a better way to fix this...

jamonation (approved):
Sounds like this is the proper fix to match the tutorial so +1

PR #1295, opened 2024-01-14
Audit the EDU workflows for permissions.

mattmoor:
Ensure all workflows have `permissions:` blocks, and drop some unnecessary permissions from a number of them that seem to have been copied around. Document why each permission is necessary.

This also eliminates a number of top-level `env` blocks that held secrets, since generally putting secrets into environment variables is a bad practice.

In certain cases where the environment variable was necessary (e.g. using `GH_TOKEN` to pass creds to `gh`) a much more narrowly scoped environment variable is now used.


.github/workflows/add-issues-to-project-platform.yaml
mattmoor, line 1: Pretty sure these workflows are defunct ~everywhere, so cleaning it up.

jedsalazar (approved):

jamonation (approved):

PR #1326, opened 2024-01-29
Refactor rumble to generate all csv and json files using insights data

jamonation:
## Type of change
rumble improvement

### What should this PR do?
This PR cuts rumble over to the insights database. It switches the nightly CSV and JSON generation for vulnerability comparison pages to use a single Go binary, instead of using 3 different GitHub actions.

Insights is the authoritative source for image scans now, and the GitHub actions rely on using gcloud commands, which were failing and leaking auth details in unexpected ways.

The actions in this PR should run and generate files.

This PR should trigger the actions. But running the tool locally should work too:

```
go run main.go legacy-csv --project prod-images-c6e5  --db insights_ds --gcs-project chainguard-academy --bucket chainguard-academy 
go run main.go vulns --project prod-images-c6e5  --db insights_ds --gcs-project chainguard-academy --bucket chainguard-academy
go run main.go image-csv --project prod-images-c6e5  --db insights_ds --gcs-project chainguard-academy --bucket chainguard-academy --rumble-json-path ../../data/rumble.json

### CODE REVIEW

**Comment by jamonation** [comment]
This seems to work! The new version runs successfully over here to generate all the CSV and JSON files in one GitHub action: https://github.com/chainguard-dev/edu/actions/runs/7698817821/job/20978995889

**Review by erikaheidi — APPROVED** [review]
I can't provide a thorough code review since I'm not familiar with Go, but the workflow looks good to me, the job run successfully and from the output everything seems to be tight 👍🏼

File: .github/workflows/rumble-vulnerability-data.yaml [hunk 1]
```diff
@@ -4,44 +4,69 @@ on:
...
 
 jobs:
-  generate-vulnerability-json:
+  generate-csvs-and-json:
     runs-on: ubuntu-latest
 
     permissions:
-      id-token: write # federate with GCP
-
+      contents: read
```
jedsalazar, line 25: Can we retain and document the permissions in the workflows?

.github/workflows/rumble-vulnerability-data.yaml
```diff
@@ -4,44 +4,69 @@ on:
     - cron: "1 5 * * *"
   workflow_dispatch:
   push:
-    branches: [rumble-vulnerability-data]
+    branches: [rumble-insights]
+
+env:
```
jedsalazar, line 9: If these secrets are scoped to this workflow, any chance we can leverage [environment secrets](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#environment-secrets) instead of environment variables here?

jedsalazar (commented):
A nit and open question on the Actions workflows. Thanks for putting this together!

.github/workflows/rumble-vulnerability-data.yaml
jamonation: Ah yeah, no idea where that got lost in the shuffle, will add it back for sure

.github/workflows/rumble-vulnerability-data.yaml
jamonation: I think so! We don't have environments set up for Academy yet, but it makes sense to move to using them.

PR #1342, opened 2024-02-04
Set up a DNS zone and GCLB for edu.

mattmoor:
The DNS zone itself and the GCLB will be dormant until we make some changes to the Chainguard DNS records, here's my suggestion:
1. We roll this out,

2. We mirror the `A` record that the serverless GCLB module created into the Chainguard records (dropping the CNAME record),

_(this will incur a small 🤞 amount of downtime while the TLS certificate provisions, in my experience these can take up to 10m)_

3. We set up `NS` records at the `chainguard.dev` level to delegate `edu.chainguard.dev`'s nameservers to the zone we provisioned here (preserving the `A` record while it rolls out),

_(wait 24h for the nameserver rollout to take effect)_

4. Clean up the `A` record at the `chainguard.dev` level (leaving just the `NS` record).

At each step along the way, I believe the rollback procedure should be relatively quick if things are broken.  With the caveat that we expect it to take a bit of time for the HTTP01 challenge to complete after step 2.

### DIFF
--- a/iac/main.tf
+++ b/iac/main.tf
@@ -10,7 +10,7 @@ provider "google-beta" { project = var.project_id }
 
 module "networking" {
   source  = "chainguard-dev/common/infra//modules/networking"
-  version = "0.2.0"
+  version = "0.4.6"
 
   name       = var.name
   project_id = var.project_id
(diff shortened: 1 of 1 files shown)


jamonation (approved):
Makes sense, this gets a step closer to cutting over to the newer stack

Issue #1369, opened 2024-02-13
Add nginx Cache-Control headers to help with sidebar refreshing


jamonation:
@pdeslaur noticed that there are no cache-control headers in the nginx container, which leads to unsightly sidebar reloads while navigating around the site. When adding this, also include a `stale-while-revalidate` option in the header.

I'm thinking 300s is a long enough `max-age` value to start with, and the same for the revalidate option.

PR #1374, opened 2024-02-15
Add cache-control header to nginx for html files

jamonation:
## Type of change
feature

### What should this PR do?
refs #1369 

This adds cache-control headers to html pages that should help with sidebar loads. The initial value is 300s, so a page will be refreshed after 5 minutes.

Shouldn't be anything visible to a user.

Won't work on netlify, staging looks like this:
```
curl -sI --user ... https://<staging_url>/chainguard/chainguard-images/fips-images/ |grep cache-control
cache-control: public, max-age=300, stale-while-revalidate=300
```

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -99,6 +99,10 @@ http {
             rewrite ^ /vulnerabilities/index.html break;
         }
 
+        location ~* /(\index.html)?$ {
+            add_header Cache-Control "public, max-age=300, stale-while-revalidate=300";
+        }
+
 	# use hugo's built in 404 page for now
 	error_page 404 /404.html;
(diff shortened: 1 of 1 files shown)


SharpRake (approved):
Honestly I was having a hard time confirming that this change is doing what it's supposed to, but I can at least confirm that it doesn't break anything!

jamonation:
> Honestly I was having a hard time confirming that this change is doing what it's supposed to, but I can at least confirm that it doesn't break anything!

It should be an invisible change - our deployments take longer than the cache max age so it shouldn't really be noticed at all. Thanks for checking!

PR #1389, opened 2024-02-23
Revert https://github.com/chainguard-dev/edu/pull/1358

mattmoor:
The first commit is a pure revert, the second is the fix.

I forgot there was a breaking change in our networking module, and so this adds the `netnum_offset = 1` to preserve the old IP spaces for the subsets to avoid triggering replacements.


jamonation (approved):
+1, reviewed on a call together

PR #1394, opened 2024-02-27
Add note about grype version results in diff api doc

jamonation:
## Type of change
docs

### What should this PR do?
Add a note about grype versions used to scan images to the diff api page

Help independently verify grype results based on diff api output.

Running through the [Using the Diff API](https://edu.chainguard.dev/chainguard/chainguard-images/using-the-image-diff-api/) page and checking output matches would be good, but this is mainly an informational change.


SharpRake (approved):
awesome, thanks so much Jamon!

PR #1467, opened 2024-03-22
IAM Updates - batch 1

SharpRake:
## Type of change

This is the first batch of updates relating to recent IAM changes. Any references to IAM groups were changed to organizations and I renamed the `IAM & Groups` bucket to `IAM & Organizations`. I also added a directory redirect to reflect this change.

The most significant changes happened in the Overview doc where I added a section outlining what organizations are and deleted a reference to Gulfstream.

Otherwise I included changes that I caught in my editing process, mostly spelling errors.

### What should this PR do?

Does not resolve anything entirely

Reflecting changes to IAM structure + commands

Please give all the changes a close look. 

Note that `npm run start` wasn't working for me (increasing verbosity and checking the logs weren't helpful?) so I fear there may be some outputs that have rendered poorly. I'll fix these in another commit if I can, assuming the deploy preview works correctly. 

EDIT: second commit should fix  these, feel free to flag any bad looking outputs or tables.


erikaheidi (approved):
It all looks good to me! I run most commands from the new guide, and checked the formatting, I couldn't catch any issue. great updates, thanks Mark!

PR #1475, opened 2024-03-26
Update autodocs-enforce.yaml

jamonation:
## Type of change
bug

### What should this PR do?
This adds the octo-sts provider to academy so that the workflow action can create a PR with docs changes. I don't know if I'm using it correctly though.

The autodocs-enforce.yaml workflow is failing because it doesn't have permissions to create a PR.

### DIFF
--- a/.github/chainguard/edu.sts.yaml
+++ b/.github/chainguard/edu.sts.yaml
@@ -1,0 +1,8 @@
+issuer: https://token.actions.githubusercontent.com
+subject: repo:chainguard-dev/edu:ref:refs/heads/platform-docs
+claim_pattern:
+  job_workflow_ref: chainguard-dev/edu/.github/workflows/autodocs-platform.yaml@refs/heads/platform-docs
+
+permissions:
+  contents: write
+  pull_requests: write
--- a/.github/workflows/autodocs-platform.yaml
+++ b/.github/workflows/autodocs-platform.yaml
@@ -1,12 +1,12 @@
-name: Build Enforce Docs
+name: Build Platform Docs
 
 on:
   schedule:
     - cron: "0 * * * *"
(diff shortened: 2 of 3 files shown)


.github/workflows/autodocs-enforce.yaml
```diff
@@ -103,6 +103,13 @@ jobs:
     - name: npm run build
       run: npm run build
 
+    - name: Set up Octo-STS
+      uses: chainguard-dev/octo-sts-action@6177b4481c00308b3839969c3eca88c96a91775f # v1.0.0
+      id: octo-sts
+      with:
+        scope: chainguard-dev/edu
+        identity: edu
```
jdolitsky, line 111: Im not sure the proper thing to do here, but I will share what I know about octo-sts..

The `identity` should match a file name in the `scope` repo, under `.github/chainguard`: https://github.com/chainguard-dev/edu/tree/main/.github/chainguard

So the fact that there isn't a `edu.sts.yaml` there means this would not work

jdolitsky (commented):
For opening PRs etc, try following the usage of digestabot here:

- https://github.com/chainguard-images/images/blob/main/.github/workflows/digestabot.yaml#L28
- https://github.com/chainguard-images/images/blob/main/.github/chainguard/digestabot.sts.yaml

Note the `pull_requests: write` part in the last file

.github/workflows/autodocs-enforce.yaml
jamonation: Makes sense, thanks for explaining how this works. I've updated this PR with an `edu.sts.yaml` file

jdolitsky (approved):

erikaheidi (approved):

PR #1493, opened 2024-04-03
IAM docs update, focus on roles and role-bindings

SharpRake:
## Type of change

The main goal of this PR is to add a doc outlining what roles and role-bindings are within the context of Chainguard. As part of this, I pulled the "role bindings" section (actually focused on roles, not role bindings) from the IAM overview and reworked it to be more correct, and also expanded on it to flesh out the overview. 

I also added a new section about identities to the IAM overview. I also thought it made sense to remove "enforce" from the IAM model overview doc's URL and add a redirect

Lastly, I did a code search to change any instance of "rolebinding" or "role binding" to "role-binding".

### What should this PR do?

Resolves https://github.com/chainguard-dev/internal/issues/3548

Improving our IAM documentation

It should read well, changes should make sense, and the commands in the new doc should work!

Copyedit and tech test, please!


sheesh:
All the copy looks good! For some reason, I don't have permission to manage roles so I can't check all the commands. Will try to get my permissions fixed, and see if another team member can check the commands in the meantime.

content/chainguard/administration/iam-organizations/roles-role-bindings.md
```diff
@@ -0,0 +1,94 @@
...
+```sh
+chainctl iam roles create my-role
+```
+
+After running this command, an interactive prompt will appear asking you to select what capabilities the new role should have and the organization under which the role should be created.
+
+You can avoid using the interactive prompt by including the `--parent` and `--capabilities` options in this command. 
+
+```sh
+chainctl iam roles create new-role --parent=example-org --capabilities=roles.create,roles.delete,roles.list
```
amouat, line 53: This doesn't work for me:

```
chainctl iam roles create new-role --parent=adrian-test --capabilities=roles.create,roles.delete,roles.list
Creating role "new-role" under location adrian-test.
Do you want to continue? [Y,n]: y
Error: [InvalidArgument] failed to create role: cannot add capability roles.create to custom roles
```
Note that I can use roles.list

```
identidock-cg/identidock on  v1 [!?] via 🐳 desktop-linux on ☁️  adrian@chainguard.dev took 3s
❯ chainctl iam roles create new-role --parent=adrian-test --capabilities=roles.list
Creating role "new-role" under location adrian-test.
Do you want to continue? [Y,n]: y
                             ID                             |   NAME   | DESCRIPTION | CAPABILITIES
------------------------------------------------------------+----------+-------------+---------------
  eb73f3f34112ccc96d3c4e64441b3a1001dfc26e/ca13a60e2f5678f0 | new-role |             | roles (list)
```

amouat (changes requested):
I didn't want to mess with existing identities so didn't do the final step of adding a role-binding. The syntax looks correct however.

I think just removing the `roles.create` and `roles.delete` from the `roles create` command is enough.

amouat (approved):

PR #1547, opened 2024-05-03
Reorganizations

ltagliaferri:
* Renames "Images Features" to "Working with Images" (plus redirects)
* Moves apko and melange under "Build Tools" and I also made some edits across docs (plus redirects)
* Swapped out the diff api featured article for a migration guide article on the home page


amouat (approved):
LGTM. Definite improvement.

PR #1585, opened 2024-05-24
Getting Started with the nginx Chainguard Image Tutorial

mcaveety:
## Type of change
Addition to documentation - adding new getting started tutorial

### What should this PR do?
Adds a new getting started tutorial to the Academy for the nginx Chainguard Image
https://github.com/chainguard-dev/internal/issues/3904

Expands Academy content that is helpful for new users, especially if a user may be interested in using an image for their nginx application(s)

### What are the acceptance criteria? // How should this PR be tested?
1. Please try to run this demonstration locally. I have had success with it multiple times on my MacOS laptop but I have not had an external test yet
2. Please read through the tutorial for any mistakes or for any suggestions for wording changes. I aimed to follow the style of the other "getting started" guides, but there has not been a formal review yet here

> Note: I did not create a demo with a multi-stage build. If we want to add something more complex, I am happy to try and figure that out. I wasn't sure how long we wanted this to be. It already seemed lengthy to me.

3. There is a pull request here https://github.com/chainguard-dev/edu-images-demos/pull/17 for the demo tutorial components which should be approved & merged before this article goes live since there is a reference to the repo link in the guide
4. Once the demo is tested, the other request merged, and any suggestions applied, it should be ready to go live

### DIFF
--- a/content/chainguard/chainguard-images/getting-started/nginx/index.md
+++ b/content/chainguard/chainguard-images/getting-started/nginx/index.md
@@ -1,0 +1,311 @@
+---
+title: "Getting Started with the NGINX Chainguard Image"
+type: "article"
+description: "Tutorial on how to get started with the NGINX Chainguard Image"
+date: 2023-01-09T11:07:52+02:00
+lastmod: 2023-01-19T11:07:52+02:00
+tags: ["Chainguard Images", "Products"]
+draft: false
+images: []
+menu:
+  docs:
+	parent: "getting-started"
+weight: 610
+toc: true
+---
+ <!-- This is outdated -->
+The nginx images maintained by Chainguard are a mix of development and production distroless images that are suitable for building and running nginx workloads.
+
+Because nginx applications typically require the installation of third-party dependencies via Composer, using a pure distroless image for building your application would not work. In cases like this, you'll need to implement a [multi-stage Docker build](https://docs.docker.com/build/building/multi-stage/) that uses one of the `-dev` images to set up the application.
+
+In this guide, we'll set up a distroless container image based on Wolfi as a runtime to execute a command-line nginx application.
+<!-- Update the above -->
+
+{{< details "What is distroless" >}}
+{{< blurb/distroless >}}
+{{< /details >}}
+
+{{< details "What is Wolfi" >}}
+{{< blurb/wolfi >}}
+{{< /details >}}
+
+{{< details "Chainguard Images" >}}
+{{< blurb/images >}}
+{{< /details >}}
+
+## Step 1: Setting up a Demo Application
+
+We'll start by creating a basic nginx application to serve as a demo. This app will instantiate a local webserver to display HTML in your browser. You will need to have nginx installed on your machine to follow along. If you do not already have nginx installed, you can do so using Homebrew:
+
+```shell
+brew install nginx
+```
+
+Alternatively, you can install nginx using another method [found here](https://nginx.org/en/docs/install.html).
+
+For this tutorial, you will be copying code to files you create locally. You can find this code throughout the demo or you can find a complete collection of files at the associated [GitHub repository here](https://github.com/chainguard-dev/edu-images-demos/tree/main/nginx).
+
+With nginx installed, create a directory for your app. In this guide we'll call the directory `nginxdemo`:
+
+```shell
+mkdir ~/nginxdemo/ && cd $_
+```
+
+Within this directory, we will create a `data` directory to contain our content hosted by the webserver.
+
+```shell
+mkdir data && cd $_
+```
+
+Using a text editor of your choice, create a new file `index.html` for the HTML content that will be served. In our case, we will use `nano`.
+
+```shell
+nano index.html
+```
+
+The following HTML file displays an image of Inky, our mascot, alongside a fun octopus fact.
+
+```html
+<!DOCTYPE html>
+<html>
+<head>
+<title>Chainguard NGINX Demo Application</title>
+<link rel="stylesheet" href="stylesheet.css">
+</head>
+
+<body>
+
+<h1>NGINX Demo Application</h1>
+
+<i><h2>from the <a href="https://edu.chainguard.dev/" target="_blank">Chainguard Academy</a></h2></i>
+
+<img src="inky.png" class="corners" width="250px">
+
+<i><h3>Did you know?</h3></i>
+<p>The Wolfi octopus is the world's smallest octopus, weighing in on average at less than a gram! They are found near the coastlines of the west Pacific Ocean.</p>
+
+</body>
+
+</html>
+```
+
+Copy this code to your `index.html` file, save, and close it.
+
+Next, create a CSS file named `stylesheet.css` using the text editor of your choice, and copy in the following code.
+
+```css
+/*
+Chainguard Academy NGINX demo application
+*/
+
+body {
+    text-align: center;
+    background-color: #fcebfc;
+    font-family: Arial, sans-serif;
+}
+
+h1 {
+    color: #df45e6;
+}
+
+h2, h3 {
+    color: #9745e6;
+}
+
+p {
+    color: #583ce8
+}
+
+.corners {
+    border-radius: 10px;
+}
+
+```
+
+After copying the code into the `stylesheet.css` file, save and close it.
+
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-images/getting-started/nginx/index.md
```diff
@@ -0,0 +1,262 @@
...
+    image/png                                        png;
+}
+```
+
+Save and close the `mime.types` file when you are finished editing it.
+
+Next, copy the filepath to the `nginx.conf` file you created earlier. Replacing the example path with this copied path, execute the following command to initialize the nginx server:
+
+```shell
+nginx -c /Users/username/nginxdemo/nginx.conf
```
smythp, line 198: This does run on Ubuntu 22, but I get a warning:

```
nginx: [emerg] open() "/usr/share/nginx/nginx.conf" failed (2: No such file or directory)
```

You might also want to warn that it has to be an absolute path, I tried relative first and it doesn't work. (I know you showed an absolute path, but guessing that a relative will also work might be somewhat common, and it's unexpected that it doesn't.)

content/chainguard/chainguard-images/getting-started/nginx/index.md
```diff
@@ -0,0 +1,262 @@
...
+    keepalive_timeout  65;
+
+    server {
+
+        listen 1313;
+        server_name localhost;
+        charset koi8-r;
+
+        location / {
+            root /Users/username/nginxdemo/data; # Update the file path for your system
```
smythp, line 164: There's kind of no way around these absolute paths in this tutorial, but you may want to put a note in somewhere with an example of how it would look on Linux, i.e. "/home/username/"

content/chainguard/chainguard-images/getting-started/nginx/index.md
```diff
@@ -0,0 +1,262 @@
...
+
+Save and close the `mime.types` file when you are finished editing it.
+
+Next, copy the filepath to the `nginx.conf` file you created earlier. Replacing the example path with this copied path, execute the following command to initialize the nginx server:
+
+```shell
+nginx -c /Users/username/nginxdemo/nginx.conf
+```
+
+To view the HTML content, navigate to `localhost:8080` in your web browser of choice. You should see a simple landing page with a picture of Inky and a fun fact about the Wolfi octopus.
```
smythp, line 201: Didn't you specify listen 1313 as a directive in the config? I think it shouldn't be appearing at 80. I'd triple check this (make sure your default config isn't loading, kill all nginx processes, run it again and see what port it appears on), as it should be appearing on 1313 and not 80 (I think).

content/chainguard/chainguard-images/getting-started/nginx/index.md
```diff
@@ -0,0 +1,262 @@
...
+
+In the `nginxdemo` directory, create the `Dockerfile` with the text editor of your choice. We will use `nano`:
+
+```shell
+nano Dockerfile
+```
+The following Dockerfile will:
+
+1. Start a new build based on the `cgr.dev/chainguard/nginx:latest` image;
+2. Expose port 8080 in the container for nginx to listen on;
```
smythp, line 217: This is pretty fiddly and annoying, but the Dockerfile isn't actually exposing the port here, the expose keyword just creates metadata. It's the -p flag that actually does the work. (Try removing the expose line in the Dockerfile and it should still have the same behavior as a test.) Basically just saying not to say that the Dockerfile will expose the port, you could change the line to say something about metadata, take it out, etc, up to you.

smythp (approved):
Great work on this. nginx is pretty fiddly, I got a warning with Ubuntu. With a tutorial like this I would usually recommend testing in Linux if you have access, but I understand you may not have a Linux environment for testing. It did run, just got that one warning.

I made a number of suggestions. In general, it's up to you to decide to address something or not. That is, you don't have to change something just because I said to. Feel free to ask questions in the comments or to push back on something, or just to ignore me on a point and move forward, that's all part of the process.

Cool test website, I appreciate the attention to detail there. Really clean copy, didn't find any copy edits which is rare. I'd honestly consider this good work from any colleague here, keep it up. :)

Not sure if you have merge access. If not, resolve the comments and then @ me and I'll merge.

content/chainguard/chainguard-images/getting-started/nginx/index.md
mcaveety: Did you get this warning only when you used a relative path? Or did you get it even with using an absolute path? Thanks for mentioning this.

content/chainguard/chainguard-images/getting-started/nginx/index.md
mcaveety: Good catch, thanks! I wanted it to be 8080, I think I changed to 1313 for testing at some point

mcaveety:
Waiting to merge to investigate the warning message & waiting for the edu-images-demo repo to get updated as well.

content/chainguard/chainguard-images/getting-started/nginx/index.md
smythp: This is actually worse than I thought. It doesn't seem possible to override the default accessing of log files in the /var/ directory, which by default is not accessible to a user in Ubuntu.  I've tried adding directives to write access_log or error_log to directories with permissions, but it still seems to want to access the default location first. There's supposed to be an -e flag to suggest error log location, but that option is removed, seemingly.If sudo is used, then the user gets a403 forbidden when visiting locaclhost:4000.

This is the command that got it to the point of working, though with a warning:

```
sudo chown -R patrick:patrick /var/log/nginx
```

The only real solution I've found here is having the user change permissions on the folder with the logs (/var/logs/nginx). In general, I'd say the nginx command line option is pretty useless, and nginx mostly needs to be configured globally and run with systemd. The default packaging for ubuntu is also pretty silly (logging to a folder with no permissions).

Not sure how you want to handle this. You may want to put in a note in about permissions. Typically, nginx is configured globally and run from systemd or similar, that's how I've used it, only using the command line utility to reload the global config. Happy to do a meeting to work through this, though I'll be away next week. 

@ltagliaferri  in case you have thoughts. I think Michelle would need access to a linux box to figure this part out. The issue here is in the non-containerized part of the tutorial.

mcaveety: When working on the demo, I intentionally removed the access_log and error_log directives in the nginx.conf file. I did this because I was running into issues with the file paths they wanted to default to, and I thought that it may be confusing for users to have to change these paths themselves, considering the default location depends on how you install nginx (I used Homebrew, for example). Taking these directives out seemed to resolve the problem and I thought that would be okay for the scope of this demo.

It sounds like you already tried to route the directives to a new location, which I think I tried (before I opted to take the directives out entirely) and I was not having success there. 

I agree that the command line option is quite useless. I would have used the default configuration file, but with the location of it varying with each installation method, I didn't think this made sense. 

:// I can give it a shot on Linux this weekend. I've been meaning to do an install on a spare drive so now is my chance!

mcaveety:
Great news! I have access to a Linux machine now. I'll be trying to troubleshoot this today.

content/chainguard/chainguard-images/getting-started/nginx/index.md
mcaveety: @smythp, I was able to test this out on Linux Mint and I reached a similar conclusion. Without changing the permissions on that default folder, I was not able to get things running. 

If I do update the permissions, and run nginx using `nginx -c /home/michelle/nginxdemo/nginx.conf`, then the local site loads fine.
However, I get this warning when I stop the server:
`nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:1`


For whatever reason, if I use `sudo` with `nginx -c /home/michelle/nginxdemo/nginx.conf` I get a 403 forbidden error when navigating to the local site. With `sudo nginx -s quit` there are no command-line warnings. 

My solution for now will be to mention this in the tutorial. I will add that nginx may need permissions to write to default log directories if it does not already have them, and exemplify this. 

I would choose to do a global configuration, however, I was worried that doing so would make the tutorial less flexible. From what I can tell, nginx is installed in different places depending on the OS and installation methods, so I wouldn't be able to cover each case. I installed using Homebrew on mac and apt on Linux and even then that's just two examples. Let me know what you think, I'll try to accomodate this permission problem in the tutorial as best I can, but if we think that things should be structured differently then we can make those changes

ltagliaferri: I think following your solution makes sense, I would not spend more time on this.

smythp:
Nice to see this merged, great tutorial! :tada:

PR #1586, opened 2024-05-24
Fix nav buttons

### LINKED ISSUE #860: Prev/Next buttons don't have logical ordering
**Describe the bug**

Previous/Next buttons don't seem to link to expected pages - they appear to jump around with no logical ordering (though the links are consistent).

**To Reproduce**

Next chain:
- https://edu.chainguard.dev/chainguard/chainguard-enforce/installation/preflight-checklist/
- https://edu.chainguard.dev/chainguard/chainguard-enforce/iam-groups/identity-examples/enforce-github-identity/
- https://edu.chainguard.dev/chainguard/chainguard-enforce/authentication/example-idps/okta/
- https://edu.chainguard.dev/chainguard/chainguard-enforce/authentication/log-in-chainguard-enforce/
- https://edu.chainguard.dev/chainguard/chainguard-enforce/policies/chainguard-policies-cli/
- ...

**Expected behavior**

Next button follows side bar navigation, e.g.

- https://edu.chainguard.dev/chainguard/chainguard-enforce/installation/preflight-checklist/
- https://edu.chainguard.dev/chainguard/chainguard-enforce/installation/alternative-installation-methods/

**Screenshots**

 - OS: macOS Ventura 13.4.1
 - Browser: Chrome 
 - Version: 114.0.5735.198

**Additional context**

n/a

smythp:
## Type of change

Bug fix

### What should this PR do?

This PR rewrites the site navigation Instead of going to random pages, it creates a depth-first run of the site (up to five levels) to collate all sections in order, extracts relevant pages from each section, creates a master listed page collection in order of weight, then moves the user along site using that flattened collection.

One detail here is that the implementation will go through all "regular" pages on a level, then go down or forward (in the tree) to the next  section. This does make a difference, as in a few places we have a section nested in the middle of another section, but I think it works fine as a previous/next implementation.

The previous and next buttons have a functionally random behavior on the current version of the site.

We should check that this solution doesn't break anything. In terms of behavior, it can only be an improvement to the previous and next buttons, but we also don't want anything else to get affected on the site.

Try out the behavior of the previous and next buttons. They should (mostly) line up with the left sidebar, and I used a similar approach to the sidebar code (depth-first to five levels). The overview page is the first page and should only have a next button. I think the video on sigstore is the last page. Unlisted pages should not appear.

### Notes

I just want to complain about how Hugo handles stuff. The previous and next logic is reversed (why?) and their previous and next method implementation is incredibly naive.

Resolves #860

### DIFF
--- a/layouts/partials/footer/main/docs-navigation.html
+++ b/layouts/partials/footer/main/docs-navigation.html
@@ -1,9 +1,17 @@
-{{ if or .Prev .Next -}}
+<h1>Prev</h1>
+<p>
+{{ .PrevInSection }}
+</p>
+  <h1>Next</h1>
+<p>
+{{ .NextInSection }}
+  </p>
+
+{{ if or .PrevInSection .NextInSection -}}
 <div class="docs-navigation">
-	<!-- https://www.feliciano.tech/blog/custom-sort-hugo-single-pages/ -->
-	{{ $pages := where site.RegularPages "Section" .Section -}}
-	{{ with $pages.Next . -}}
-		<a class="docs-navigation-button docs-navigation-button-previous" href="{{ .RelPermalink }}">
+
+  {{ with .NextInSection }}
+  <a class="docs-navigation-button docs-navigation-button-previous" href="{{ .RelPermalink }}">
 			<div class="chevron-rotator">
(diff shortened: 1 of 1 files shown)


ltagliaferri (approved):
Works great, thank you!

PR #1600, opened 2024-06-04
Updating platform docs workflow

erikaheidi:
This PR tries to fix the "Generate Platform Docs" workflow that has been broken for a couple months.

Changes:

- Simplified some parts and removed unecessary inputs
- Changed cron schedule to run once a day instead of every hour
- Changed changelog location to `autodocs/platform-changelog.md`
- Changed PR step to use the same action used by images-autodocs, which has been working without issues for a long time. This is also updated to use octo-sts
- Added a step to send a notification to Slack when a PR is open, just like images-autodocs

Testing this is tricky, it will need to be merged before we can test, since it depends on a new file (platform-changelog.md) that is in the PR.

I have followed instructions from [this doc](https://github.com/google-github-actions/auth?tab=readme-ov-file#direct-wif) to use identity federation without a service account.

### DIFF
--- a/.github/workflows/autodocs-platform.yaml
+++ b/.github/workflows/autodocs-platform.yaml
@@ -27,16 +27,14 @@ jobs:
         egress-policy: audit
 
     - name: 'Checkout default branch to $GITHUB_WORKSPACE dir'
-      uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3
+      uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
 
     - name: Authenticate to Google Cloud
       id: auth
-      uses: google-github-actions/auth@ceee102ec2387dd9e844e01b530ccd4ec87ce955 # v0
+      uses: google-github-actions/auth@71fee32a0bb7e97b4d33d548e7d957010649d8fa # v2.1.3
       with:
-        token_format: 'access_token'
         project_id: "${{ secrets.PROJECT_ID }}"
         workload_identity_provider: "${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}"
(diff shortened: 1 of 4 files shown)


erikaheidi:
I have followed instructions from the [google-github-auth](https://github.com/google-github-actions/auth?tab=readme-ov-file#usage) to use identity federation for authenticating with GCP instead of using federation through a service account.

.github/workflows/autodocs-platform.yaml
```diff
@@ -69,26 +69,23 @@ jobs:
...
       uses: chainguard-dev/actions/setup-gitsign@main
 
     - name: Authenticate to Google Cloud
       id: auth
-      uses: google-github-actions/auth@ceee102ec2387dd9e844e01b530ccd4ec87ce955 # v0
+      uses: google-github-actions/auth@71fee32a0bb7e97b4d33d548e7d957010649d8fa # v2.1.3
       with:
-        token_format: 'access_token'
         project_id: "${{ secrets.PROJECT_ID }}"
         workload_identity_provider: "${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}"
```
cpanato, line 82: you still need to have the svcc account, see this as example: https://github.com/chainguard-dev/infra-images/blob/main/.github/workflows/deploy-staging-images-bundle.yaml#L29-L33

.github/workflows/autodocs-platform.yaml
```diff
@@ -108,31 +105,33 @@ jobs:
       id: octo-sts
       with:
         scope: chainguard-dev/edu
-        identity: edu
+        identity: autodocs
```
cpanato, line 108: seems this sts policy https://github.com/chainguard-dev/edu/blob/main/.github/chainguard/autodocs.sts.yaml need to update to match this workflow

cpanato (approved):

PR #1619, opened 2024-06-13
Update how-to-install-chainctl.md

JoshuaCooper:
```install``` expects a path rather than a filename and path combination

## Type of change
Documentation

### What should this PR do?
Clarify installation process on linux systems. 
https://manpages.debian.org/bookworm/coreutils/install.1.en.html

Current instructions cause the following error. 
```install: target '/usr/local/bin/chainctl': No such file or directory```

Success with the command.

set $GID, set $UID run command with sudo.


amouat (approved):

amouat:
Thanks! LGTM.

PR #1625, opened 2024-06-19
Change example from nginx to caddy.

amouat:
## Type of change
Updated example code to use caddy instead of nginx.

(This was done as there is currently a shell in the nginx image).


amouat:
Pausing this as nginx change was rolled back

ltagliaferri:
Should we close this?

amouat:
I think so. I left it open for a while in case there was back and forth :)

PR #1639, opened 2024-06-27
removing quotas prereq from custom IDP docs

SharpRake:
## Type of change

Minor change removes a prereq from our custom IDP examples. Each org now has a quota of 1 by default, so this is unnecessary.

### What should this PR do?

resolves https://github.com/chainguard-dev/internal/issues/3931

https://chainguard-dev.slack.com/archives/C03H64P08UT/p1719438496333249

Minor change, I think just a quick glance will do.


smythp (approved):
Looks good :+1:

PR #1681, opened 2024-07-09
Remove sonatype reference in AR docs

imjasonh:


SharpRake (approved):
:grimacing: Good catch!

PR #1674, opened 2024-07-04
Add 301s for image reference pages

julienv3:
## Type of change
Sunsetting edu's image reference pages

### What should this PR do?
Add 301 redirects for image reference pages to images.chainguard.dev

### Why are we making this change?
Part of 24g priorities

### What are the acceptance criteria? 
Landing on the images list or individual image reference pages redirects to the appropriate location on images.chainguard.dev

### How should this PR be tested?
Netlify preview deployment doesn't seem to use the nginx config, unlike production?
Tested using curl locally against nginx.

PR #1690, opened 2024-07-15
Removing reference docs from edu

erikaheidi:
## Type of change
This PR removes the image reference docs from EDU, following PR #1674 that adds redirects to images directory.

### What should this PR do?
Clean up the EDU site after redirects are set to images directory

We're centralizing images reference docs in the images directory to avoid duplication of content

Site is functional without image reference docs. Be aware that redirects won't work in the Netlify preview. Redirects can be tested directly on the live site (this was confirmed in the previous PR)

Check the preview site to confirm image reference docs were removed


amouat (approved):
LGTM.
There's  too many changes to go through, but as best  I can tell it is just removing old docs.

PR #1697, opened 2024-07-17
Fix redirects

### LINKED ISSUE #1696: fix broken Assumable IDs redirects.
Evan [flagged some broken redirects](https://chainguard-dev.slack.com/archives/C03H64P08UT/p1721218004736579). This issue is to track the work to fix these.

SharpRake:
## Type of change

Evan flagged some broken redirects for us. This PR fixes those.

I also have had a sticky note on my desk to rearrange the Getting Started guides into alphabetical order, so I added those changes to this PR as well. 

### What should this PR do?

resolves #1696 

I updated the links directly so they should all work, no need to check the actual redirects. I did adjust the relevant doc's alias, though, which should catch anything I missed.


mcaveety (approved):
In the preview, I double-checked the alphabetical ordering, and that looks good. I also checked all of the links that were changed and they all go to the right from what I can see :) I think it's good to go!

PR #1699, opened 2024-07-19
Adding new WordPress guide

erikaheidi:
This PR adds a new "Getting Started Guide" for our WordPress images.

Preview: https://deploy-preview-1699--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/getting-started/wordpress/

Raw draft doc: https://docs.google.com/document/d/1PkiQaDh8cNJ6K9h2OY_JJXpR4d0hOZ0JtlnNKmm7VLM/edit?usp=sharing

### DIFF
--- a/content/chainguard/chainguard-images/getting-started/wordpress.md
+++ b/content/chainguard/chainguard-images/getting-started/wordpress.md
@@ -263,6 +263,11 @@ services:
       WORDPRESS_DB_USER: $WORDPRESS_DB_USER
       WORDPRESS_DB_PASSWORD: $WORDPRESS_DB_PASSWORD
       WORDPRESS_DB_NAME: $WORDPRESS_DB_NAME
+      WORDPRESS_CONFIG_EXTRA: |
+        # Disable plugin and theme update and installation
+        define( 'DISALLOW_FILE_MODS', true );
+        # Disable automatic updates
+        define( 'AUTOMATIC_UPDATER_DISABLED', true );
     volumes:
       - document-root:/var/www/html
 
@@ -297,14 +302,14 @@ You can now build and run your environment with:
 docker compose up --build
 ```
 
-The behavior of this WordPress setup should be the same as in the previous examples, but this time, the image is self-contained and doesn't rely on volumes set up within the host machine to work, in addition to not allowing new package installations or login through a shell.
+The behavior of this WordPress setup should be similar to the previous examples, but this time, the image is self-contained and doesn't rely on volumes set up within the host machine to work, in addition to not allowing new package installations or login through a shell. We also set up the `WORDPRESS_CONFIG_EXTRA` environment variable to disable the installation of new themes and plugins, and to block automatic updates. This increases security by blocking file changes in the container.
 
 To stop the services, type `CTRL+C` in the terminal where the services are running, and then run:
 
 ```bash
 docker compose down
 ```
-Please notice that although this setup allows for customizations through the WordPress dashboard, changes won't be persisted through container rebuilds.
+To keep your WordPress installation up-to-date with latest versions, you can use [digestabot](https://edu.chainguard.dev/chainguard/chainguard-images/videos/digestabot/), a GitHub Action that works in a similar way to Dependabot, sending a pull request to a repository whenever a new version of a container image is available. This will ensure you're always running the most recent version of WordPress available in Wolfi.
 
 ## Advanced Usage
 
--- a/content/chainguard/chainguard-images/getting-started/wordpress.md
+++ b/content/chainguard/chainguard-images/getting-started/wordpress.md
@@ -302,4 +302,4 @@ Please notice that although this setup allows for customizations through the Wor
 
 ## Advanced Usage
 
-{{< blurb/images-advanced image="Laravel" >}}
+{{< blurb/images-advanced image="WordPress" >}}
--- a/content/chainguard/chainguard-images/getting-started/wordpress.md
+++ b/content/chainguard/chainguard-images/getting-started/wordpress.md
@@ -99,7 +99,7 @@ WORDPRESS_DB_NAME=wordpress
 You can change these values to suit your needs. You can start the services by running:
 
 ```bash
-docker-compose up
+docker compose up
 ```
 
 If you navigate to `http://localhost:8000` in your browser, you should access the WordPress installation page. Follow the on-screen instructions to complete the WordPress setup. Keep in mind that any customizations will be lost once the environment is turned down.
@@ -107,10 +107,10 @@ If you navigate to `http://localhost:8000` in your browser, you should access th
 To stop the services, type `CTRL+C` in the terminal where the services are running, and then run:
 
 ```bash
-docker-compose down
+docker compose down
 ```
 
-This will remove the containers and networks created by the `docker-compose up` command. In the next example, we'll demonstrate how you can set up a volume with proper permissions to be able to persist customizations such as themes and plugins.
+This will remove the containers and networks created by the `docker compose up` command. In the next example, we'll demonstrate how you can set up a volume with proper permissions to be able to persist customizations such as themes and plugins.
 
 ## Example 2: Customizing a New WordPress Installation
 To persist customizations made to your WordPress site, such as the installation of new themes and plugins, you'll need to set up a volume with proper permissions in order to keep data between container rebuilds. This requires having a system user in the container with the same UID as your local system user on the host machine. To set this up, we'll create a custom Dockerfile that adds a `wordpress` user with the specified UID (set to `1000` by default, which is typically the UID of a regular user on Linux-based systems) to the `latest-dev` variant of the Chainguard WordPress Image. The Dockerfile also changes default permissions on the `/var/www/html` directory to allow the `wordpress` user to write to it.
@@ -179,13 +179,13 @@ Only the `app` service has changed in this example. We've added a `build` sectio
 To build your custom image and pass along your own UID as build argument, run:
 
 ```shell
-docker-compose build --build-arg UID=$(id -u) app
+docker compose build --build-arg UID=$(id -u) app
 ```
 
 You should get output indicating that the image was successfully built. Now you can get your environment up with:
 
 ```shell
-docker-compose up
+docker compose up
 ```
 
 Once the environment is up and running, you can access your WordPress installation from your browser at `localhost:8000`.
@@ -296,7 +296,7 @@ The behavior of this WordPress setup should be the same as in the previous examp
 To stop the services, type `CTRL+C` in the terminal where the services are running, and then run:
 
 ```bash
-docker-compose down
+docker compose down
 ```
 Please notice that although this setup allows for customizations through the WordPress dashboard, changes won't be persisted through container rebuilds.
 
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+
+volumes:
+  document-root:
+
+```
+
+You can now build and run your environment with:
+
+```shell
+docker compose up --build
```
mcaveety, line 291: It appears that both `docker compose` and `docker-compose` exist as commands... so far I've only noticed you using `docker-compose` in the commands. I'm not sure I understand the difference, but I'm flagging this since you might know, or so you can change it if it was unintentional.

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+
+To stop the services, type `CTRL+C` in the terminal where the services are running, and then run:
+
+```bash
+docker-compose down
+```
+
+This will remove the containers and networks created by the `docker-compose up` command. In the next example, we'll demonstrate how you can set up a volume with proper permissions to be able to persist customizations such as themes and plugins.
+
+## Example 2: Customizing a new WordPress Installation
```
mcaveety, line 115:
```suggestion
## Example 2: Customizing a New WordPress Installation
```
Minor, but I thought it looked better to have "New" capitalized here, since it's an adjective

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Example 3: Using the Distroless Variant of the WordPress Image
+This demo uses a multi-stage Docker build to create a final Distroless image to improve overall security. The Distroless image contains the necessary dependencies to run WordPress and won't allow for new package installations or shell access, reducing the image attack surface.
+
+The main difference here is that we're calling the entrypoint script at **build time** instead of run time. This is done to ensure the image is self-contained and doesn't rely on volumes set up within the host machine in order to work. Any customizations should be included in the `wp-content` folder that will be copied to the image at build time. Although this increases final image size due to the inclusion of custom content at build time, it limits what can be changed or added to the image once it's built.
+
+This demo includes a theme ([cue](https://wordpress.org/themes/cue/), a simple blogging theme) and a plugin ([imsanity](https://wordpress.org/plugins/imsanity/), a popular plugin used to resize images) to demonstrate how to include custom content in the image.
```
mcaveety, line 223:
```suggestion
This demo includes a theme ([Cue](https://wordpress.org/themes/cue/), a simple blogging theme) and a plugin ([Imsanity](https://wordpress.org/plugins/imsanity/), a popular plugin used to resize images) to demonstrate how to include custom content in the image.
```
It looks like these two plugins are stylized with capitals at the links you share. I would leave it uncapitalized if that's how it's usually written elsewhere, but I am not familiar with these projects, so you'd have better judgement on that

mcaveety (approved):
This is very clear and well written, @erikaheidi ! The commands all worked well for me and it was easy to follow along. Most of my comments are for stylization, so I'm approving this and you can make final edits if you feel like they are applicable. :)

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+```bash
+docker-compose down
+```
+
+This will remove the containers and networks created by the `docker-compose up` command. In the next example, we'll demonstrate how you can set up a volume with proper permissions to be able to persist customizations such as themes and plugins.
+
+## Example 2: Customizing a new WordPress Installation
+To persist customizations made to your WordPress site, such as the installation of new themes and plugins, you'll need to set up a volume with proper permissions in order to keep data between container rebuilds. This requires having a system user in the container with the same UID as your local system user on the host machine. To set this up, we'll create a custom Dockerfile that adds a `wordpress` user with the specified UID (set to `1000` by default, which is typically the UID of a regular user on Linux-based systems) to the `latest-dev` variant of the Chainguard WordPress Image. The Dockerfile also changes default permissions on the `/var/www/html` directory to allow the `wordpress` user to write to it.
+
+Navigate to the `02-customizing` directory to follow along. This is how the described Dockerfile included in this directory looks like:
```
max-allan-cgr, line 118:
```suggestion
Navigate to the `02-customizing` directory to follow along. This is how the described Dockerfile included in this directory looks:
```

(Or "This is what the Dockerfile looks like" - but not "how" and "looks like".)

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+The behavior of this WordPress setup should be the same as in the previous examples, but this time, the image is self-contained and doesn't rely on volumes set up within the host machine to work, in addition to not allowing new package installations or login through a shell.
+
+To stop the services, type `CTRL+C` in the terminal where the services are running, and then run:
+
+```bash
+docker-compose down
+```
+Please notice that although this setup allows for customizations through the WordPress dashboard, changes won't be persisted through container rebuilds.
```
max-allan-cgr, line 301: Rather than leave this "hole" would it be better to add:
```
services:
  app:
    environment:
      WORDPRESS_CONFIG_EXTRA |
        define('DISALLOW_FILE_MODS',true);
```
In the compose file? Then it disables most changes that are not persisted in the database.

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+This demo uses a multi-stage Docker build to create a final Distroless image to improve overall security. The Distroless image contains the necessary dependencies to run WordPress and won't allow for new package installations or shell access, reducing the image attack surface.
+
+The main difference here is that we're calling the entrypoint script at **build time** instead of run time. This is done to ensure the image is self-contained and doesn't rely on volumes set up within the host machine in order to work. Any customizations should be included in the `wp-content` folder that will be copied to the image at build time. Although this increases final image size due to the inclusion of custom content at build time, it limits what can be changed or added to the image once it's built.
+
+This demo includes a theme ([cue](https://wordpress.org/themes/cue/), a simple blogging theme) and a plugin ([imsanity](https://wordpress.org/plugins/imsanity/), a popular plugin used to resize images) to demonstrate how to include custom content in the image.
+
+Navigate to the `03-distroless` directory to follow along. This is how the Dockerfile included in this directory looks like:
```
max-allan-cgr, line 225:
```suggestion
Navigate to the `03-distroless` directory to follow along. This is what the Dockerfile included in this directory looks like:
```

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+description: "Tutorial on how to get started with the WordPress Chainguard Image"
+date: 2024-07-19T11:07:52+02:00
+lastmod: 2024-07-19T11:07:52+02:00
+tags: ["Chainguard Images", "Products"]
+draft: false
+images: []
+menu:
+  docs:
+    parent: "getting-started"
+weight: 610
```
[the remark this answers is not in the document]
erikaheidi: Nice! I wasn't aware. Thank you!

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+      MARIADB_PASSWORD: $WORDPRESS_DB_PASSWORD
+      MARIADB_DATABASE: $WORDPRESS_DB_NAME
+    ports:
+      - 3306:3306
+
+volumes:
+  document-root:
+```
+
+Only the `app` service has changed in this example. We've added a `build` section that references the custom Dockerfile, and we've set the `UID` build argument to `1000` by default. This can be overwritten at runtime when you call `docker compose up`. We've also added a volume share to persist the contents of the `wp-content` folder to the host machine.
```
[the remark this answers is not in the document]
erikaheidi: in fact, Docker Compose is now built-in with Docker, so at least here on my machine I can simply call `docker compose`, and `docker-compose` gives me _command not found_. does that not work for you?

content/chainguard/chainguard-images/getting-started/wordpress.md
erikaheidi: The file is `docker-compose.yaml`, but the command is `docker compose`

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+This is only possible because of the custom Dockerfile we created, which added a `wordpress` user with the same UID as the local user on the host machine.
+
+You can now install new themes and plugins, and they will persist between container rebuilds.
+
+To stop the services, type `CTRL+C` in the terminal where the services are running, and then run:
+
+```bash
+docker-compose down
```
erikaheidi, line 213:
```suggestion
docker compose down
```

erikaheidi: This was probably a wrong autocomplete from my editor! supposed to be `docker compose`

content/chainguard/chainguard-images/getting-started/wordpress.md
erikaheidi: Oh I see now there was a wrong `docker-compose` in the guide as command, that was probably a wrong autocomplete from my editor!

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+In this guide, we'll demonstrate 3 different ways in which you can use the WordPress Chainguard Image to build and run WordPress projects.
+
+## Preparation
+This tutorial requires Docker to be installed on your local machine. If you don't have Docker installed, you can download and install it from the [official Docker website](https://docs.docker.com/get-docker/).
+
+### Cloning the Demos Repository
+Start by cloning the demos repository to your local machine:
+
+```shell
+git clone git@github.com:chainguard-dev/edu-images-demos.git
```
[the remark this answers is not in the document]
erikaheidi: That's a good point! I don't think we have a convention, but as a general best practice I would think using SSH would be the way to go. We can bring this up at the weekly, for now I think I will keep SSH just in case.

content/chainguard/chainguard-images/getting-started/wordpress.md
```diff
@@ -0,0 +1,305 @@
...
+  document-root:
+```
+
+In this Docker Compose example, we define 3 services: `app`, `nginx`, and `mariadb`. Here's a breakdown of each service:
+
+- The `app` service uses the `latest-dev` variant of the Chainguard WordPress Image, and is configured to connect to the `mariadb` service. The entrypoint script in the WordPress image looks for environment variables to set up a custom `wp-config.php` file. The volume `document-root` defines a volume that will be shared between the `app` and the `nginx` services.
+- The `nginx` service uses the [Chainguard Nginx Image](https://images.chainguard.dev/directory/image/nginx/overview), and is configured to serve the WordPress application on port 8000.
+- The `mariadb` service uses the [Chainguard MariaDB Image](https://images.chainguard.dev/directory/image/mariadb/overview), and is configured with the necessary environment variables to create a database for the WordPress application.
+
+The environment variables used in this example are defined in a `.env` file located in the same directory as the `docker-compose.yaml` file:
```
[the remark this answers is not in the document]
erikaheidi: Good point! I will add a note

content/chainguard/chainguard-images/getting-started/wordpress.md
erikaheidi: These are great suggestions! I think it makes total sense for the distroless setup. I will update the guide to include these and add notes about updates

PR #1710, opened 2024-07-25
Increase navbar display depth

smythp:
## Problem

The left navbar only displays down four levels in our content. as reported in https://github.com/chainguard-dev/internal/issues/4079.

## Description

This change allows content to be nested up to five levels. It also  updates the left navbar names for the Trivy and Grype tutorials.

## Notes

Resolves https://github.com/chainguard-dev/internal/issues/4079

cc @SharpRake


SharpRake (approved):
:+1:  Nice! I think this looks great. I was worried about it getting cluttered but the layout looks fine to me

Issue #1717, opened 2024-07-29
Remove depricated UI section


smythp:
Remove the depricated UI section and fix the left sidebar logic so that all content displays correctly.

smythp:
Closed in https://github.com/chainguard-dev/edu/pull/1718

PR #1718, opened 2024-07-29
Remove UI section

smythp:
## Description

We've been keeping the deprecated UI section around because removing it caused problems with the left navbar. This PR removes the UI section and resolves the problems with the left navbar.

## Notes

This PR  also removes a now-unnecessary workaround from the previous/next navigation template.

Resolvves #1717


smythp:
cc @SharpRake

SharpRake (approved):
looks great!

PR #1724, opened 2024-07-30
minor updates found while reading

matthewhelmke:
## Type of change

Just a few minor things I found while reading to familiarize myself with our docs. There was an extra space in the middle of one sentence and then a page where the spacing around headings was inconsistent. I'm not sure what we prefer (blank line before and after??) but I did make the page internally consistent. 

### What should this PR do?

Fixes some small things.

n/a

n/a

n/a


SharpRake (approved):
These changes lgtm, though I don't think line breaks around headings matter necessarily

PR #1728, opened 2024-07-31
Academy glowup

smythp:
## Description

Implement the new brand glowup colors, logos, iconography, fonts, and footer. Also refreshed margins, left nav, social buttons, etc.

The PR also implements a number of other Academy improvements tracked under https://github.com/chainguard-dev/internal/issues/4072

## Changes

- [x] Change all typography across Academy
- [x] Update logo usage across Academy, y
- [x] Update color palette across Academy
- [x] Update global footer to match new .dev website footer

## Notes

Resolves https://github.com/chainguard-dev/internal/issues/3914


mcaveety:
you are doing amazing work!!

sheesh:
Amazing work, @smythp! Just reviewed the site and found a few tiny possible issues:
* Logo font in top left corner looks slightly different than logo font in title page. Not sure if that's intentional. 
image
image
* Image is broken in left nav: 
image
* I think the caption text across the landing page should not be bold:  
image
* There's something unhappy going on at the bottom of the page, not sure if it's a ghost of Jamon's terminal: 
image
* And maybe you're still working on it, but I think Kaylin wanted the swirl design incorporated in the cards. 

Thanks so much, you've accomplished a lot!

sheesh (approved):
I'm approving so that I don't hold up the launch, but there are a few issues you should address before merging. I wrote them in another comment. Thanks!

ltagliaferri:
The body text on the tutorial pages is hard to read for me. It appears bold?

smythp:
Good catch on the top right "Chainguard Academy" font, that should have been our title font, should be fixed now.

The footer seems to be the same font as all regular body text. If the regular font is too heavy, that's another discussion, but it's Lato, weight 400, normal style. Personally I wouldn't want to mess with it and mess up our contrast ratio, but I'm also not the best person to talk to about fonts and the effect the have. Let me know if you think I'm missing something here, I'm using the browser calculations and also brought someone in to give it an eye. 

Edit: OK, looked at this more. I think the issue is that in the previous site, we used a lighter gray on the caption text. I'll experiment with making the caption text regular purple. Might be  Monday thing.

The broken image in the left nav is our scarf pixel. It looks like it quickly hides itself on the current deployed site, I predict it will hide itself once the site is fully deployed but I can look into it if the issue persists after going live.

Similarly, the bottom broken thing is netlify trying to inject something related to this being a preview. I have no idea why it's so broken, but it's not actually part of our site's code and gets stuck in by netlify in the preview, so hoping it goes away...

Re: swirl, they haven't shared the assets yet and she said it's fine for launch, I'll add it when I get the assets.

Thanks, Erin :)

smythp:
tl;dr Font enboldening fixed, title on top left font fixed, broken looking stuff a deploy preview artifact (I hope, but pretty sure), and waiting on swirly assets before I can make that change, but they said not to wait on it.

PR #1732, opened 2024-08-01
First pass at rewriting the tokens/API section.

max-allan-cgr:
Signed-off-by: Max Allan <max.allan@chainguard.dev>

## Type of change
Documentation change

### What should this PR do?
Fixes some incorrect content on how to get a token.

The suggested method to get a token for private images was incorrect.

Review by docs team.

Try out the commands proposed.

### DIFF
--- a/content/chainguard/chainguard-images/working-with-images/using-the-tag-history-api.md
+++ b/content/chainguard/chainguard-images/working-with-images/using-the-tag-history-api.md
@@ -41,26 +41,30 @@ Where `IMAGE_NAME` is the name of the image that you want to pull the tag histor
 
 For public images (tagged as `latest` or `latest-dev`), you can request a registry token anonymously, without providing any pre-existing auth.
 
-The following command will obtain a token for the **Python** image and register a variable called `tok` with the resulting value, which you can use in a subsequent command to obtain the tag history:
+The following command will obtain a token for the **Python** image and register a variable called `auth_header` with the resulting value, which you can use in a subsequent command to obtain the tag history:
 
 ```shell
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-images/working-with-images/using-the-tag-history-api.md
```diff
@@ -41,46 +43,58 @@ Where `IMAGE_NAME` is the name of the image that you want to pull the tag histor
...
 ```
 
-To make sure your token is set, you can run the following command:
+You may use the `crane` tool to get your token instead:
 
 ```shell
-echo $tok
+auth_header="$(crane auth token -H cgr.dev/ORGANIZATION_NAME/IMAGE_NAME)"
 ```
+Where `ORGANIZATION_NAME` is the name of your organization, for example: `company.com`, `IMAGE_NAME` is the name of the image, for example: `chainguard-base` (Note: the image name specified here does not need to be the same image you pass to the API later! But it needs to be a valid image from your registry.)
```
amouat, line 66: A little confusing.

Maybe:

> Replace `ORGANIZATION_NAME` and`IMAGE_NAME` as required. For example, if your organization is `foo.com` and you're interested in the `chainguard-base` image:
> 
> ```
> auth_header="$(crane auth token -H cgr.dev/foo.com/chainguard-base)"
> ```

 

I'm not sure if it's worth mentioning matching image names, just in case it changes.

amouat (approved):
LGTM, minor suggestions.

amouat:
There's still a section at the bottom on "enterprise images". I'm not sure if that also needs updated or if we can just delete. Let's not let it get in the way of this fix though.

PR #1745, opened 2024-08-05
Make monospace links in dark mode distinguishable as links

### LINKED ISSUE #1707: Bug: Monospaced links in dark mode are visually indistinguishable
**Describe the bug**
On the Academy, we have some monospaced text which also serves as a link. In light mode, both the monospacing and the light-blue color telltale of a link is visible as well. However, when you switch to dark mode, the coloring disappears and you cannot determine it is a link without hovering over the monospaced text. I don't suspect this is an issue in most cases, but a recent article I published has this issue.

**To Reproduce**
1. Go to https://edu.chainguard.dev/chainguard/chainguard-images/working-with-images/compiled-programs/
2. Navigate to the `wolfi-base` image section
3. Observe that, in light mode, the text appears as both monospaced and a link.
4. Switch the site to dark mode
5. Observe that, in dark mode, the text only appears as monospaced, though the link still functions

**Expected behavior**
I would expect that in either dark or light mode, the text still appears as blue. I understand that in dark mode the text becomes white, which is likely why this occurs. Though, it makes monospaced links visually impossible to detect, and you'd need to hover over to figure out it actually links anywhere. 

**Screenshots**
Light mode: 
Screenshot 2024-07-23 at 9 34 11 AM
Dark mode: 
Screenshot 2024-07-23 at 9 35 08 AM

 - OS: macOS Sonoma 14.5
 - Browser Google Chrome
 - Version 126.0.6478.183 (Official Build) (arm64)

**Additional context**
I actually ran into this issue myself without realizing it -- I was confused, when in testing, sometimes the links would appear visibly and other times it wouldn't. If I had to guess, I imagine switching to dark mode would cause all text to turn white and it overrides the link color. 🤷

smythp:
## Problem

In dark mode, monospace + links just look like monospace.

## Description

This adds link color to monospace, keeping the dark background and monospace font.

## Notes

Resolves #1707

cc @mcaveety

### DIFF
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -1084,11 +1084,11 @@
     }
 
     [data-dark-mode] body a code:not(.hljs) {
-        color: var(--primary-2)
+        color: var(--article-link-color)
     }
 
     [data-dark-mode] body code:not(.hljs) a {
-        color: var(--primary-2)
+        color: var(--article-link-color)
     }
 
     .pagination {
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -1083,6 +1083,10 @@
         letter-spacing: 0.0025em;
     }
 
+    [data-dark-mode] body code:not(.hljs) {
+        color: var(--primary-1)
+    }
+
     .pagination {
         justify-content: center;
         margin-top: 48px;
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -1083,7 +1083,11 @@
         letter-spacing: 0.0025em;
     }
 
-    [data-dark-mode] body code:not(.hljs) {
+    [data-dark-mode] body a code:not(.hljs) {
+        color: var(--primary-1)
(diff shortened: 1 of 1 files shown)


mcaveety (commented):
It looks like _all_ monospaced text in dark mode is now link-blue. See screenshots:

Dark mode
Light mode

smythp:
Good catch. Can you check now?

mcaveety (commented):
This looks better - however, can I ask that you change the link color in the monospace to match the teal of the other links? I didn't notice that before, sorry 😰 you can see the different colors in this screenshot to the left and right
Screenshot 2024-08-05 at 4 06 27 PM

mcaveety (approved):
I'm going to go ahead and approve it so you're able to merge, but I do think the color change in the previous comment ^^ should be implemented. :)

smythp:
Can you check the new version? I'll wait to merge. Thanks for QAing this :)

mcaveety:
yay looks great! I will go ahead and merge it since I think we're on the same page!

PR #1746, opened 2024-08-05
Updating console URL in docs

SharpRake:
## Type of change

We recently started redirecting requests to `console.enforce.dev` to `console.chainguard.dev`. This PR updates the any cases where `console.enforce.dev` comes up to point to the new URL.

### What should this PR do?

resolves https://github.com/chainguard-dev/internal/issues/4073

Part of keeping our docs current and accurate

Links should work!

Feel free to test the links yourself but a visual scan ought to do.

Note that two of the custom IDP docs (Azure AD + PingID) include screenshots that point to the old URL. Additionally, Azure AD has been rebranded, so that whole doc will likely need to be udpated.

I tried logging into both of these platforms to see about grabbing new screenshots but unfortunately I'm locked out of both accounts (I think the free trials I signed up for have expired and now I'm persona non grata). I will keep exploring my options there but for now I'm saying those updates are beyond the scope of this.


content/chainguard/administration/manage-chainctl-config.md
```diff
@@ -47,14 +47,13 @@ Flags:
...
 Global Flags:
-      --api string                   The url of the Chainguard platform API. (default "https://console-api.enforce.dev")
-      --audience string              The Chainguard token audience to request. (default "https://console-api.enforce.dev")
-      --config string                A specific chainctl config file.
-      --console string               The url of the Chainguard platform Console. (default "https://console.enforce.dev")
-      --issuer string                The url of the Chainguard STS endpoint. (default "https://issuer.enforce.dev")
-  -o, --output string                Output format. One of: ["", "json", "id", "table", "terse", "tree", "wide"]
-      --timestamp-authority string   The url of the Chainguard Timestamp Authority endpoint. (default "https://tsa.enforce.dev")
-  -v, --v int                        Set the log verbosity level.
+      --api string        The url of the Chainguard platform API. (default "https://console-api.enforce.dev")
```
amouat, line 50: Did you rerun the command? Just checking it's updated there as well.

amouat (approved):
LGTM. Just want to verify the chainctl command has been updated as well (otherwise I'd leave that change until it is).

content/chainguard/administration/manage-chainctl-config.md
SharpRake: Haha, I actually didn't at first then realized I probably ought to.This is the command's output as of ~yesterday.

`console-api.enforce.dev` and `issuer.enforce.dev` aren't going to be updated, at least not any time soon

PR #1783, opened 2024-08-23
Modifying STIGs overview to include checklist download

SharpRake:
## Type of change

This PR adds one command showing how to download the STIG datastream file, which is like a checklist of 

### What should this PR do?

resolves https://github.com/chainguard-dev/customer-issues/issues/1632

Clarifying how people can use the STIG.

Should make sense as written.

There is a new command, but it's a simple `curl` command. Feel free to test it but I'm mostly concerned about clarity + content etc.


erikaheidi (approved):
looks good to me, and the curl command works fine. 👍🏼

PR #1816, opened 2024-09-19
Adding CIS Benchmarks overview

SharpRake:
## Type of change

This PR adds a high-level overview of CIS Benchmarks to the Compliance section.

### What should this PR do?

Resolves https://github.com/chainguard-dev/internal/issues/4179

Expanding our compliance section, and hopefully educating users on what CIS benchmarks are and why they might be valuable.

No tech test, just a close read please!


erikaheidi (approved):
Looking good to me!

PR #1817, opened 2024-09-20
fix inky to linky in wolfi intro

kacole2:
## Type of change
Added documentation. This is a fix

### What should this PR do?
All the references to linky were spelled as inky so the demo wasn't working correctly when pulling the images and files in the wolfi intro

Simple edit to the docs

n/a

check my spelling


cpanato (approved):
lgtm

but i will defer to @erikaheidi for final approval

erikaheidi (approved):
Ok, confirmed that the demos repository is already updated with Linky in place of Inky, so this should be fine.

Issue #1797, opened 2024-09-03
Installing the chainctl binary permissions set fails


maxgio92:
**Describe the bug**

Following the documentation to install the `chainctl` tool with `curl` using `bash` shell, the permissions set fails because the `GID` variable is leveraged, but Bash does not set it by default, unlike [other ones](https://www.gnu.org/software/bash/manual/bash.html#Bash-Variables) like `UID`.

**To Reproduce**

Download the binary:

```shell
curl -o chainctl "https://dl.enforce.dev/chainctl/latest/chainctl_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/aarch64/arm64/')"
```

Moving it and setting its permissions:

```shell
sudo install -o $UID -g $GID -m 0755 chainctl /usr/local/bin/
[sudo] password for firstname.lastname:
install: invalid group '-m'
```

> As for the [guide for installing with curl](https://edu.chainguard.dev/chainguard/administration/how-to-install-chainctl/#install-with-curl).

**Expected behavior**
To correctly set permissions bits on the file.

**Screenshots**
NA

**Desktop (please complete the following information):**
 - OS: Fedora Silverblue
 - Browser: chrome
 - Version: 40

**Smartphone (please complete the following information):**
NA

**Additional context**

The `GID` group id is not always set. This is the case of Bash which doesn't set it by default. It does it instead for the `UID` environment variable. Bash reference about this [here](https://www.gnu.org/software/bash/manual/bash.html#Bash-Variables).

A possible alternative I think it could be using the `id` (`id -u`/`id -g`) shell builtin which may be implemented by most shells.

ltagliaferri:
This was fixed in #1822

PR #1822, opened 2024-09-24
GID is not set in bash so use id -g

murraybd:
The instructions for installing `chainctl` fail in bash due to GID not being a variable set by the shell. 

This resolves that and https://github.com/chainguard-dev/edu/issues/1797 .


SharpRake (approved):
I can confirm the original command doesn't work but this change fixes it. Thanks @murraybd!

PR #1828, opened 2024-09-26
rumble: update bq queries to use dedicated views for rumble

cmdpdx:
Update the bigquery queries to use views specifically created for rumble. These views already have the image and date filtering, so they have been removed from these queries for readability.


k4leung4 (approved):

SharpRake (approved):
Thanks @cmdpdx! The data is loading in the preview (which is expected) but I'll let you know if I notice any issues after the changes go live.

PR #1831, opened 2024-10-01
rumble: fix legacy csv query and csv output to match what the chart js expects

cmdpdx:
Drop unused fields `scanner_version` and `scanner_db_version`, append the tag to the end of the `image` name as expected by the javascript generating the chart, and exclude images with `latest-dev` tags since those are never used in image comparisons (cuts `data.csv` down from ~8MB to 4MB 📉 )


jamonation:
Looks good, we saw charts rendering locally with these changes, so +1 from me!

SharpRake (approved):
Thanks @cmdpdx!

PR #1850, opened 2024-10-15
Remove Wolfi community reference

jspeed-meyers:
Remove the outdated reference


SharpRake (approved):
Thanks for catching this!

PR #1854, opened 2024-10-18
Sunsetting terminal

erikaheidi:
This PR removes the terminal from content that still had it in their front matter, and also removes the partial that renders the terminal on pages.

ref: https://github.com/chainguard-dev/internal/issues/4143

cc @jamonation


matthewhelmke (approved):
LGTM
I also appreciate seeing trailing spaces removed and typos fixed!

erikaheidi:
Alright, it is passing now so I will merge. Thanks @matthewhelmke for the quick review!

PR #1861, opened 2024-10-21
Remove card view from console public images doc

angela-zhang:
## Type of change
small update to docs to remove the card view in "Public images" in the console


SharpRake (approved):
Thanks @angela-zhang!

PR #1863, opened 2024-10-23
massage FIPS compliance messaging

matthewhelmke:
This PR will complete https://github.com/chainguard-dev/internal/issues/4268 once we are done with reviews.

May I request that @xnox give the additions here a look and please let me know whether more needs to be said? I took my info from the Istio issue and discussion in https://github.com/chainguard-dev/customer-issues/issues/1717 where you seemed to have the best handle on what would be most useful. I would love to prevent future customer support calls on this.


xnox (changes requested):
there are typos, and also that's not quite right.

As in, proposed requirements wordings is too much  / more than actually necessary.

matthewhelmke:
I'd love a suggestion or hint at how to correct it.

matthewhelmke:
@SharpRake and @sheesh , thoughts?

SharpRake (approved):
I'm not 100% sure of the correctness of the new info (sounds right to me, but I'm not an expert). Style/content wise, though, I give it a :+1:

matthewhelmke:
I've made changes, @xnox, based on your review as well as [this newer issue](https://github.com/chainguard-dev/customer-issues/issues/1797). Is this suitable? I'm inferring a little here and don't want to make a doc update based on my intelligent guesses.

xnox:
This describes more things of what not to do. But doesn't address the actual customer needs from documentation as to "but what should I do?" and to most cases the answers "nothing, as by default everything is configured to just work - and if it doesn't it means it is api/abi incompatible or non-compliant and a bunch of engineering work needs to be done to make it work". Which this documentation update does not address.

PR #1871, opened 2024-10-29
Removing outdated recommendation

erikaheidi:
This PR removes an outdated note from our "Custom IDPs" guide. The context is that a customer tried to follow the guide and had issues with that specific recommendation as reported [here](https://chainguard-dev.slack.com/archives/C03H64P08UT/p1729742200757819) and on [this GitHub issue](https://github.com/chainguard-dev/internal/issues/4361).

Since the Issuer URL is usually an information made available by the service provider, I believe we can safely remove the note here and let users consult this info directly from their OAuth app / settings page on their provider of choice.


smythp (approved):
LGTM, it's a weirdly strongly worded paragraph in any case.

PR #1873, opened 2024-10-30
fixes a race while traversing the links concurrently

Dentrax:
## Type of change
Fix

### What should this PR do?
Fixes `fatal error: concurrent map iteration and map write`

`-`

`-`

`-`


developer-guy:
here is the link where you can be able to see the issue: https://github.com/chainguard-images/images-private/actions/runs/11590038133/job/32266778061?pr=5892

jamonation (approved):
Works well here, definitely fixes the occasional race that I was seeing. Thanks for tracking this down!

PR #1883, opened 2024-11-06
Remove reproducible reference as it's not accurate.

amouat:
## Type of change
Fix to text, it's not accurate to say melange is reproducible.

From Jason Hall:

>We don't record the versions of tools used to build a package, for example. So if there's a go release since that package was built, you can't guarantee the same go version, and there goes your reproducibility


erikaheidi (approved):
lgtm, but I think we may find this in other articles. the wolfi overview mentions `Fully declarative and reproducible build system` under "Wolfi Features" , for example.

PR #1888, opened 2024-11-08
Maintenance tools

erikaheidi:
This PR includes a custom tool under `autodocs/` to help with maintenance and content audits. This is discussed in more detail in internal issue [4360](https://github.com/chainguard-dev/internal/issues/4360).

This is a CLI tool written in PHP. It can be expanded with more utilities as we need it.

Having the tool alongside the content is a lot easier to manage, so instead of creating a separate repository as I did in the past with Autodocs, I reused the existing audodocs folder to set up this tool.


autodocs/README.md
```diff
@@ -0,0 +1,57 @@
...
+### Get articles with invalid timestamps
+
+```shell
+./autodocs audit --invalid
+```
+
+### Get top articles from analytics csv
+
+```shell
+./autodocs audit topcontent data=input.csv
```
SharpRake, line 33: Not a blocking comment, just a question - will we have an "easy" way of getting this data?

SharpRake (approved):
This is a really cool tool! Thanks so much for your work on this Erika.

I do want to note that I had to install the `php8.3-xml` package to get it to work, but I doubt that's worth flagging or anything.

autodocs/README.md
erikaheidi: yes! I am working on a spreadsheet with the results from this, for our first sprint it will be more manual but the plan is to automate as much as possible later on.

erikaheidi:
> This is a really cool tool! Thanks so much for your work on this Erika.
> 
> I do want to note that I had to install the `php8.3-xml` package to get it to work, but I doubt that's worth flagging or anything.

Thank you for testing it! I am planning to create a Dockerfile to make it easier to run it without a local php dev.

PR #1923, opened 2024-11-21
Modify rumble JSON to remove two images

ltagliaferri:
https://github.com/chainguard-dev/edu/issues/1852

@jdolitsky @jamonation is removing the images from the JSON what we needed to do for that issue?


SharpRake (approved):
approving! But let's also see what Jamon/Josh have to say.

PR #1941, opened 2024-12-04
Added feature flag functionality

smythp:
## Description

Adds feature flag / feature preview functionality when assigning `preview: true` in front matter.

## Details

See https://github.com/chainguard-dev/internal/issues/4463

When `preview: true` is set in the front matter of any article page, the following should now occur:

- The page is delisted in the left nav
- The page is taken off the previous and next button rotation
- A banner is displayed with a message about how this documentation is in preview
- A robots.txt entry is created for this page's permalink URL so that the page is temporarily not indexed. (go to `localhost:1313/robots.txt`  or `previewdomain/robots.txt` to see this.

## Testing

To test, add `preview: true` as in the below snippet to any article page.

```
draft: false
preview: true
images: []
```

Note that the location doesn't matter, just that the front matter value is set. Draft should remain false, as draft pages are not built by Hugo (though, confusingly, they still get served in the dev server).

Once the preview is set, check that the page is delisted on left nav, delisted on prev/next buttons at bottom of articles, that the banner appears correctly, and that the robots.txt has an appropriate entry.  Try to catch any edge cases :)

Might be obvious, but once the preview flag is set, you'll need to navigate directly to the page, so you might want to save the URL first before setting the preview flag.

## Notes

We didn't previously have a `robots.txt`, so this PR creates one.

Resolves https://github.com/chainguard-dev/internal/issues/4463


layouts/partials/notice.html
```diff
@@ -1,11 +1,22 @@
...
+
+
+{{if $page.Params.preview }}
+{{ $previewnotice := dict "note" $previewmessage "preview" "true" "type" "cta"}}
+{{ $notices = $notices | append $previewnotice }}
+{{ end }}
+
+{{ range $i := $notices }}
+{{ if or (eq $i.title $.Params.Title) (eq $i.preview "true") }}
+<section style="margin-bottom:-1em" class="section section-cta">
```
ltagliaferri, line 15: This is a nit and out of scope of this PR, but I wonder if the CTAs can be a different color on dark mode. They don't stand out so much to me since it's the same color as the tags.

dark mode cta

ltagliaferri (approved):

ltagliaferri:
Tested and works well, thank you!

layouts/partials/notice.html
smythp: OK, we discussed this and sounds like you might pick out a color combo and do a quick PR on this :+1:

PR #1943, opened 2024-12-04
Adding canonicalUrl frontmatter option

### LINKED ISSUE #1942: Add canonical URL functionality to Hugo front matter
I'm reworking a pair of Chainguard blog posts to publish them on Academy. There was some concern about SEO cannibalization and it was suggested that we use canonical URLs. I found [this tutorial](https://blog.concannon.tech/tech-talk/hugo-canonical-url/) on how to add that functionality to Hugo and will implement it.

SharpRake:
## Type of change

I'm reworking a pair of Chainguard blog posts to publish them on Academy. There was some concern about SEO cannibalization and it was suggested that we use canonical URLs.

I found [this tutorial](https://blog.concannon.tech/tech-talk/hugo-canonical-url/) on how to add that functionality to Hugo and this PR implements it, and adds a canonical URL to the doc I published yesterday.

### What should this PR do?

resolves #1942

We should be able to use canonical URLs if we want!

I'm honestly not sure exactly how this sort of thing is supposed to work. In my testing all seemed well but I'm very open to any suggestions.

### DIFF
--- a/content/chainguard/chainguard-images/recommended-practices/strategies-tools-updating-images/index.md
+++ b/content/chainguard/chainguard-images/recommended-practices/strategies-tools-updating-images/index.md
@@ -3,6 +3,7 @@ title: "Strategies and Tooling for Updating Container Images"
 linktitle: "Update Strategies and Tools"
 aliases: 
 - /chainguard/chainguard-images/recommended-practices/strategies-tools-updating-images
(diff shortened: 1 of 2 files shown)


SharpRake:
This wasn't working as expected (generating 2 canonical links). Brought it to the team and determined that we probably don't need this feature at this point in time.

amouat:
If we ressurect this, my best guess is that our theme had support for canonical URLs and the two implementations were conflicting. So we need to figure out how to do it in the theme.

PR #1946, opened 2024-12-06
Fix renovate doc

max-allan-cgr:
## Type of change
Bugfix

### What should this PR do?
Fixes some issues with the renovate doc.

### Why are we making this change?
Customer ticket : https://chainguardhelp.zendesk.com/agent/tickets/1553

### What are the acceptance criteria? 
Review

### How should this PR be tested?
Follow the doc to setup a renovate on a repo with a Dockerfile pointing to a private chainguard image.

Note: the image will need to be in a personal repo not chainguard-private because you need to create a pull token for it.

PR #1947, opened 2024-12-06
Fix json example.

amouat:
@max-allan-cgr  pointed out some issues with the JSON example.

I think he might have more improvements for the page later.


amouat:
Superseded by #1946

PR #1964, opened 2024-12-12
Added redirects.yaml file to create new redirects

smythp:
## Description

We need a redirect to depricate an istio guide. It seems we've been adding redirects to nginx.conf . Unless I'm missing somethin, we could do this a bit more cleanly when generating the redirects file rather than doing it on the proxy level. Added a data file to add manual redirects and will check if it works.

## Notes

References https://github.com/chainguard-dev/internal/issues/4498


smythp:
OK, after a little poking around it seems the preview doesn't do redirects, so we will need to merge to test (living in the fast lane).

matthewhelmke (approved):
Testing in prod. Let's go!

PR #1968, opened 2024-12-16
Updates Cosign go install instructions

smythp:
## Description

Cosign is now on v2. I've updated the Cosign go install instructions to install v2 instead of the outdated v1.

## Notes

Simplified a little of the language.

Already updated the AI course.

Resolves https://github.com/chainguard-dev/internal/issues/4458


ltagliaferri (approved):
Great, thank you!

PR #1973, opened 2024-12-18
fixes Internal #4401

matthewhelmke:
## Type of change
This adds an instruction for pre-populating the Organization name on the Chainguard Console login screen via the request URL.

### What should this PR do?
Resolves [Internal #4401](https://github.com/chainguard-dev/internal/issues/4401)

Requested by a customer

Is it clear? Does it work?

Check for typos, check the URL


amouat (approved):

amouat:
I've approved, but this seems slightly odd. Is the problem they are members of multiple orgs and want to choose a default one?

I wonder if we're add documentation that should actually be replaced with a config setting.

PR #1975, opened 2024-12-18
FIPS update

ltagliaferri:
This is an update as requested in https://github.com/chainguard-dev/internal/issues/4234


matthewhelmke (approved):
Great news and the update LGTM

PR #1978, opened 2024-12-19
Increase image limits #4514

erikaheidi:
This PR should fix our issues with blurred images, especially screenshots with text.

Issue [#4514](https://github.com/chainguard-dev/internal/issues/4514)

### DIFF
--- a/config/_default/params.toml
+++ b/config/_default/params.toml
@@ -49,8 +49,8 @@ themeColor = "#fff"
 # Images
 quality = 85
 bgColor = "#fff"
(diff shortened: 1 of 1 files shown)


erikaheidi:
Hmm the preview is not showing changes, not quite sure why as it worked locally with `hugo serve`. Might be a Netflix thing. I will investigate a bit more

erikaheidi:
Moving this to WIP because I wasn't able to figure it out yet, and I am taking some time off for the holidays.

PR #1979, opened 2024-12-20
Images section reorganization, 2024

SharpRake:
## Type of change

This PR is a massive reorganization of the Images section on academy.

### What should this PR do?

resolves https://github.com/chainguard-dev/internal/issues/4492

The images section continues to grow and has become rather unwieldy in recent months. This is an attempt to streamline information and improve discoverability.

I did my best to implement the team's suggestions so far. I also tried to be diligent about adding aliases for each doc I moved. I think a close look at everything is in order.

Just click around and let me know if I missed anything important!

preview link: https://deploy-preview-1979--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/


smythp (approved):
I like the order of the overview and initial About items.

1. It seems that a bunch of videos made it into About. Not sure if that's intended, at the very least the Beyond Zero presentation video shouldn't be there.
2. The naming scheme where the title says (Video) looks a little odd, I think, maybe "Video:"? But this might be a matter of opinion, feel free to ignore.
3. Redirects are probably the most important part here. I checked a couple, are we feeling good on those?

Cheers, thanks for doing the difficult process of taking on a lot of feedback for this reorder :)

SharpRake:
> I like the order of the overview and initial About items.
> 
>     1. It seems that a bunch of videos made it into About. Not sure if that's intended, at the very least the Beyond Zero presentation video shouldn't be there.
> 
>     2. The naming scheme where the title says (Video) looks a little odd, I think, maybe "Video:"? But this might be a matter of opinion, feel free to ignore.
> 
>     3. Redirects are probably the most important part here. I checked a couple, are we feeling good on those?
> 
> 
> Cheers, thanks for doing the difficult process of taking on a lot of feedback for this reorder :)

1. Yeah, some of those could be better elsewhere, Slacking you about this.
2. agreed! Will adjust that
3. I do feel good about the redirects. I set up `aliases` frontmatter for every doc I moved and included both the old and new locations, so everything "should" work

PR #1981, opened 2024-12-20
Angela's updates

ltagliaferri:


content/chainguard/chainguard-images/how-to-use/images-directory/index.md
```diff
@@ -1,15 +1,13 @@
 ---
-title: "Using the Chainguard Images Directory"
+title: "Using the Chainguard Directory and Console"
 linktitle: "Images Directory"
```
SharpRake, line 3: I (and @angela-zhang) think the linktitle should be either "Browsing Images" or "Using the Chainguard Directory and Console". Personally, I prefer the former just because it's shorter :smile:

SharpRake (approved):
Changes look great! Just one note about the linktitle

content/chainguard/chainguard-images/how-to-use/images-directory/index.md
ltagliaferri: I discussed with Angela that this would be done when we do the batch renaming, thanks!

PR #2003, opened 2025-01-10
Minify resources

smythp:
## Description

We're working with an SEO consultant and one of the action items is to minify our CSS, JS files.

We're working from this spreadsheet: https://docs.google.com/spreadsheets/d/1nrti84sfUC-ZzB-I0MBpBSUR6OtOgCKP/edit?gid=501769198#gid=501769198

My rading of that document suggests it's these files we're primarily concerned with on edu:

```
/js/replacer.js
/js/righthand.js
/js/rumble-comparison.js
/js/rumble-vulnerability.js
/js/rumble/base.js
/js/scrollRighthand.js
/js/snippets.js
/scss/righthand.css
/scss/rumble.css
/scss/snippets.css
```

The spradhseet has more information on what pages these particular files appear on.

I've added pipelines to minify and fingerprint these resource files

## Testing

Maybe check some reprentative number of these resources to check that they're minified. The fingerprinting adds a sha to the filename so some of these files will no longer load in the form they appear above, you'll need to navigate to a page where the resource is listed as appearing, find the resource, and load it to check that it's minifed.

## Notes

I've added some explicit config settings related to minification.

In CSS, comments don't get minified/removed in this system. I think that's because they frequently contain licenses. I don't think this will be a problem.

I tried sha512 but it generates overlong filenames, my filesystem didn't like it.

Resolves https://github.com/chainguard-dev/internal/issues/4544

### DIFF
--- a/layouts/partials/footer/script-footer.html
+++ b/layouts/partials/footer/script-footer.html
@@ -1,7 +1,7 @@
 {{ $indexTemplate := resources.Get "js/index.js" -}}
 {{ $index := $indexTemplate | resources.ExecuteAsTemplate "index.js" . | minify | fingerprint "sha512" -}}
 
-{{ $bs := resources.Get "js/bootstrap.js"  -}}
+{{ $bs := resources.Get "js/bootstrap.js" -}}
 {{ $bs := $bs | js.Build -}}
 
 {{ $highlight := resources.Get "js/highlight.js" | minify | fingerprint "sha256" -}}
--- a/layouts/partials/footer/script-footer.html
+++ b/layouts/partials/footer/script-footer.html
@@ -1,63 +1,63 @@
-{{ $indexTemplate := resources.Get "js/index.js" -}}
+{{ $indexTemplate := resources.Get "js/index.js" | minify | fingerprint "sha512" -}}
 {{ $index := $indexTemplate | resources.ExecuteAsTemplate "index.js" . -}}
 
-{{ $bs := resources.Get "js/bootstrap.js" -}}
+{{ $bs := resources.Get "js/bootstrap.js" | minify | fingerprint "sha512" -}}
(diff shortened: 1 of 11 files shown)


layouts/partials/footer/script-footer.html
```diff
@@ -1,63 +1,63 @@
 {{ $indexTemplate := resources.Get "js/index.js" -}}
-{{ $index := $indexTemplate | resources.ExecuteAsTemplate "index.js" . -}}
+{{ $index := $indexTemplate | resources.ExecuteAsTemplate "index.js" . | minify | fingerprint "sha512" -}}
 
-{{ $bs := resources.Get "js/bootstrap.js" -}}
+{{ $bs := resources.Get "js/bootstrap.js"  -}}
```
amouat, line 4: was this intentional? (Also this is the most pedantic comment I think I've ever left)

amouat (approved):
Nothing seems to have broken, so again I'm thinking we just go for this. Only comment is my own curiosity about moving from sha512 to sha256, but I don't think it's important.

layouts/partials/footer/script-footer.html
smythp: Nope, haha. I had to go back and forth on some of these lines to make sure the piplines were in the right place. I actually saw this in the PR and was too lazy to fix it.

I started making an excuse why I should leave it, but screw it, added a commit, lol.

smythp:
The sha512 to 256 is because some of the sha512 filenames pushed into the file length limits and caused Hugo to choke.

The fingerprints are to cache bust so that users get the latest and greatest when they come to the site. There's really no advantage between a 256 and 512, at 256 the chance of a collision is 2.66×10^38 and even if it collided, that would just mean one user saw one slightly old resource.

PR #2022, opened 2025-01-17
301 redirects

smythp:
## Description

Adds 301 redirects for all aliases on edu.

## Details

We've been asked to change redirects implemented in meta refresh (HTML headers) with full 301 redirects for SEO purposes. All our Hugo aliases (redirects implemented in front matter) are meta refresh redirects.

After poking around in Cloud Run, the only functionality they have for this is the load balancer. That didn't seem like the most ideal solution for a number of reasons.

This PR adds application code that takes our Hugo aliases and adds them to our nginx.conf as an included file containing rewrite directives. An "_aliases" file gets built in the output folder and is copied to the image to be included in the nginx.conf. Some regex is included to handle situations related to trailing slashes, etc. In addition, many of the aliases redirect to their own pages, which would cause a redirect loop in nginx, and I've added application logic to check for these situations and exclude those aliases.

## Testing

The Netlify preview will not assist with testing. Redirects can be tested by building the edu image locally. You can check curl headers for 301 redirects. I tried to look for edge cases but I could have missed something.

## Notes

cc @ltagliaferri  @jamonation  (for interest)

Resolves https://github.com/chainguard-dev/internal/issues/4542

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM cgr.dev/chainguard/nginx:latest
 
 COPY public/ /usr/share/nginx/html/
+COPY redirects /etc/nginx/redirects
 COPY nginx.conf /etc/nginx/nginx.conf
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM cgr.dev/chainguard/nginx:latest
 
 COPY public/ /usr/share/nginx/html/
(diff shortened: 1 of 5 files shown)


ltagliaferri (approved):
Patrick and I talked through this. We may want to add something in the future to make the nginx.conf a little nicer, but that's not a blocker now.

PR #2038, opened 2025-01-23
fix malformed link in image update strategies docs

techalchemy:


SharpRake (approved):
Thanks for catching this Dan!

Issue #2039, opened 2025-01-23
Porting a Sample Application Guide Fails on dnmonster port


techalchemy:
**Describe the bug**
Following along with the [porting a sample application guide](https://edu.chainguard.dev/chainguard/migration/porting-apps-to-chainguard/#updating-the-nodejs-microservice), I was able to build the original Dockerfile, but after I made the first round of changes to the `dnmonster` nodejs microservice using the provided Dockerfile, I was unable to successfully rebuild the microservice.

**To Reproduce**
1. Go to the [porting a sample application guide](https://edu.chainguard.dev/chainguard/migration/porting-apps-to-chainguard/#updating-the-nodejs-microservice)
2. Follow the instructions to update the nodejs microservice
3. Use the provided updated Dockerfile
4. `docker build -t 

**Expected behavior**
I would expect the build to succeed as described in the documentation. Note that the build initially fails due to claiming a need for a python runtime. I did attempt to rebuild with the python runtime added in the dockerfile, but got another failure which is what led me to file this bug.

**Initial Failure**
```console
$ docker build --pull -t dnmonster-cg .
[+] Building 46.4s (12/14)                                                                                                                                                                                                                                         docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                         0.0s
 => => transferring dockerfile: 685B                                                                                                                                                                                                                                         0.0s
 => [internal] load metadata for cgr.dev/chainguard/node:latest-dev                                                                                                                                                                                                          2.2s
 => [internal] load .dockerignore                                                                                                                                                                                                                                            0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                              0.0s
 => [ 1/10] FROM cgr.dev/chainguard/node:latest-dev@sha256:e0edbc930256c37344540f6f111f4e29a4ec01db3ddf199f418292f57c3f9620                                                                                                                                                  9.9s
 => => resolve cgr.dev/chainguard/node:latest-dev@sha256:e0edbc930256c37344540f6f111f4e29a4ec01db3ddf199f418292f57c3f9620                                                                                                                                                    0.0s
 => => sha256:193c8dda9fe8142bc27b098dd7c5cd132347f81fe2c55b352a8b19a42ff05636 248.45MB / 248.45MB                                                                                                                                                                           7.4s
[...]

PR #2059, opened 2025-02-06
Update Porting Tutorial

amouat:
## Type of change
Fixes to tutorial. 

Fixes for #2039 

If you have time to run through, that would be fantastic. Otherwise just a sanity check.

### DIFF
--- a/content/chainguard/migration/porting-apps-to-chainguard/index.md
+++ b/content/chainguard/migration/porting-apps-to-chainguard/index.md
@@ -9,7 +9,7 @@ description: "This article works through porting a small but complete applicatio
 we'll see, this is relatively straightforward, but it is important to be aware of some of the
 differences to other common images."
 date: 2024-04-10T12:56:52-00:00
-lastmod: 2024-08-08T14:44:52-00:00
+lastmod: 2025-02-05T14:44:52-00:00
 draft: false
 tags: ["Images", "Product", "Procedural"]
 images: []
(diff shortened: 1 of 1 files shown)


amouat:
Actually, let me update the python section as well. More updates incoming.

content/chainguard/migration/porting-apps-to-chainguard/index.md
```diff
@@ -539,35 +544,36 @@ And take a look at it:
...
+   ├── ✔ File metadata                   [1,667 locations]
+   └── ✔ Executables                     [135 executables]
+ ✔ Scanned for vulnerabilities     [1 vulnerability matches]
+   ├── by severity: 0 critical, 0 high, 0 medium, 0 low, 0 negligible (1 unknown)
+   └── by status:   0 fixed, 1 not-fixed, 0 ignored
+NAME         INSTALLED  FIXED-IN  TYPE  VULNERABILITY  SEVERITY
+python-3.13  3.13.1-r5            apk   CVE-2025-0938  Unknown
+```
+
+The result of all these changes is that the production image is only 122 MB (down from 1.51GB, so an enormous saving of over a GB) and has 1 CVEs (down from hundreds). This is a huge improvement!
```
matthewhelmke, line 569:
```suggestion
The result of all these changes is that the production image is only 122 MB (down from 1.51GB, so an enormous saving of over a GB) and has 1 CVE (down from hundreds). This is a huge improvement!
```

matthewhelmke (approved):
One small change, otherwise LGTM. Thanks!

PR #2067, opened 2025-02-12
chore: simplify the autodocs workflow

IdlePhysicist:
This commit removes checking from the autodocs workflow.  The workflow was checking to see if a changelog file has changed, though we no longer update that changelog file.

### DIFF
--- a/.github/workflows/autodocs-platform.yaml
+++ b/.github/workflows/autodocs-platform.yaml
@@ -12,6 +12,8 @@ jobs:
   integrate-platform-docs:
     runs-on: ubuntu-latest
 
+    if: github.repository == 'chainguard-dev/edu'
+
     permissions:
       contents: read # reads from the repo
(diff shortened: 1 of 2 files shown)


smythp (approved):
I've pushed to our staging site and shared, but I suspect we'll have to try this live. Approving.

.github/workflows/autodocs-platform.yaml
```diff
@@ -7,78 +7,15 @@ on:
...
     branches:
       - platform-docs
-env:
-  PLATFORM_CHANGELOG: "autodocs/platform-changelog.md"
 
 jobs:
-  check-new-docs:
-    runs-on: ubuntu-latest
-
-    if: github.repository == 'chainguard-dev/edu'
```
IdlePhysicist, line 17: Last minute thing, I'm going to put this into the `integrate-platform-docs` job below.  Don't want forks running this workflow

PR #2090, opened 2025-02-26
Fix warning when building dockerfile.

amouat:
 
Fix to Dockerfile to suppress warning message. As suggested by @kallewesterling


kallewesterling (approved):
LGTM!

One small thing:
- "you may want to consider changing to use the" might be shortened to something like "you may want to ~~consider changing to~~ use the" on L184

PR #2093, opened 2025-02-28
fixes to example Dockerfile (per AM)

SharpRake:


matthewhelmke (approved):
LGTM, thanks!

PR #2099, opened 2025-03-04
Fix typo.

amouat:
Also reword to avoid double "access".


kallewesterling (approved):
Looks good to me!

PR #2101, opened 2025-03-05
Updating Postgres Getting Started guide with Configuration options

SharpRake:

## Type of change

This PR adds a new section to the PostgreSQL Getting Started guide outlining how you can configure the Postgres image.

### What should this PR do?

resolves https://github.com/chainguard-dev/internal/issues/4619

This is part of fulfilling a customer request for improved documentation for our Postgres image.

Content was approved for the image README already, but open to any feedback.

### DIFF
--- a/content/chainguard/chainguard-images/getting-started/postgres/index.md
+++ b/content/chainguard/chainguard-images/getting-started/postgres/index.md
@@ -202,10 +202,10 @@ Of course, you likely won't be regularly managing your containerized databases o
 
 ### Configuring the PostgreSQL image
 
-You can extend Chainguard's Postgres image with environment variables. Chainguard's Postgres image is compatible with the environment variables available in the official PostgreSQL image, including the following:
+You can extend Chainguard's PostgreSQL image with environment variables. Chainguard's PostgreSQL image is compatible with the environment variables available in the official PostgreSQL image, including the following:
 
 * `PGDATA`: This variable allows you to define another location for database files. The default data directory is `/var/lib/postgresql/data`.
-* `POSTGRES_PASSWORD`: This environment variable sets the superuser password for PostgreSQL. This variable is required to use the Postgres image.
+* `POSTGRES_PASSWORD`: This environment variable sets the superuser password for PostgreSQL. This variable is required to use the PostgreSQL image.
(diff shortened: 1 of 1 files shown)


amouat:
I ran the demo and couldn't get it work. In compose the logs were:

```
app-1  | [06-Mar-2025 10:01:34] NOTICE: PHP message: PHP Warning:  PHP Startup: Unable to load dynamic library 'pgsql.so' (tried: /usr/lib/php/modules/pgsql.so (/usr/lib/php/modules/pgsql.so: undefined symbol: zend_register_internal_class_with_flags), /usr/lib/php/modules/pgsql.so.so (/usr/lib/php/modules/pgsql.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
app-1  | [06-Mar-2025 10:01:34] NOTICE: [pool www] 'user' directive is ignored when FPM is not running as root
app-1  | [06-Mar-2025 10:01:34] NOTICE: [pool www] 'group' directive is ignored when FPM is not running as root
app-1  | [06-Mar-2025 10:01:34] NOTICE: fpm is running, pid 8
app-1  | [06-Mar-2025 10:01:34] NOTICE: ready to handle connections
app-1  | NOTICE: PHP message: PHP Fatal error:  Uncaught Error: Call to undefined function pg_connect() in /app/public/index.php:3
app-1  | Stack trace:
app-1  | #0 {main}
app-1  |   thrown in /app/public/index.php on line 3
app-1  | 172.20.0.3 -  06/Mar/2025:10:03:16 +0000 "GET /index.php" 500
app-1  | 172.20.0.3 -  06/Mar/2025:10:03:22 +0000 "GET /index.php" 500
app-1  | NOTICE: PHP message: PHP Fatal error:  Uncaught Error: Call to undefined function pg_connect() in /app/public/index.php:3
app-1  | Stack trace:
app-1  | #0 {main}
app-1  |   thrown in /app/public/index.php on line 3
```

amouat (approved):
~~Can you double check the bit on variables only having an effect with empty databases? I wonder if it's just some variables are ignored.~~

It's correct! I guess it makes sense, I suppose you can't change passwords etc on a existing db.

SharpRake:
> PHP message: PHP Warning:  PHP Startup: Unable to load dynamic library 'pgsql.so'

@amouat Noted. I'm hitting the same errors. I'm not sure what the cause is (and it looks like I'm the one who first added this example, which is a very bad sign in and of itself).

I will publish these changes while chasing down a solution for the errors.

PR #2106, opened 2025-03-07
update gcr doc

ltagliaferri:
Updating from this [Slack thread](https://chainguard-dev.slack.com/archives/C03H64P08UT/p1741299752918959)


SharpRake (approved):
I made a few tweaks to reflect an upstream change to the example application. moving into the new iac directory wasn't necessary so I removed that too.

PR #2111, opened 2025-03-10
adding new draft of dev containers doc

SharpRake:

## Type of change

This PR adds my edited version of @max-allan-cgr's original guide on using Chainguard images as dev containers.

### What should this PR do?

resolves https://github.com/chainguard-dev/internal/issues/4607

Dev containers are useful tools and Chainguard Images provide a secure way to use them for development.

I will ask Max to review it for content before getting a close review for style, etc.

Tech test would be helpful.

Relevant preview link: https://deploy-preview-2111--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/how-to-use/dev-containers/


content/chainguard/chainguard-images/how-to-use/dev-containers/index.md
```diff
@@ -0,0 +1,223 @@
...
+    * You can check by running your image locally and using the `id` command. The following example shows that the Python image uses the nonroot user by default while the Go image uses root by default:
+
+``` shell
+docker run -it --entrypoint id chainguard/python:latest-dev
+```
+```Output
+uid=65532(nonroot) gid=65532(nonroot) groups=65532(nonroot)
+docker run -it --entrypoint id chainguard/go:latest-dev
+uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
+```
```
kallewesterling, line 191: Do we indent example code like this, that belongs to the bullet point above? I couldn't see anything about this in the style guide.

content/chainguard/chainguard-images/how-to-use/dev-containers/index.md
```diff
@@ -0,0 +1,223 @@
...
+EOF
+```
+
+Here's what each line of this Dockerfile does:
+
+* `FROM`: Many Chainguard images are available at Docker Hub, meaning you can set the `FROM` line in the `Dockerfile` like this example. Alternatively, you could set this with `cgr.dev/chainguard/python` or, if your Chainguard organization has access to a Production Python image, `cgr.dev/$ORGANIZATION/python:$TAG`.
+* `USER root`: Including this line forces the next commands to run as root instead of the unprivileged user. Whether you need to include this line or not will depend on what image you're using.
+    * You can check by running your image locally and using the `id` command. The following example shows that the Python image uses the nonroot user by default while the Go image uses root by default:
+
+``` shell
```
kallewesterling, line 184: \`\`\`shell should not have a space here, I believe?

kallewesterling (approved):
Looks great with the exception of two small comments!

content/chainguard/chainguard-images/how-to-use/dev-containers/index.md
SharpRake: I didn't think indenting code blocks was possible, but apparently it is!  Thanks for flagging, it looks much much better now.

I don't think there's anything specific about this in the style guide. Maybe worth adding something though.

content/chainguard/chainguard-images/how-to-use/dev-containers/index.md
SharpRake: :100:

max-allan-cgr (approved):

PR #2112, opened 2025-03-12
Center images and figures

smythp:
## Description

This PR centers images on Academy (figures) created using the markdown syntax Alt text

It also sets a max height for images. The width is set automatically if the height is more than 45em (that's a size that scales on different screens, it's a function of the line height). I went with a pretty big height so that really really tal screenshots don't get too small, but it should prevent the gigantic ones.  It's hard to come up with behavior that works for both vertical and horizontal screenshots so I tried to be conservative.

One issue with the okta page is that it uses HTML  tags. This means the hugo doks CSS classes dont' get placed on the image. I think we should try to stick to the markdown syntax swhere possible, if we use the img tags we'll have to put the classes on manually. If this is a major issue I can  come up with something but I think we use the markdown syntax most places. I've swapped the  tags to markdown syntax in that tutorial in this PR.

## Testing

It would be helpful to poke around and make sure this isn't centering any other images that we didn't expect. I didn't see anything like this and I believe these classes should only be applying in article text.

I'd appreciate someone more visually orientied looking at a wide variety of screenshots. Make sure there's no scaling (making images bigger which would make them blurry) and no stretching, I've aimed to prevent these.

## Notes

@SharpRake  Going to ask you to review since this was your issue.

Resolves https://github.com/chainguard-dev/internal/issues/4493

### DIFF
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -1243,3 +1243,10 @@ blockquote p a {
     margin-top: 1em;
     margin-bottom: -3em;
 ;}
+
+
+.figure, .figure-img, .img-fluid {
+    display: block !important;
+    margin: auto !important;
+
+}
--- a/assets/scss/common/_custom.scss
+++ b/assets/scss/common/_custom.scss
@@ -1245,8 +1245,22 @@ blockquote p a {
 ;}
 
 
+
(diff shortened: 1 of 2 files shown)


SharpRake (approved):
Looks good to me! Thanks so much @smythp!

PR #2138, opened 2025-03-25
chainctl: fix cosign verify command

wlynch:
I think the versioning output may have changed over time, so this was returning `v0.1.2` rather than `0.1.2`, throwing off the URLs resulting in the error:

Error: error during PEM decoding
main.go:74: error during command execution: error during PEM decoding

This fixes the version so the cert can be fetched.


SharpRake (approved):
thank you @wlynch!

PR #2156, opened 2025-04-03
Moving maintenance tool to new location and renaming it

erikaheidi:
This PR moves our experimental maintenance tool to a new location, to avoid confusion with the "Autodocs" workflow that updates chainctl reference docs.

Historically, we had another tool called Autodocs that was responsible for building public-facing images docs, but this has been decomissioned in july 2024 when image docs was moved under engineering.

When creating the first iteration of our maintenance tool, I used the same "autodocs" directory for simplicity. But the current chainctl reference docs workflow uses the same directory for logs, so to avoid confusion and to better organize our repository, I moved the maintenance tool under "tools" and renamed it to **chaindocs**.

We'll soon bootstrap a new maintenance sprint and the tool will need some updates, which I am scoping now.

## Testing the PR
This is more of an organizational change, there is not much to test here. To try the tool itself you need a PHP environment for now ( working on a Dockerfile ). Run `composer install`  to install dependencies and then you can execute the tool with `./chaindocs audit` .


matthewhelmke (approved):
LGTM, thank you for taking care of this!

PR #2169, opened 2025-04-07
add callout to privileges necessary in the Azure console

jimmy-ungerman:
- [x] Check if this is a typo or other quick fix and ignore the rest :)

Adding users to an Entra app requires admin permissions in Azure. Without them, you'll be able to do every step on this page, but you will not be able to add any users to the new application. This PR calls out to setup the integration as an Admin user in Azure


kallewesterling (approved):
LGTM

PR #2179, opened 2025-04-08
Update cosign installation instructions

priyawadhwa:
This PR updates cosign installation instructions to the latest version of cosign. It's ok if this falls a bit out of date, the main reason for the change is to make sure people are installing Cosign V2 instead of Cosign V1.


ltagliaferri (approved):
Thank you, Priya!

PR #2181, opened 2025-04-08
Correct order in Product Docs navigation

mosabua:
### What should this PR do?

Set the order in the Product Docs navigation section so it makes sense

* Containers
* Libraries
* Registry
* Migration
* Administration
* chainctl Usage
* chainctl Reference

Also note that I am NOT updating timestamps since this is not a content change.

Somehow adding the timestamps for Libraries bumped it to the very top and it should not be above all the container related content.

review

run locally and verify

### DIFF
--- a/content/chainguard/administration/_index.md
+++ b/content/chainguard/administration/_index.md
@@ -7,5 +7,5 @@ date: 2023-10-26T08:48:45+00:00
 lastmod: 2023-10-26T08:48:45+00:00
 draft: false
 images: []
-weight: 40
+weight: 050
 ---
--- a/content/chainguard/chainctl-usage/_index.md
+++ b/content/chainguard/chainctl-usage/_index.md
@@ -8,7 +8,7 @@ lastmod: 2025-03-0320T08:49:15+00:00
 draft: false
 tags: ["chainctl", "Usage", "Product", "Overview"]
 images: []
-weight: 50
+weight: 060
 ---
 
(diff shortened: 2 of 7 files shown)


ltagliaferri:
What about putting Libraries under Containers at least for the time being?

mosabua:
> What about putting Libraries under Containers at least for the time being?

You mean nested within? .. that would be wrong in my opinion since its a new top level product .. same as VMs soon

Or do you mean right after Containers .. I eventually we want that because Registry and Migration should be under Containers (nested) but for now I think they (Containers, Registry, and Migration) belong together..

ltagliaferri:
> Or do you mean right after Containers .. I eventually we want that because Registry and Migration should be under Containers (nested) but for now I think they (Containers, Registry, and Migration) belong together and therefore Libraries is below

Yeah, I mean right after Containers because it is the major product, with Registry being secondary. Migration I think can include Libraries or at least VMs eventually.

mosabua:
I am fine with that too although it splits the container related content apart.. ultimately I think Registry should go inside Containers (same as we have repo manager stuff in Libraries) and migration probably also should go inside.. and we might have migration inside of Libraries and VMs eventually as well..  I dont think Migration needs to be a top level item on its own

mosabua:
Updated now .. do you want me to send a PR to move the Registry stuff @ltagliaferri ?

smythp (approved):
LGTM

ltagliaferri (approved):

PR #2185, opened 2025-04-09
Add a docs page for the new chainctl images history command

matthewhelmke:

## Type of change
Add a new page

### What should this PR do?
Fixes https://github.com/chainguard-dev/internal/issues/4705

Is it accurate? Is it clear?

I would like one of @mgreau or @blancadesal to review for accuracy as stakeholders.

For the DevEd team review, does it conform to our style guide, it is clear, etc.

### DIFF
--- a/content/chainguard/chainctl-usage/chainctl-images-history.md
+++ b/content/chainguard/chainctl-usage/chainctl-images-history.md
@@ -122,11 +122,16 @@ The command returns a reverse-chronological history of when a specific tag was u
 
 You aren't required to include the release version tag. If you omit it, you will be presented with a menu, like this:
 
+
+When the release version tag is not provided, the command will present you with a menu that lets you select which tag you'd like to obtain the history for. For example:
+
+
 ```
 chainctl images history python --parent=chainguard.edu
 ```
 
-Presents you with this:
+This will present you with a menu similar to this:
+
 
 ```
     Which tag of python would you like to view history for?  
--- a/content/chainguard/chainctl-usage/chainctl-images-history.md
+++ b/content/chainguard/chainctl-usage/chainctl-images-history.md
@@ -118,7 +118,7 @@ The returned list is longer than we'll show here, but here's an excerpt:
 
 ```
 
-The command returns a reverse-chronological list of image builds with timestamps, digests for the build processes, and digests for specific architecture builds with build sizes.
+The command returns a reverse-chronological history of when a specific tag was updated to point to a new manifest digest. If images are not multi-arch, only a single digest without architecture will be displayed.
 
 You aren't required to include the release version tag. If you omit it, you will be presented with a menu, like this:
 
--- a/content/chainguard/chainctl-usage/chainctl-images-history.md
+++ b/content/chainguard/chainctl-usage/chainctl-images-history.md
@@ -32,8 +32,8 @@ You will be presented with a list of organizations available to your account. Fo
     Which organization would you like to list images from?                                                       
                                                                                                                         
   > [MyCorp-prod]     This group holds the production Chainguard Images hosted under    cgr.dev/MyCorp-prod                   
-    [MyCorp-starter]     This group holds the starter Chainguard Images hosted under    cgr.dev/chainguard-starter  
-    [MyCorp-eval]     This group holds the evaluation Chainguard Images hosted under    cgr.dev/chainguard-eval  
+    [MyCorp-starter]     This group holds the starter Chainguard Images hosted under    cgr.dev/MyCorp-starter  
+    [MyCorp-eval]     This group holds the evaluation Chainguard Images hosted under    cgr.dev/MyCorp-eval  
 ```
 
 Move the `>` up and down with the arrow keys and hit `Enter` to select the org you want to you.
--- a/content/chainguard/chainctl-usage/chainctl-images-history.md
+++ b/content/chainguard/chainctl-usage/chainctl-images-history.md
@@ -1,0 +1,121 @@
+---
+title: "chainctl images history"
+lead: ""
+description: "chainctl images history basics"
+type: "article"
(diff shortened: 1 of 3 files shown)


blancadesal:
Thank you for working on this!

I think it might be worth mentioning in the docs that only the IMAGE argument is strictly necessary. If a user is member of more than one org, there's a selector menu just like in `images list`. And if no tag is specified, there's another selector menu for that:

```
$ chainctl images history python

   ✔ Selected organization some-org

    Which tag of python would you like to view history for?

  > 3
    3-dev
    3.10
    3.10-dev
    3.10.11
    3.10.11-dev
    3.10.11-r0
    3.10.11-r0-dev
```

wdyt?

matthewhelmke:
> wdyt?

Oh! That's a great thing to add. I didn't even think to test the command without adding a tag. Nice!

matthewhelmke:
I've added that in, @blancadesal , thanks again! How's the rest of the info look? Complete? Accurate?

content/chainguard/chainctl-usage/chainctl-images-history.md
```diff
@@ -0,0 +1,150 @@
...
+    arm64: sha256:f76a0a2f49418b030f3a31bcd2c8bddb8bbef7b006006aa59b74282955ab671d (22.41 MB)
+- time: 2024-10-01 23:40:56 UTC
+  digest: sha256:0b0daf09eeb92741efe0eae51dbdeea5a66bed870e0e00895630de729b233b7f
+  architectures:
+    amd64: sha256:637af1b20e5f8cee7e538b07a1ae3934297769216a65acc454e34fac3dcd3828 (23.46 MB)
+    arm64: sha256:8dce068942fa4dd155c87b6b8a3e4b8e2482a5fdb5232cb9dd73c39e63003038 (22.41 MB)
+
+```
+
+The command returns a reverse-chronological list of image builds with timestamps, digests for the build processes, and digests for specific architecture builds with build sizes.
```
blancadesal, line 121: small nit: each entry often corresponds to a build, but technically, it's a history of when a specific _tag_ was updated to point to a new manifest digest.

suggestion (feel free to ignore): slightly rephrasing "list of image builds" to something like "list of tag updates" or "history records for the tag" for technical precision.

blancadesal (commented):
LGTM overall, small nit inline.

I'm not sure if this is worth mentioning, but for non-multi-arch images, only a single digest without architecture information will be displayed (mentioned in the command's long description).

content/chainguard/chainctl-usage/chainctl-images-history.md
matthewhelmke: This is exactly the kind of feedback I love. Very helpful, thank you! I'll update and get this moving ahead. Maybe we can get it live today!

blancadesal (approved):
LGTM! 🚀

content/chainguard/chainctl-usage/chainctl-images-history.md
```diff
@@ -0,0 +1,150 @@
...
+tags: ["chainctl", "images", "history", "Product"]
+images: []
+weight: 75
+---
+
+This page demonstrates how to use the `chainctl images history` command. For a full reference of all commands with details and switches, see [chainctl Reference](/chainguard/chainctl/).
+
+At times it may be useful to examine the history of an image tag in chainctl, such as to find when it was updated and the associated digests for each update. This is done using `chainctl images history`, which is part of the [chainctl images](./chainctl-images.md) set of commands.
+
+Let's walk through a typical scenario. In our example we use an internal organization created for the Developer Enablement team to use when creating educational resources. Change our examples to use your org. You can do this by selecting from the options presented in a menu as you issue commands, as in [List chainctl images](#list-chainctl-images)
```
kallewesterling, line 18: > In our example we use an internal organization created for the Developer Enablement team to use when creating educational resources

Looks like we have "MyCorp" the example below. Should this be edited to reflect what we actually do in the below instructions?

kallewesterling (approved):
just one question mark - if I'm wrong, feel free to override :)

content/chainguard/chainctl-usage/chainctl-images-history.md
matthewhelmke: Doh! Yes. I started out using the list of orgs I have access to and decided to make it more obvious to the reader that they won't have the same org list. I'll fix this.
TY!!

content/chainguard/chainctl-usage/chainctl-images-history.md
```diff
@@ -0,0 +1,150 @@
...
+  digest: sha256:0b0daf09eeb92741efe0eae51dbdeea5a66bed870e0e00895630de729b233b7f
+  architectures:
+    amd64: sha256:637af1b20e5f8cee7e538b07a1ae3934297769216a65acc454e34fac3dcd3828 (23.46 MB)
+    arm64: sha256:8dce068942fa4dd155c87b6b8a3e4b8e2482a5fdb5232cb9dd73c39e63003038 (22.41 MB)
+
+```
+
+The command returns a reverse-chronological history of when a specific tag was updated to point to a new manifest digest. If images are not multi-arch, only a single digest without architecture will be displayed.
+
+You aren't required to include the release version tag. If you omit it, you will be presented with a menu, like this:
```
erikaheidi, line 123: I would just maybe rephrase this bit slightly to be more descriptive instead of relying on "like this". something like "When the release version tag is not provided, the command will present you with a menu that lets you select which tag you'd like to obtain the history for. For example:" then add the command, followed by "This will present you with a menu similar to this:" -- sort of ack'ing that this can change as new versions are released.

matthewhelmke: I like that. I'll edit this. TY!

PR #2190, opened 2025-04-10
(WIP) Move Chainguard Registry under Containers

mosabua:
### What should this PR do?

De-emphasize Chainguard Registry by moving it below Containers in the nav and in the structure.

Potentially also just renaming to registry .. since it is just a means to and end for using containers and not really a product or so. 

Also to be consistent we would also think about about the different repository/registry support for Chainguard Libraries for Java, Python, JavaScript and beyond - and it is better NOT to do that.

See above

review

locally with container


mosabua:
Turns out that nginx.conf contains:
```
"~^/chainguard/chainguard-images/registry/(.+)?$"  /chainguard/chainguard-registry/;
```

Which is the location I am currently moving it to .. I think this would create an infinite redirect loop for some users .. 

Maybe we have to hold this and do a bigger and better path refactor to 

```
/chainguard/containers/
```

and hence also 

```
/chainguard/containers/registry
```

smythp:
We spoke about this on Slack, but I don't think we should redirect the images path, it has too much under it and the benefit will be minimal. It has a large percentage of our total docs under it.

If there's a redirect to a location and your moving the content to that location, you can just delete the redirect. that was there before. 

Since we're moving a whole folder, we should just redirect the higher portion of the path, not each individual file. In theory this PR should be just a few lines

I think we can backburner this for a bit and come back to it when the Libraries work cools down.

mosabua:
We decided to delay this .. whenever we pick it back up a new branch and PR will be easier than working with this

PR #2193, opened 2025-04-10
Add init.gradle example

mosabua:
### What should this PR do?

Add a working example for a init.gradle file.

Local testing revealed that the existing docs are not sufficient in detail.

Output from https://github.com/chainguard-dev/internal/issues/4847

Review

Could be tested locally if desired. I did that already with a local Nexus and a test project.

### DIFF
--- a/content/chainguard/libraries/java/build-configuration.md
+++ b/content/chainguard/libraries/java/build-configuration.md
@@ -306,7 +306,7 @@ Gradle can also be configured to use a local Maven repository with a repository
 configuration in the global `init.gradle` or a project specific `build.gradle`
 file:
 
-```
+```groovy
 repositories {
    ...
     mavenLocal()    
@@ -333,7 +333,7 @@ and any other repositories, and adds a replacement definition with the URL of th
 repository group or virtual repository from your repository manager
 `https://repo.example.com/group/` and any applicable authentication details.
 
-```
+```groovy
 repositories {
     maven {
         url = uri("https://repo.example.com/group/")
@@ -358,7 +358,7 @@ Libraries access](/chainguard/libraries/access/) replacing the placeholders
 repository is located above the `mavenCentral` repository and any other
 repositories:
 
-```
+```groovy
 repositories {
     maven {
         url = uri("https://libraries.cgr.dev/maven/")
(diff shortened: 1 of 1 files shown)


content/chainguard/libraries/java/build-configuration.md
```diff
@@ -371,6 +371,34 @@ repositories {
...
+The following listing shows a valid `init.gradle` file. It wraps the
+`repositories` element with `allprojects` so that the scope of the file affects
+all projects built locally with Gradle. It also allows for `buildscript`
+download from the remote URL. Lastly the example shows use of an internal
+repository manager that only serves artifacts without authentication using HTTP
+only. Since this is not advisable unless other networking setups allow a secure
+use with HTTP, the override with the property `allowInsecureProtocol` is
+required:
+
+```
```
SharpRake, line 383: One thing to be aware of is that if you don't add a label to code blocks they will be labeled with `Output` once the article is published. Not sure what you'd want to put here but I think you ought to add some label so that doesn't happen.

content/chainguard/libraries/java/build-configuration.md
```diff
@@ -371,6 +371,34 @@ repositories {
 }
 ```
 
+The following listing shows a valid `init.gradle` file. It wraps the
+`repositories` element with `allprojects` so that the scope of the file affects
+all projects built locally with Gradle. It also allows for `buildscript`
```
SharpRake, line 376: The sentence "It also allows for `buildscript` download from the remote URL." doesn't read correctly to me, like it should be one of the following:

"It also allows for `buildscript` to be downloaded from the remote URL."

or 

"It also allows for `buildscript` downloads from the remote URL."

I'm not clear on which of those are correct, or even if the original phrasing is correct.

SharpRake (approved):
Two comments, but it looks good otherwise. Let me know if you have questions or want me to take another look!

content/chainguard/libraries/java/build-configuration.md
mosabua: Thanks for the help on slack with that ..

content/chainguard/libraries/java/build-configuration.md
mosabua: Reworded to make it more explicit.

PR #2201, opened 2025-04-14
Use standard UTM medium value.

amouat:

It is a typo fix I think, but it's a _lot_ of them. I think UTM medium should be "referral", not "website". This is causing Google Analytics to mark visits as unassigned.

Please double check I'm correct about this and my search and replace doesn't have any casualties.


kallewesterling (approved):
see comment on Slack about `utm_source=docs` appearing alongside `utm_source=cg-academy`

amouat:
There was a bit of discussion about this, but no good reason not to change, so merging now.

PR #2207, opened 2025-04-15
Fix more utm links.

amouat:
A bunch more UTM fixes, this time for ones without "utm_medium" set.

Note this does make some of the URLs long, but I don't think that matters.


kallewesterling (approved):
Did spot checks; looks good!

amouat:
Thanks! Let's try it then.

PR #2215, opened 2025-04-17
Updating tool name, converter vs conversion

erikaheidi:

Change to "Dockerfile Converter Tool" instead of "Dockerfile Conversion Tool", request by @tonyjcamp 

The URL was not changed in order to avoid having redirects, and because the slug makes sense as dockerfile-conversion.


kallewesterling (approved):
LGTM

PR #2221, opened 2025-04-24
Merge chainctl update and version pages

matthewhelmke:
## Type of change
Update info and improve user experience in docs.

### What should this PR do?
Fixes https://github.com/chainguard-dev/internal/issues/4726 by merging the info in the short pages for `chainctl version` and `chainctl update` into one more useful page.

Is the info clear?

Do we need redirects from the old page URLs to the new page? They pages didn't exist more than a month or two, so I kind of think not.


content/chainguard/chainctl-usage/chainctl-version-update.md
```diff
@@ -0,0 +1,141 @@
...
+
+## View your chainctl version
+
+To see which `chainctl` version you have installed, use:
+
+```shell
+chainctl version
+```
+
+This command tells you more than just a release number. The release is shown here as `GitVersion`.
```
kallewesterling, line 27: Do we need to explain what the `GitVersion` denotes?

content/chainguard/chainctl-usage/chainctl-version-update.md
```diff
@@ -0,0 +1,141 @@
...
+To update your `chainctl` installation to the latest release, use:
+
+```shell
+chainctl update
+```
+
+Updating requires administrative privileges, so be prepared to enter your machine's admin password.
+
+```shell
+sudo chainctl update
```
kallewesterling, line 80: Is there a need here for a note here about running it in `sudo` mode rather than the above example?

kallewesterling (approved):
Just a few questions about a few potential clarifications needed

smythp:
I think we're good without redirects here. There's likely no SEO or external links to preserve and redirects aren't without cost.

content/chainguard/chainctl-usage/chainctl-version-update.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Update your chainctl install
+
+To update your `chainctl` installation to the latest release, use:
+
+```shell
+chainctl update
+```
+
+Updating requires administrative privileges, so enter as `root` or use `sudo` and be prepared to enter your machine's admin password.
```
mosabua, line 77: This is not true

mosabua: At least not on macos with the brew install

content/chainguard/chainctl-usage/chainctl-version-update.md
mosabua: The sentence above says why .. but in my experience thats not true...

content/chainguard/chainctl-usage/chainctl-version-update.md
matthewhelmke: That's an interesting point. I only have linux machines and tested on those, but didn't think to test on macos (mea culpa).

What do you think if I change that to "Updating on some operating systems requires..."?

matthewhelmke: I could also omit the admin bit completely. It's not like people who are confronted with the need won't know what to do.

matthewhelmke: I deleted the mention https://github.com/chainguard-dev/edu/pull/2222

PR #2222, opened 2025-04-24
chainctl update does not always require admin rights

matthewhelmke:

See [this comment](https://github.com/chainguard-dev/edu/pull/2221#discussion_r2058762949) for why this is being pushed. It turns out that admin privileges are not needed in some instances.


kallewesterling (approved):
lgtm

PR #2239, opened 2025-05-02
Fix another alias.

amouat:

Add alias for chainctl-config page


matthewhelmke (approved):
LGTM, tested and worked in preview

PR #2250, opened 2025-05-07
Combine and improve chainctl images pages

matthewhelmke:
Fixes https://github.com/chainguard-dev/internal/issues/4724 by transforming the images page from more of a reference to something more like a short tutorial.

This also deletes a page dedicated to the `chainctl images history` sub-command and inserts the information into the main page. Does this one need a redirect, @smythp and @ltagliaferri ? The history page was released with a bit of fanfare as a cooperative effort and I would hate to unnecessarily lose the goodwill. At the same time, it really doesn't need it's own page.


content/chainguard/chainctl-usage/chainctl-images.md
```diff
@@ -11,21 +11,64 @@ images: []
 weight: 070
 ---
 
-This page presents some of the more commonly used `chainctl images` commands. For a full reference of all commands with details and switches, see [chainctl Reference](/chainguard/chainctl/).
+This page presents some of the more common ways to learn about the Chainguard Containers that are available to you. We use `chainctl images` commands to list available repositories and container images, examine images more closely, and compare them to one another.
```
erikaheidi, line 14: I know the product name has changed, but in some contexts I feel it can be confusing; in my first pass on this sentence I imagined a command that worked like `docker ps`. I would probably use "Chainguard Container Images" in this spot specifically to make it clearer that we're talking about images.

erikaheidi:
I reviewed the content just in case, had only one comment to add. Will defer the approval to @ltagliaferri / @smythp because of the redirect question.

content/chainguard/chainctl-usage/chainctl-images.md
mosabua: I dont think we should introduce a new title-cased product name like "Chainguard Container Images" .. I think the proposed change is good.

matthewhelmke: Thank you. I think both of you are right. Much appreciated!

ltagliaferri (changes requested):
I would do the alias redirect

ltagliaferri (approved):

PR #2260, opened 2025-05-13
Remove unnecessary page

matthewhelmke:
Fixes https://github.com/chainguard-dev/internal/issues/4725

This page was super short. The one command on the page fits well in the [Get Started with chainctl](https://edu.chainguard.dev/chainguard/chainctl-usage/getting-started-with-chainctl/) page. I decided that is the best course of action, but as I went to add the command into the get started page, I found that I had already inserted it in a previous PR.

This PR is just cleaning up by removing the unneeded page and finishing the outstanding issue.


erikaheidi (approved):
lgtm!

PR #2262, opened 2025-05-14
Add link to global configuration.

amouat:
Added some context to point users to global config in first case.


smythp (approved):
This is a good addition, thank you.

content/chainguard/libraries/python/overview.md
```diff
@@ -22,6 +22,13 @@ Chainguard Libraries for Python enables access to a growing collection of Python
 
 ## Technical Details
 
+Most organisations will consume Chainguard Libraries for Python through a repository manager such as
```
mosabua, line 25: Avoid future tense and we should keep this consistent with the Java one ..

PR #2296, opened 2025-06-02
add chainctl v console page

matthewhelmke:
This adds a new page to the `chainctl` Usage section in Academy. When ready it will close https://github.com/chainguard-dev/internal/issues/4687 .

PR #2299, opened 2025-06-03
remove directory that is no longer needed

matthewhelmke:
The `terminal-images` directory is old and no longer used. It is causing build errors [like this one](https://github.com/chainguard-dev/edu/pull/2296#issuecomment-2935003675). After a team discussion on Slack, we decided it can be removed to see if that eliminates the problem.


mosabua (approved):
Bye bye bye by bye

PR #2301, opened 2025-06-03
Correct manual access verification

mosabua:
### What should this PR do?

- Adjust to the new use of redirects
- Add info for netrc and env variable use
- Link to Python and Java section

Old info no longer works and was incomplete

review

Could be manually reproduced. I did that already myself.


wiedenmeier (approved):
+1 on using `-L`

matthewhelmke (approved):
LGTM

PR #2302, opened 2025-06-03
nonce attempt

ltagliaferri:
nonce implementation attempt


mosabua:
Pr seems to mix a few things .. what the motivation to add a csp nonce ?

PR #2316, opened 2025-06-09
network requirements: add virtualapk.cgr.dev

imjasonh:

## Type of change

Adding a network requirement for a new endpoint we're going to be communicating to users soon.

### What should this PR do?

Add a documented network requirement

To let users know we'll be serving packages from an additional endpoint.

Docs look good

Check the preview


SharpRake (approved):
Addition looks good to me! Thanks @imjasonh

PR #2323, opened 2025-06-16
add note to tag history api page re using chainctl

matthewhelmke:
Fixes https://github.com/chainguard-dev/internal/issues/4868 by adding a note on the API page with guidance on considering using `chainctl images history` for quick checks.

The request was to rewrite the page to use only this command, but it's likely there are customers using the API and others who have use cases where it is still useful. Since the chainctl command is already documented, pointing to that as an alternative seems a reasonable compromise.


kallewesterling (approved):
LGTM

PR #2333, opened 2025-06-18
Small edits for clarity, added one step to adding virtual repository

smythp:
## Description

I ran through Artifactory setup and local config for pip for Learning Lab, these are some small edits on that path. Most notably added a (pretty obvious) step to the section on adding a virtual repository. I made one other edit in the assumption that some would be following the docs without being the same person who set up the repositories (i.e. not assume that they used our default name, but sharing what the default name is).

cc @mosabua

### DIFF
--- a/content/chainguard/libraries/python/build-configuration.md
+++ b/content/chainguard/libraries/python/build-configuration.md
@@ -67,6 +67,8 @@ accessing your organization's JFrog Artifactory repository manager.
 1. Select **Administration** in the top navigation bar.
 1. Select **Repositories** in the left hand navigation.
 1. Select the **Virtual** tab in the repositories view.
+1. Locate the *python-all** repository row and press the three dots
+   (**...**) in the last column on the right.
 1. Locate the virtual repository set up by your organization to use Chainguard Libraries and select the three dots (`...`) in the last column on the right. (*python-all** is the recommended name for this virtual repository.)
 1. Select **Set Me Up** in the dialog.
 1. Select **Generate Token & Create Instructions**
--- a/content/chainguard/libraries/python/build-configuration.md
+++ b/content/chainguard/libraries/python/build-configuration.md
@@ -67,8 +67,7 @@ accessing your organization's JFrog Artifactory repository manager.
 1. Select **Administration** in the top navigation bar.
 1. Select **Repositories** in the left hand navigation.
 1. Select the **Virtual** tab in the repositories view.
-1. Locate the *python-all** repository row and press the three dots
-   (**...**) in the last column on the right.
+1. Locate the virtual repository set up by your organization to use Chainguard Libraries and select the three dots (`...`) in the last column on the right. (*python-all** is the recommended name for this virtual repository.)
 1. Select **Set Me Up** in the dialog.
 1. Select **Generate Token & Create Instructions**
 1. Copy the generated token value to use as the password for authentication.
--- a/content/chainguard/libraries/python/build-configuration.md
+++ b/content/chainguard/libraries/python/build-configuration.md
@@ -69,7 +69,6 @@ accessing your organization's JFrog Artifactory repository manager.
 1. Select the **Virtual** tab in the repositories view.
 1. Locate the *python-all** repository row and press the three dots
    (**...**) in the last column on the right.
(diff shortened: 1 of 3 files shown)


content/chainguard/libraries/python/build-configuration.md
```diff
@@ -67,8 +67,7 @@ accessing your organization's JFrog Artifactory repository manager.
 1. Select **Administration** in the top navigation bar.
 1. Select **Repositories** in the left hand navigation.
 1. Select the **Virtual** tab in the repositories view.
-1. Locate the *python-all** repository row and press the three dots
-   (**...**) in the last column on the right.
+1. Locate the virtual repository set up by your organization to use Chainguard Libraries and select the three dots (`...`) in the last column on the right. (*python-all** is the recommended name for this virtual repository.)
```
mosabua, line 70: I would invert this .. and talk about python-all first and foremost .. that is what we document in the settings and everywhere else as the happy path. And then you can talk about different virtual repo use in a specific org later here or completely separately.

content/chainguard/libraries/python/global-configuration.md
```diff
@@ -117,13 +117,16 @@ Configure a remote repository for the PyPI public index:
...
 1. Press **Create a Repository** and choose the **Virtual** option.
+1. Select *PyPI* as the Package type.
 1. Set the **Repository Key** to `python-all`.
 1. In the **Repositories** section, find the `python-chainguard` and
    `python-public` repositories. Ensure the `python-chainguard` repository is
    the first in the displayed list. Use the icon on the right of the repository
    name to drag and drop repositories into the desired position.
 1. Select **Create Virtual Repository**.
 
+At this point, you have a virtual repository set up in Artifactory that will allow you or others at your organization to access Chainguard Libraries for Python with your chosen tools. This setup will fall back to the public PyPI index in cases where a package is not available in Chainguard's index
```
mosabua, line 128:
```suggestion
At this point, you have a virtual repository set up in Artifactory that allows you or others in your organization to access Chainguard Libraries for Python with your chosen tools. This setup falls back to the public PyPI index in cases where a package is not available in Chainguard's index.
```

content/chainguard/libraries/python/overview.md
```diff
@@ -108,16 +108,15 @@ apiflask-0.12.0.tar.gz
...
 unique identifiers and cannot be determined without browsing. The list uses
 ascending order for the full name including the version.
 
 Use the search functionality on [pypi.org](https://pypi.org/) to locate packages
 of interest and then browse in the simple index to determine available versions
 in Chainguard Libraries for Python.
 
-Use `curl`, specify the username and password retrieved with [chainctl for basic
-user authentication](/chainguard/libraries/access/) and use the URL of the file
+Use `curl`, specifying the username and password retrieved with [chainctl](/chainguard/libraries/access/), and use the URL of the file
```
mosabua, line 119: wrap please

mosabua (changes requested):

content/chainguard/libraries/python/global-configuration.md
smythp: I think you broadly have an issue with future tense but this is subjunctive / hypothetical. But I'll agree to disagree here and accept the suggestion.

content/chainguard/libraries/python/build-configuration.md
mosabua: No .. I think we should stick with the working example first and foremost on a clean install .. since that helps people who are getting started .. and then a hint for more experienced admin .. anyone running those repo managers in prod knows anyway

content/chainguard/libraries/python/global-configuration.md
mosabua: Yes.. it adds no value and there is no consistent way to choose when to use future tense and when not .. if you have a list of steps .. do you go future of future of future or something. For technical writing its typically considered bad style - https://developers.google.com/style/tense

content/chainguard/libraries/python/build-configuration.md
smythp: Reverting for now.

smythp:
Resolved issues

content/chainguard/libraries/python/build-configuration.md
mosabua: Seems to be still there.. did you forget to push?

mosabua (approved):

PR #2344, opened 2025-06-26
Set-up should be set up

patrickjburke245:
In the first paragraph, “set-up” should be 2 words (set up) without a hyphen. “Set up” is a verb, whereas “set-up” is a noun.


smythp (approved):
Good catches, appreciate the contribution.

SharpRake (approved):
Great catch - thank you @patrickjburke245!

PR #2347, opened 2025-06-27
Bump LB module consumption to remedy TLS issue

mattlorimor:

## Type of change
GCP infra

### What should this PR do?

Bumps the serverless LB consumer to a version that has a TLS policy that enforces at least `TLSv1.2`.

We require all in-transit encryption to meet a minimum standard of `TLSv1.2` or higher.

We should see the TF plan add a new SSL policy to LB.


jamonation (approved):
Should modernize the academy's TLS so +1

ltagliaferri (approved):

PR #2350, opened 2025-06-29
Add llms.txt

ltagliaferri:
Adding an llms.txt file and modifying the Hugo config file. The llms.txt file should render in the root website once generated.


SharpRake (approved):
lgtm!

PR #2353, opened 2025-06-30
Add button for copying a page's Markdown

ltagliaferri:
This change adds a button that allows the user to copy the Markdown of a page for LLM ingestion.

### DIFF
--- a/assets/js/copypage.js
+++ b/assets/js/copypage.js
@@ -55,6 +55,10 @@ async function getPageAsMarkdown() {
     '.notice',
     '.rumble-comparison',
     '.rumble-vuln',
+    '.tooltip-container',
+    '.tooltip',
+    '.top-container',
+    '.expand-btn-container',
     'nav',
     'script',
     'style'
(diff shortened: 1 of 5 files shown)


SharpRake (approved):
One thing I'm a bit wary of with this button is that it leaves some odd artifacts in the copied text.

One global example I found is that it includes the label of a command block within the block.

So for example, in the [Starting up a k3s cluster section](https://deploy-preview-2353--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/getting-started/cilium/#start-up-a-k3s-cluster) cilium getting started guide, this is what gets copied in the first code block:

> To do so, create a YAML manifest named `k3d.yaml` with the following command:
>
>```sh
> sh**Copycat > k3d.yaml <<EOF
> apiVersion: k3d.io/v1alpha5

I've also found a few odd formatting things where it will add a lot of extra spaces around title headings or quote blocks:

```

# Getting Started with the Cilium Chainguard Containers

        Tutorial on the Cilium Chainguard Containers



















      Cilium is open source software for transparently securing the network connectivity between application services deployed using Linux container management platforms like Docker and Kubernetes. At the foundation of Cilium is a new Linux kernel technology called [eBPF](https://ebpf.io/), which enables the dynamic insertion of powerful security visibility and control logic within Linux itself. Because eBPF runs inside the Linux kernel, Cilium security policies can be applied and updated without any changes to the application code or container configuration.
```

If we're okay with this presentation (and I'm not sure if we ought to be, as it's 'for' LLMs but idk what impact these artifacts will have) then take my approval and merge it, otherwise happy to do a deeper dive into what I find.

PR #2359, opened 2025-07-01
update robots.txt for ai crawling

ltagliaferri:


mosabua:
Is that needed and not covered by the generic * pattern .. also why do we not exclude the previews for the bots?

PR #2363, opened 2025-07-01
Add related content for LLMs and front end. Also add time to read

ltagliaferri:
This modifies enhanced-structured-data.html to include related content so that LLMs can have an easier pathway to search related information. While making this change, I added a Related Articles section on the bottom to expose this to human readers on the front end. Additionally, added the read time to these related articles and to the front matter.


matthewhelmke (approved):
LGTM

PR #2382, opened 2025-07-07
Update Footer

ltagliaferri:
This updates the footer to bring it in line with the www site. Please confirm that this looks right. I opted not to add the icons as edu is considered a separate site in the www footer.


SharpRake (approved):
Changes lgtm!

PR #2383, opened 2025-07-07
use-with-openshift.md: use generic image

ribbybibby:
The image used in the example is a custom image that is not generally available.

Switch the example to a generic image. Also use the `-dev` tag because the generic image doesn't include a shell.

I also noticed we got the permissions slightly wrong in one part.


matthewhelmke (approved):
LGTM, thank you!!

PR #2405, opened 2025-07-14
Add warning about matching user statement.

amouat:

Fix to clarify issue with USER statements when adding APK repo.


erikaheidi (approved):
looking good to me!

PR #2410, opened 2025-07-14
fix: chainctl expected build output

klrmngr:


ltagliaferri (approved):
Thanks for fixing the wonkiness here!

PR #2419, opened 2025-07-16
AI-optimized content phase 1

ltagliaferri:
This updates leads and some of the initial writing in a number of key docs to support AI optimization, following recommendation from @amouat


amouat (approved):
These changes are great!

I think "built-in" might be a bit confusing, so I'd just drop it.

PR #2421, opened 2025-07-16
Add dates to precommit

ltagliaferri:
This should add dates and lastmod automatically. Documentation also updated on how to use


mosabua (approved):
Great new feature and docs for using. Do you want someone to test this before merge?

PR #2423, opened 2025-07-18
Dockerfile for Hugo.

amouat:
Adds a very simple Dockerfile for building and running the docs server.

If iterating on code, you will want to mount the code directory as a volume.

Biggest advantage right now is you don't need install npm and old libraries on your host. Going forward it might give us a consistent base for testing changes (e.g. updating hugo version).

We don't need to merge this, but wanted to keep the Dockerfile somewhere at least.

### DIFF
--- a/Dockerfile.hugo
+++ b/Dockerfile.hugo
@@ -1,0 +1,11 @@
+FROM --platform=linux/amd64 cgr.dev/chainguard/node:latest-dev
+
+COPY . .
+USER root
+RUN npm install
+USER node
+ENTRYPOINT ["/bin/npm"]
+CMD ["run", "start"]
+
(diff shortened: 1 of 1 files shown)


kallewesterling:
Love it

kallewesterling (approved):

kallewesterling:
Just thinking, should we add something in the `README.md` for the `Dockerfile.hugo` as well?

amouat:
Ah, another solution is to change the COPY line to only copy over the essential directories (it shouldn't really copy everything)

mosabua:
I think this is great.. in fact I think we should also add a script that runs the site with the docker container ..

PR #2429, opened 2025-07-18
Improve pre-commit hook use

mosabua:
- Add script to run it to avoid having to type a long command 
  all the time for manual checks.
- Change so all changed files are checked, not just staged ones.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -50,9 +50,9 @@ APPROVED_TAGS = {
     "Product", "chainctl"
 }
 
-def get_staged_files():
-    """Get list of staged markdown files."""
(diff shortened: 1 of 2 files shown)


.githooks/pre-commit
```diff
@@ -212,7 +212,7 @@ def check_spelling(filepath):
         return {}, [f"Error checking spelling: {str(e)}"]
 
 def update_dates():
-    """Update date fields in staged markdown files (date for new files, lastmod for modified)"""
+    """Update date fields in changed markdown files (date for new files, lastmod for modified)"""
```
ltagliaferri, line 215: Does updating the dates for changed vs. staged files make a big difference for your use case?

ltagliaferri (approved):

PR #2430, opened 2025-07-18
Add section for videos about the factory

mosabua:
### What should this PR do?

- Add page for videos about the factory
- Fix https://github.com/chainguard-dev/internal/issues/5228

More info for audience .. more media.

Review and confirm order in shorts list.

Check links refer to right video


ltagliaferri (commented):
What do you think of having them be separate pages for each short with the transcript and maybe a little more detail in them? Not sure that anyone will click through the links to see the shorts on YouTube.

mosabua:
So for now as next iteration I added all shorts to one page and did not create and further descriptions and such. I think for initial launch this is good enough. wdyt @ltagliaferri 

Also.. I used Chainguard Factory as tag .. but maybe I should not?

mosabua:
Whoa @ltagliaferri ... massive update.. Thanks for completing this for me.

PR #2440, opened 2025-07-23
CodeQL suggested edits

kallewesterling:
See https://github.com/chainguard-dev/courses/pull/37 for the errors that CodeQL discovered when I tried running the latest edits to `chainlink`. This should fix all that. I've tested the script, and it should be running fine.

### What should this PR do?
Resolves the problems identified by CodeQL that chainlink has "Log entries created from user input" which are classified as a high priority item.

We don't want our internal tooling to cause any issues like this.

Test should run without problem

After installing go and cloning this repo, switching into the correct branch:
1. `cd tools/chainlink`
2. `go run . -contentDir ../../content`
3. Output should contain lines that state "redirect: ..."

### DIFF
--- a/tools/chainlink/links.go
+++ b/tools/chainlink/links.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"strings"
 )
 
 func (l *link) processLink(path string) {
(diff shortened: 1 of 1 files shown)


SharpRake (approved):
Tested locally, changes work just fine for me!

PR #2445, opened 2025-07-25
Fixing 404s: July 25, 2025

SharpRake:
[ x ] Check if this is a typo or other quick fix and ignore the rest :)

### DIFF
--- a/content/chainguard/libraries/python/build-configuration.md
+++ b/content/chainguard/libraries/python/build-configuration.md
@@ -161,7 +161,7 @@ Note the different syntax for `index-url` in the two files.
 
 Refer to the official documentation for [configuring authentication with
 pip](https://pip.pypa.io/en/stable/topics/authentication/) if you are not using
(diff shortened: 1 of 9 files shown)


content/chainguard/libraries/python/build-configuration.md
```diff
@@ -161,7 +161,7 @@ Note the different syntax for `index-url` in the two files.
 
 Refer to the official documentation for [configuring authentication with
 pip](https://pip.pypa.io/en/stable/topics/authentication/) if you are not using
-[.netrc for authentication](/chainguard/libraries/access#netrc).
+[.netrc for authentication](/chainguard/libraries/access/#netrc-for-authentication).
```
mosabua, line 164: this is the wrong anchor .. there is an explicit anchor inserted in the page that wont change even if we change title ..

SharpRake: thanks for letting me know, fixed these here: https://github.com/chainguard-dev/edu/pull/2447

PR #2446, opened 2025-07-25
Tweak learning labs pages

mosabua:
## Type of change

Docs improvement

### What should this PR do?

- Add missing link to index page
- Remove use of "we"
- Remove link to event pages since labs notes pages replaces them

Follow up to prior PR.

Review

n.a.

### DIFF
--- a/content/software-security/learning-labs/ll202507.md
+++ b/content/software-security/learning-labs/ll202507.md
@@ -41,7 +41,7 @@ The July 2025 Learning Lab with Patrick Smyth covers AI with Hardened Container 
 
 ## Demo
 
-In the demo, we train and run inference on an image classification model using PyTorch and Chainguard's hardened container image. The model classifies images of octopuses, whales, and penguins, demonstrating how to work with AI workloads securely.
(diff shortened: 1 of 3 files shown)


content/software-security/learning-labs/ll202507.md
```diff
@@ -91,11 +91,9 @@ The demo showcases how Chainguard's hardened PyTorch image provides the same fun
 
 ## Resources
 
-- [Event page](https://goldcast.ondemand.goldcast.io/on-demand/3ec17730-bff5-40b3-b5d9-23c90ac26828)
```
ltagliaferri, line 94: I would keep the event page

ltagliaferri (approved):
I would keep the event page

content/software-security/learning-labs/ll202507.md
mosabua: It has no relevant content that is not on this page and actually will probably be deleted ... so I think we should not have it here .. if we really want .. it should be the last link .. not the first one

mosabua: it only makes sense to have it when it is the registration page before the event and if we publish the learning labs page before the evnet .. which happened last time .. but now it makes no sense

PR #2463, opened 2025-07-31
Fix manual wheel download example

mosabua:

Tested it manually for two files.. should be good to go

Fixes https://github.com/chainguard-dev/internal/issues/5260


wiedenmeier (approved):
Awesome, thanks

ltagliaferri (approved):

PR #2470, opened 2025-08-04
Adding doc on using init containers

SharpRake:
## Type of change
  
**New Resource**
This PR adds a new how-to guide to the documentation site titled **"Using Init Containers with Chainguard Containers"**. The guide explains how to reconfigure Chainguard’s minimal nginx container image using init containers within Kubernetes deployments.

### What should this PR do?
  
Resolves https://github.com/chainguard-dev/internal/issues/5027

Customers migrating from the default Docker Hub nginx image to the minimal Chainguard nginx image often rely on startup configuration mechanisms such as `docker-entrypoint.d`, which are not present in the minimal Chainguard images. This new doc demonstrates a secure and modular alternative using Kubernetes init containers to handle configuration logic without modifying the core Chainguard image, helping users avoid common security pitfalls while maintaining functionality.

* [x] New documentation appears under the "How to Use" section
* [ ] Content is technically accurate
* [ ] Code snippets and Kubernetes YAML are properly formatted and syntactically correct
* [ ] Example workflow can be reproduced by following the steps
* [ ] Stakeholders in developer experience and containers teams are tagged for review

Just follow the doc as it's written and give everything a close read. It should all work as written!

Note that testing this will require docker, a K8s cluster, and push access to a container registry.

[PREVIEW LINK](https://deploy-preview-2470--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/how-to-use/init-containers/)

### DIFF
--- a/content/chainguard/chainguard-images/how-to-use/init-containers.md
+++ b/content/chainguard/chainguard-images/how-to-use/init-containers.md
@@ -16,19 +16,18 @@ weight: 035
 toc: true
 ---
 
+Chainguard Containers are designed with minimalism and security in mind. By including fewer packages and tools, Chainguard Containers have a smaller attack surface than their counterparts. However, there are cases where the external counterparts have certain desirable features, like useful startup scripts or configuration defaults.
 
-There are a number of ways you can customize Chainguard Containers. For example, you can use [Custom Assembly](/chainguard/chainguard-images/features/ca-docs/custom-assembly/) to add packages to an otherwise minimal Chainguard container image.
+There are several ways to customize Chainguard Containers. For example, you can use [Custom Assembly](/chainguard/chainguard-images/features/ca-docs/custom-assembly/) to add packages to an otherwise minimal Chainguard container image. Changing a Chainguard container image's configuration — such as updating its entrypoint or adding startup scripts — requires a different strategy. One method for doing so in Kubernetes deployments is to use *init containers*. 
 
-Changing the container image's configuration — such as updating its entrypoint or adding startup scripts — requires a different strategy. One method for doing so in Kubernetes deployments is to use *init containers*. 
-
-This guide provides a brief overview of what init containers are as well as their benefits. It then outlines an example of how you can use an init container to reconfigure Chainguard's nginx container image.
+This guide provides a brief overview of what init containers are as well as their benefits. It then outlines an example of how you can build an init container and use it to reconfigure Chainguard's nginx container image.
 
 ## Prerequisites
 
 In order to follow this guide, you will need the following:
 
 * Docker, which you will use to create an image to be used by an init container, installed on your local machine. Refer to the [Docker documentation](https://docs.docker.com/engine/install/) to set this up.
-* A container registry you can push the init container image to. There are many options for this, but [Docker Hub](https://hub.docker.com/) is free to use and has found wide adoption.
+* A container registry you can push the init container image to. This guide assumes you have access to a [Docker Hub registry](https://hub.docker.com/), though the example can be adjusted for other registries.
 * Access to a Kubernetes cluster. This could be a cluster that you run locally with a tool like [minikube](https://minikube.sigs.k8s.io/docs/) or [kind](https://kind.sigs.k8s.io/), or a cloud-hosted cluster such as [Amazon EKS](https://aws.amazon.com/eks/) or [Google GKE](https://cloud.google.com/kubernetes-engine).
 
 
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-images/how-to-use/init-containers.md
```diff
@@ -0,0 +1,252 @@
...
+       - name: nginx
+         image: cgr.dev/chainguard/nginx:latest
+         ports:
+           - containerPort: 80
+         volumeMounts:
+           - name: nginx-config
+             mountPath: /etc/nginx
+EOF
+```
+
```
erikaheidi, line 195: Something odd is happening when I paste this command, the $REGISTRY and $NAMESPACE variables are removed somehow and I end up with this:

```
initContainers:
       - name: run-entrypoint
         image: //nginx-init
         volumeMounts:
           - name: nginx-config
             mountPath: /etc/nginx
```

I know you are supposed to replace them, but the placeholder helps identify the place to edit. The issue seems to be with the `$`, not sure how that works in the EOF block, but since these aren't real variables it would be better to remove the `$` also to avoid confusion

content/chainguard/chainguard-images/how-to-use/init-containers.md
```diff
@@ -0,0 +1,252 @@
...
+Again, be sure to change `$REGISTRY/$NAMESPACE/nginx-init` as necessary to reflect your own setup.
+
+### Testing the example deployment
+
+Using the manifest you just created, create a Kubernetes deployment:
+
+```shell
+kubectl create -f nginx-deployment-cg.yaml
```
erikaheidi, line 203:
```suggestion
kubectl create -f init-deployment.yaml
```

erikaheidi (dismissed):
I tested all commands and everything worked great, I just had to fix a small issue with variables and a mismatched file name. The procedure was easy and it works super well, loved to learn more about it.

What I did struggle a bit was understanding how it all worked at first, I wasn't expecting we would be using an external image in the process so I felt confused. After re-reading the introduction I understood it as a process similar to multi stage builds, which makes total sense!

content/chainguard/chainguard-images/how-to-use/init-containers.md
SharpRake: :grimacing: good catch!

content/chainguard/chainguard-images/how-to-use/init-containers.md
```diff
@@ -0,0 +1,252 @@
...
+* **Unintended Script Execution**: If you accidentally include a shell script (or malicious one gets included through CI/CD or shared build layers), it will be executed, regardless of its purpose or trustworthiness.
+* **Privilege Escalation**: If the container runs as root (which is common in base images), any `*.sh` script executed at startup runs with full privileges. Malicious scripts can add users, exfiltrate secrets, overwrite configs, etc.
+
+With init containers, you could avoid these issues by decoupling the nginx `docker-entrypoint.sh` entrypoint script from the core image running nginx. This would involve running the startup configuration in the init container, then using a shared volume to mount these configurations at runtime. This avoids introducing the aforementioned security risks into the minimal Chainguard nginx container image.
+
+This section illustrates how to set up an nginx init container inside a Kubernetes deployment. This example involves creating a sample `nginx.conf` file, building an init container with `docker`, and deploying it in a Kubernetes pod.
+
```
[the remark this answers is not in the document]
SharpRake: Thanks for this! It was definitely lacking in that aspect. I reworked this section to more clearly explain the motivation here.

content/chainguard/chainguard-images/how-to-use/init-containers.md
```diff
@@ -0,0 +1,252 @@
...
+CMD cp /custom/nginx.conf /etc/nginx/nginx.conf && \
+   chmod 644 /etc/nginx/* && \
+   /docker-entrypoint.sh true
+EOF
+```
+
+This Dockerfile will build an image using the default nginx container image from Docker Hub. This image includes `docker-entrypoint.d` and `docker-entrpointy.sh` by default. It will copy the `nginx.conf` file you created previously into the new image when you build it. Then, when you run the container in a Kubernetes deployment, it will prepare and copy the nginx configuration at startup using the entrypoint script.
+
+After creating this Dockerfile, build the image. Note that in both this command and the following one, you will need to change the `$REGISTRY` and `$NAMESPACE` variables to reflect your own container registry and namespace. If you plan to use Docker Hub as your container registry, you can omit the `$REGISTRY` portion of the command:
```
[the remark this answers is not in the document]
SharpRake: You're right - to simplify things I reworked this example to explicitly tell users to create an env variable named `$NAMESPACE`, and reworked the prerequisites to explictly state this guide assumes you're using a Docker Hub registry, but that it can be extended  for other registries. I tested the commands with the new structure and it worked on my machine :tm: .

ltagliaferri (approved):

PR #2507, opened 2025-08-13
Documentation for the Feedback Widget

SharpRake:

## Type of change

New platform doc outlining the functionality of the feedback widget.

I added the feedback widget last week without including detailed info on how it works. This is meant to provide a general overview without giving away too many details.

Doc should serve as a starting point for folks looking to understand the feedback widget.

Just a read, this isn't going on Academy so it doesn't need to adhere to the style guide.


smythp (approved):
Looks good :+1:

PR #2545, opened 2025-08-28
[WIP] Implementing more specific code block logic

SharpRake:


mosabua:
How about implementing this kinda thing as a precommit hook instead @SharpRake and @ltagliaferri

PR #2550, opened 2025-09-02
Update direct access Python docs for pip/uv

angela-zhang:
Updating docs based on customer feedback to recommend uv when wanting to set PyPI as a fallback index repository.


ltagliaferri (approved):
Just a typo to fix, thank you!

mosabua (changes requested):
Well.. we are currently missing any documentation on how to use PyPI as a fallback completely so we should add that as well before we talk about the exception. Also how to do that with uv should probably be documented in the uv section. And similarly we should add to the pip and poetry sections.

angela-zhang:
@mosabua we have documentation on how to set PyPI as a fallback for artifact repositories, correct? `pip` does not provide a method to set a priority repository, but `uv` does. I see poetry also had a way to set priority repositories: https://python-poetry.org/docs/repositories/#project-configuration

I can edit this to add instructions to the uv and poetry on how to ensure the Chainguard Python repository is the top priority repository/index URL instead. From our investigation, there isn't a way to do this with pip though.

angela-zhang:
@manojkrgupta : I updated for pip, poetry, and uv sections. Linked out to uv and poetry docs that describe how to set priority indexes.

content/chainguard/libraries/python/build-configuration.md
```diff
@@ -286,6 +293,11 @@ poetry source add chainguard https://libraries.cgr.dev/python/simple/
 The [Poetry documentation](https://python-poetry.org/docs/) contains more
 information about your project build, dependencies, versions, and other aspects. 
 
+If you want to install Python libraries from multiple indexes with a priority 
+index resolution order, `poetry` supports setting a [primary package source](https://python-poetry.org/docs/repositories/#project-configuration). 
+For example, some organizations may choose to prioritize the Chainguard Python
```
mosabua, line 298: I think we should be stronger and recommend that .. probably use the same wording as in the repo manager section. For many use cases it is essentially required

content/chainguard/libraries/python/build-configuration.md
```diff
@@ -193,6 +193,13 @@ Example for `~/.pip/pip.conf`:
 index-url = https://CG_PULLTOKEN_USERNAME:CG_PULLTOKEN_PASSWORD@libraries.cgr.dev/python/simple/
 ```
 
+Note that `pip` does not support installing Python libraries from multiple 
+indexes while prioritizing one index over another. If you are using `pip` 
+and prefer to pull from multiple indexes while prioritizing the Chainguard
```
mosabua, line 198: I think we should stay consistent in the naming - Chainguard Libraries for Python is used everywhere else in the docs

mosabua (approved):

mosabua:
Looks great now.. lets ship it.

PR #2591, opened 2025-09-11
Adding Ko guide in new Tooling section

erikaheidi:
## Type of change
This PR adds a new section in Chainguard Academy dedicated to container building tools, or "last mile tools" as we initially called them. To avoid confusion with future tools we may (or may not) build, we'll keep it simple (also respecting menu sizing) and just call it "Tooling" for the menu. It also adds the first tutorial in this section, using Ko to build Go containers with Chainguard base images.

### What should this PR do?
This work is described [here](hhttps://github.com/chainguard-dev/internal/issues/5198). 

This is part of an initiative called "last mile container tools" to bring resources on how to use tools other than Docker to build container images.

- Tech test passes
- Content follows our technical writing guidelines
- Menu and titles look good and make sense under our current structure

- Live preview: https://deploy-preview-2591--ornate-narwhal-088216.netlify.app/chainguard/chainguard-images/tooling/building-go-containers-with-ko/

Disclaimer: I was not able to build the site locally, currently getting several errors from the Hugo build / serve, so I am depending on the live preview to see how it will look when published.

### DIFF
--- a/content/chainguard/chainguard-images/tooling/_index.md
+++ b/content/chainguard/chainguard-images/tooling/_index.md
@@ -1,0 +1,11 @@
+---
+title: "Container Tooling"
+linktitle: "Tooling"
+description: "Complimentary resources on how to use Chainguard Containers with alternative container building tools such as Ko and Jib"
+type: "article"
+date: 2025-09-11T08:49:15+00:00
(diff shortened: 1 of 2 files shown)


amouat (approved):
Looks good, only minor points.

I would add something on how to build multi-arch or for a given arch.

PR #2596, opened 2025-09-11
Add docs for library pull token creation in console

mosabua:
## Type of change

Adding docs for libraries

### What should this PR do?

Add info on how to create and modify a pull token in the console so it can be used with libraries.

Wrongly thought that this is not possible and never heard from anyone yet to the contrary.

Fix https://github.com/chainguard-dev/internal/issues/5365

Precommit hooks passed.

Pass review.

Follow documented steps

### DIFF
--- a/content/chainguard/libraries/access.md
+++ b/content/chainguard/libraries/access.md
@@ -14,7 +14,9 @@ weight: 002
 toc: true
 ---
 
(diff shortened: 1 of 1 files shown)


SharpRake (approved):
Changes look good! Just a few non-blocking suggestions

SharpRake (approved):
just one minor suggestion, don't let it block you though

PR #2609, opened 2025-09-22
Clean codeblock languages

kallewesterling:

### What should this PR do?
- Adds `hljs-audit`, a CLI tool to validate fenced code blocks in Markdown against our Highlight.js configuration.
- Updates existing documentation to fix unlabeled or incorrectly labeled code blocks.
- Fixes some inconsistent padding etc. in JSON and YAML code blocks

- Prevent drift between our documentation and frontend rendering (Highlight.js).
- Make sure all code samples display with proper syntax highlighting.
- Lay groundwork for CI enforcement of code block quality.

- tool is present & documented
- docs are clean
- correct behavior

1. install pyyaml
  ```py
  pip install pyyaml
  ```
2. check help text
  ```sh
  ./hljs_audit.py -h
  # Expect usage text; exit code 0
  ```
3. audit repo
  ```sh
  ./hljs_audit.py content -c config.yaml --strict
  echo $?   # Expect 1
  ```

  Expected output, something like:

  ```output
  hljs-audit report
  ================

  Disallowed language blocks: ...
  ```


SharpRake (changes requested):
So this is tough, because a lot of your changes (like labeling unlabeled code blocks that are clearly a command or output) make perfect sense and are helpful, but a lot (maybe even more) are questionable and I don't think we should merge them:

* removing capitalization (`Dockerfile` to `dockerfile`, for example) is unnecessary.
* adding the `sh` or `bash` labels is unnecessary, as we want these to ultimately be `shell` anyway
* there are many cases where a useful label (like a file name or `EC2`, for example) is changed to something less helpful
* there are cases where something gets the wrong label entirely
* all the changes made to docs under `chainctl-docs` will disappear the next time those docs get updated

Lastly, I'm not totally clear on what exactly the `hjs-audit.py` script does but from what I can tell it doesn't give us more flexibility with our code blocks, and actually seems to restrict what we can do.

Overall I think we need a *_much_* lighter touch here. Happy to chat whenever you want

PR #2618, opened 2025-09-23
Small note for Grype scanning

angela-zhang:
Adding a note on caveat for Grype scanning based on internal testing feedback.


mosabua (changes requested):

mosabua (approved):
Looks good now.

PR #2634, opened 2025-09-29
Google csp

jonverson:

## Type of change
Updated the CSP settings. Currently the edu site is failing to load many references to connect-src, script-src, img-src and frame-src. I added all the scripts that were failing on the edu site.

### What should this PR do?
This PR should resolve current broken CSP console errors.

To allow tag manger to track user visits to edu site and allow scripts on the site to work correctly.

The CSP console errors should be cleared up when loading the edu site

You can pull this branch down locally test if there are any CSP errors in the console. If there is a staging site this should also work there as well.


ltagliaferri (commented):
It looks like there is more than just Google Tag Manager links being added here... Would you be able to walk me through them and what they are used for? How can we access the data from these?

We removed the 6sense one based on the understanding that we were sunsetting that.

jonverson:
GTM is just the vehicle for loading all the scripts. The scripts that I
added are all ones that chrome was blocking and complaining about. If these
scripts aren't white listed the edu site will block them and they won't
load.

PR #2644, opened 2025-10-02
Google csp

jonverson:

## Type of change
Added js.hs-banner.com script to the allowed CSP list under connect-src

### What should this PR do?
This PR should fix this script fromm being blocked.

To fix the hubspot cookie consent banner from being blocked.

The js.hs-banner.com script should no longer be blocked in the network tab.

Check the network tab to make sure that the js.hs-banner.com script is no longer blocked.


ltagliaferri (approved):
Please sign the commit next time 🙂

PR #2646, opened 2025-10-02
Update chainver docs

smythp:
## Description

Updates chainver docs to reflect more of the information in the README.

- Added usage section
- Updated binary command
- Added go installation and docker instructions
- Added links for auth, getting tokens
- Added resource section
- Full cleanup and review

## Notes

Resolves https://github.com/chainguard-dev/internal/issues/5411

### DIFF
--- a/content/chainguard/libraries/verification.md
+++ b/content/chainguard/libraries/verification.md
@@ -2,7 +2,7 @@
 title: "Chainguard Libraries Verification"
 linktitle: "Verification"
 description:
-  "Learn how to verify Java and Python dependencies are from Chainguard
+  "Learn how to verify libraries and packages are from Chainguard
   Libraries using the chainver tool for enhanced supply chain security"
 type: "article"
 date: 2025-07-03T12:00:00+00:00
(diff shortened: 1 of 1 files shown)


mosabua:
Also .. for the scripting and in general we should make sure it works on linux and macos .. and maybe even windows (maybe using WSL?) ... thoughts @dakaneye ?

smythp:
I included the direct download and a section using the metadata endpoint in install (from the PR).

I adapted the binary install a little but hoping you can test on macOS. I made it clear that part is for macOS/Linux but it might also work on WSL (I put in a little adaptation for tar/zip in case). I don't have access to a WSL machine right now, however. For non-WSL the grepping and such is an issue for that command, but I think that's as far as we can/should go for Windows, if a Powershell-wielding cusotmer comes at us I'll add a section.

I don't think there's a rush on this, can wait until next week.

ltagliaferri (approved):
Some nits

mosabua (changes requested):
Amazing effort .. I think you are essentially doing the whole scope of expanding and moving the chainver docs onto the site. Love it.

I provided a whole bunch of feedback from smaller nits to other issues. Ping me if anything seems insane ;-)

mosabua (approved):
As discussed in our call this is good enough to ship now. We will follow up with further tweaks and clarifications for the token stuff and other aspects as they come up and we are able to send PRs.

PR #2650, opened 2025-10-03
Add further documentation on Disable URL Normalization in index.md

mental-lab:
Add Disable URL normalization in docs under under Pull Through for Public repos and Debugging section for more visibility and instruction

### What should this PR do?
PR clarifies the cause of the package mentioned in index not found error when using Artifactory as a pull-through cache for Chainguard APK repos

Users and myself just found out about this and have been confused. That confusion leads to misconfigured remote repos and failed apk downloads. I have included this setting under Pull Through Cache for Public Repos and added to Debugging to prevent further problems associated with this setting.


SharpRake (approved):
I made a few further tweaks, thanks for making this change @mental-lab!

PR #2654, opened 2025-10-06
Update cve-remediation.md

patrickjburke245:
Feedback:
Removed period in description to match the format of other docs.
Changed "upstream" to "newer" since it's an overloaded term in this doc.
"especially when" is something I'd be more apt to say out loud than "since".
Since would cover all reasons, whereas especially explains this is only the most common case (but not the only case).
Decreased word count without decreasing comprehensibility.
"introduce" seems more fitting.


mosabua (approved):

mosabua:
Thanks @patrickjburke245

PR #2712, opened 2025-10-24
Improve Python CVE version info

mosabua:
Discussed this change with @mosabua over Slack

Replaces https://github.com/chainguard-dev/edu/pull/2709


SharpRake (approved):
1 minor suggestion

PR #2716, opened 2025-10-27
replace ai bundle icon

ltagliaferri:

## Type of change
This replaces the icon for the AI Bundle link after feedback from @garrying


mosabua (approved):
Preview looks good to me. Not sure if there is a more standard icon so I think we should go ahead with this improvement

garrying (approved):
lgtm

PR #2738, opened 2025-11-05
Update overview.md

patrickjburke245:

### DIFF
--- a/content/chainguard/libraries/python/overview.md
+++ b/content/chainguard/libraries/python/overview.md
@@ -54,7 +54,7 @@ https://libraries.cgr.dev/python/
 https://libraries.cgr.dev/python-remediated/
 ```
 
-The first index provides all Python libraries that Chainguard builds from
+To acess these pages, navigate to [Manual Access](https://edu.chainguard.dev/chainguard/libraries/python/overview/#manual-access). The first index provides all Python libraries that Chainguard builds from
 source, without remediated versions. The second index provides remediated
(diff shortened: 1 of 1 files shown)


mosabua (changes requested):
I have to think about this more.. the mention for manual access should definitely NOT be the first thing here since this is not a normal use case for developers at all. Configuring a repo manager and build tools are .. 

In addition whatever we change here in Python should also be consistently applied in the Java side

content/chainguard/libraries/python/overview.md
```diff
@@ -54,7 +54,7 @@ https://libraries.cgr.dev/python/
 https://libraries.cgr.dev/python-remediated/
 ```
 
-The first index provides all Python libraries that Chainguard builds from
+To acess these pages, navigate to [Manual Access](https://edu.chainguard.dev/chainguard/libraries/python/overview/#manual-access). The first index provides all Python libraries that Chainguard builds from
```
ltagliaferri:
```suggestion
To access these pages, navigate to [Manual Access](https://edu.chainguard.dev/chainguard/libraries/python/overview/#manual-access). The first index provides all Python libraries that Chainguard builds from
```

ltagliaferri (commented):
@mosabua any further thoughts on this update?

mosabua:
@ltagliaferri .. I will have to review the Java side in parallel and adjust ..  let me run with this

PR #2742, opened 2025-11-06
Minor fix on curl command

shucking:
the endpoint url needs to be in single quotes


SharpRake (approved):
Nice catch! thanks @shucking

PR #2760, opened 2025-11-12
Improve manual browsing info for Python and Java

mosabua:
Replaces https://github.com/chainguard-dev/edu/pull/2738 and expands to cover Java with more detail

### DIFF
--- a/content/chainguard/libraries/python/overview.md
+++ b/content/chainguard/libraries/python/overview.md
@@ -77,6 +77,10 @@ Alternatively, you can use the token for direct access to the Chainguard
 Libraries for Python index as discussed in [Build
 configuration](/chainguard/libraries/python/build-configuration/).
 
+Follow the steps detailed in [Manual Access](#manual) to browse the Python index
+and find available packages, package versions, source distribution (sdist), and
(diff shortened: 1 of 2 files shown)


content/chainguard/libraries/python/overview.md
```diff
@@ -77,6 +77,11 @@ Alternatively, you can use the token for direct access to the Chainguard
 Libraries for Python index as discussed in [Build
 configuration](/chainguard/libraries/python/build-configuration/).
 
+Follow the steps detailed in [Manual Access](#manual) to browse the
+simple Python index and find available packages, package versions, source
```
ltagliaferri: What is the meaning of "simple" here?

ltagliaferri: If it's just a qualitative thing, I would remove it

ltagliaferri (approved):

content/chainguard/libraries/python/overview.md
mosabua: Its the technical term of the Python index that is browsable .. I can remove here though

PR #2790, opened 2025-11-24
perf: Query BigQuery once per day not 8000 times

jml:
BigQuery does a full table scan unless we select partitions. We're currently querying the `rumble_vulns` view once per vulnerability, which is quadratically expensive and costing us over $500 per day.

This changes the job to do one query, and moves the grouping logic into Go.

Since everything is already entirely in memory already, I don't expect this to cause OOMs.

We still query the local SQLite table once per vulnerability, since it works, doesn't show up on our cost bill, and I want to keep this fairly contained.

```sql
with 

base as (select * from `region-us.INFORMATION_SCHEMA.JOBS` where date(creation_time) = '2025-11-23')

select
  substr(query, 40),
  count(*) as num_queries,
  sum(total_bytes_billed) as total_bytes_billed,
  6 * sum(total_bytes_billed) / (pow(1024, 4)) as cost
from base
where user_email = 'github-chainguard-academy@chainguard-academy.iam.gserviceaccount.com'
group by 1
order by 4 desc
```

| query | num_queries |total_bytes_billed | cost |
| --- | --: | --: | --: |
| SELECT scan.image, scan.time as time | 7459 |97762042445824 | 533.4843578338623 |
| WITH ruuuumble AS (	SELECT scan.image, | 36 | 704013926400 | 3.8417816162109375 |
| SELECT DISTINCT vulnerability FROM `clo" | 1 | 13106151424 | 0.071519851684570312 |
| SELECT ROW_NUMBER() OVER (ORDER BY tim | 1 | 1669332992 | 0.0091094970703125 |

## Type of change

Performance optimization.

### What should this PR do?

Reduce the number of identical BigQuery queries.

To reduce BigQuery costs.

After we merge the PR, we should trigger a manual run of the Rumble Vulnerability Data workflow.

- [ ] Does it succeed?
- [ ] Does it generate plausibly correct data? (Someone other than me will have to check this)
- [ ] Does it result in fewer queries and fewer total bytes billed? (I'll check the BigQuery logs)

### DIFF
--- a/tools/rumble/cmd/vulns.go
+++ b/tools/rumble/cmd/vulns.go
@@ -86,7 +86,7 @@ func (v *vulnsJson) generateJSON() error {
 	}
 
 	// Extract unique CVE IDs for enrichment
-	var cveRecords []interface{}
+	cveRecords := make([]interface{}, 0, len(vulnImageMap))
 	for vulnID := range vulnImageMap {
 		cveRecords = append(cveRecords, &grype.Cve{Vulnerability: vulnID})
 	}
--- a/tools/rumble/cmd/vulns.go
+++ b/tools/rumble/cmd/vulns.go
@@ -14,12 +14,10 @@ import (
 	"os"
 	"strings"
 
-	"cloud.google.com/go/bigquery"
 	cgbigquery "github.com/chainguard-dev/edu/tools/rumble/pkg/bigquery"
 	"github.com/chainguard-dev/edu/tools/rumble/pkg/grype"
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
-	"google.golang.org/api/iterator"
 )
 
 type vulnsJson struct {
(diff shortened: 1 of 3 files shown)


SharpRake (approved):
I'm approving this but would like another review from @cmdpdx (or other SME) before merging

tools/rumble/cmd/vulns.go
```diff
@@ -73,20 +71,54 @@ func (v *vulnsJson) generateJSON() error {
...
-	v.vulns, err = v.grypeDb.ExtractCVEs(allVulnRecords)
+	// Group results by vulnerability ID
+	vulnImageMap := make(map[string][]cgbigquery.VulnWithImages)
+	for _, record := range allRecords {
+		vwi := record.(*cgbigquery.VulnWithImages)
+		vulnImageMap[vwi.Vulnerability] = append(vulnImageMap[vwi.Vulnerability], *vwi)
+	}
+
+	// Extract unique CVE IDs for enrichment
+	var cveRecords []interface{}
```
cmdpdx: We know the length of this ahead of time, probably should pre-allocate so we don't need to allocate on the fly repeatedly
```suggestion
	cveRecords := make([]interface{}, 0, len(vulnImageMap))
```

PR #2795, opened 2025-11-24
Add AWS RDS Postgres 16.11-R1

xnox:


SharpRake (approved):
Made some language tweaks, approving this but would like a review from @max-allan-cgr before merging

PR #2807, opened 2025-11-28
github-identity: Add chainctl example

ribbybibby:
I've received a lot of feedback from customers and colleagues that this page is hard to follow because it exclusively uses Terraform for its examples.

This change adds an example of using `chainctl` to create the identity.

I've also modified the Terraform example to be less procedural. I think stepping through all the files and operations just obfuscates the information that most people care about: what resources do I need to configure?

I think its safe to assume that if someone is looking for a Terraform example, then they are comfortable with defining a provider, setting up a Terraform code base and applying the configuration. Using a snippet communicates the relevant information in a short and digestable way.

### DIFF
--- a/content/chainguard/administration/assumable-ids/identity-examples/github-identity/index.md
+++ b/content/chainguard/administration/assumable-ids/identity-examples/github-identity/index.md
@@ -7,20 +7,23 @@ aliases:
 - /chainguard/administration/iam-organizations/identity-examples/github-identity/
 - /chainguard/administration/assumable-ids/identity-examples/github-identity/
 lead: ""
-description: "Procedural tutorial outlining how to create a Chainguard Enforce identity that can be assumed by a GitHub Actions workflow."
+description: "Tutorial outlining how to create a Chainguard identity that can be assumed by a GitHub Actions workflow."
 type: "article"
 date: 2023-05-04T08:48:45+00:00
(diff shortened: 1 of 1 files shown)


content/chainguard/administration/assumable-ids/identity-examples/github-identity/index.md
```diff
@@ -10,7 +10,7 @@ lead: ""
 description: "Procedural tutorial outlining how to create a Chainguard Enforce identity that can be assumed by a GitHub Actions workflow."
```
amouat: We took out "procedural" below, so I'd take it out here as well. Also remove "enforce"

amouat:
I haven't retested this or checked the screenshots. I note the screenshots are in dark mode which isn't ideal IMO.

amouat (approved):
Minor suggestions.

ribbybibby:
@amouat I think @SharpRake mentioned that we are trending towards removing screenshots in general because of their tendency to go out of date. So I'd be in favour of removing them completely.

To be honest, I think I'll remove the bits about how to create a workflow from the UI as well. Anyone reading this guide will know how to create a GitHub workflow. So I think it's fine to just share the YAML example.

amouat:
Yeah, I think that makes sense.

PR #2808, opened 2025-11-28
versions: add a section that describes how to retrieve version metadata

ribbybibby:
Customers often want to know how they can identify the images and tags that are currently maintained in their organization.

This change adds a section to the Product Release Lifecycle page that describes some methods for retrieving lifecycle information.


SharpRake (approved):
I made a few small changes, thanks @ribbybibby

PR #2811, opened 2025-11-29
assumable-ids: Add generic examples of assuming an identity

ribbybibby:
We provide generic examples of creating an identity, but no tangible examples of assuming one.

This change adds generic instructions on how to assume an identity.

Most importantly, it adds an example of doing the exchange entirely via the API. This comes up a lot with customers who are writing their own implementations with our APIs and don't want to shell out to `chainctl` for authentication.


SharpRake (approved):
Great addition! I made some tweaks and I'm approving so as not to block you, but I did have one question about a section that I think could be deleted. Once you accept my suggestion or reject it, you can go ahead and merge this. Thanks @ribbybibby!

PR #2819, opened 2025-12-08
Update macOS chip reference to M-series

patrickjburke245:
There are now 5 types of M-series chips from Apple. I changed the language to be general enough to reference all 5, rather than just 2.


imsurajsk (approved):

SharpRake (approved):
lgtm!

PR #2826, opened 2025-12-10
python: document python keyring provider

imjasonh:

## Type of change

Add documentation for Python keyring provider

### What should this PR do?

Allow readers to install and use the keyring provider, in environments where that is possible.

We should recommend the use of short-lived credentials wherever possible, and this new release makes it easier for Python users to use these tokens.

- The recommendation of short-lived credentials should be clear and motivated
- Installation and configuration instructions should be clear

By running the documented commands to pull a private Python package using the keyring provider.


content/chainguard/libraries/access.md
```diff
@@ -46,10 +46,48 @@ Successfully exchanged token.
 Valid! Id: 8a4141a........7d9904d98c
 ```
 
+<a id name="python-keyring"></a>
+
+## Python Keyring integration
```
mosabua:
```suggestion
## Python keyring integration
```

mosabua: Do we really want to call it "Python" in the title or just Keyring integration .. 

and then clarify in the context that this is for Python only .. and later expand to others as we have that

content/chainguard/libraries/access.md
```diff
@@ -46,10 +46,48 @@ Successfully exchanged token.
...
+
+Where possible, Chainguard recommends using short-lived credentials to access Chainguard Libraries.
```
mosabua: Add some intro before the recommendation..

content/chainguard/libraries/access.md
```diff
@@ -46,10 +46,48 @@ Successfully exchanged token.
...
+
+Then, when you request to install packages from Chainguard Libraries for Python, the keyring automatically retrieves short-lived credentials for you, using `chainctl`.
+
+To use the keyring with `uv`, install the keyring:
+
+```shell
+uv pip install keyring keyrings-chainguard-libraries
+```
+
+Then to use the keyring you
```
mosabua: ??? complete sentence

mosabua (commented):
I scheduled to chat tomorrow ..

content/chainguard/libraries/access.md
```diff
@@ -46,6 +46,38 @@ Successfully exchanged token.
 Valid! Id: 8a4141a........7d9904d98c
 ```
 
+<a id name="python-keyring"></a>
+
+## Python keyring integration
+
+Python users can leverage the [Chainguard keyring implementation](https://github.com/chainguard-dev/keyrings-chainguard-libraries) to provide short-lived credentials from supported environments, such as local development and CI/CD platforms that support [assumable identities](https://edu.chainguard.dev/chainguard/administration/assumable-ids/assumable-ids/).
```
mosabua: we should pick one word .. either keyring integration or implementation or support or whatever .. and then be consistent

mosabua (changes requested):
I think we have to explain this all in more detail, especially talking about using this within each project and not attempting it for a global install ..

mosabua (changes requested):
We should still add info for uv that you can use uv build and such after you manually did the pip install .. but in a configured proejct you have to do 

```
uv pip install keyring keyrings-chainguard-libraries --extra-index-url https://pypi.org/simple/
```

and then you can use uv build from Chainguard Libraries as configured otherwise in the project

mosabua (approved):
Looks good now. As discussed, if others run into trouble we can amend more in a follow up PR.

PR #2848, opened 2025-12-17
Update AWS assumable identity instructions to create identities with chainctl

ribbybibby:
We recently added commands to `chainctl` that massively simplify the process of creating identities for AWS.


SharpRake (approved):
Changes look great, thanks @ribbybibby!

PR #2849, opened 2025-12-17
Add note to Python cuda-enabled libraries 

angela-zhang:
Add a note / disclaimer on continuing to pull down NVIDIA Toolkit components from upstream sources.


content/chainguard/libraries/python/overview.md
```diff
@@ -142,6 +142,12 @@ replace the CUDA index you previously used with the corresponding
 Chainguard CUDA index, with your existing package management tools like 
 `pip`, `uv`, or `poetry`. 
 
+Note that the CUDA-enabled libraries in these repositories are not dependency 
+complete, as they do not include packages from the NVIDIA CUDA toolkit. You 
+must configure your package manager to pull these NVIDIA components from an 
+upstream index such as https://download.pytorch.org/whl/cu12x, or from PyPI
```
dannf: I don't think we should list pytorch.org here - I don't know that they carry these dependencies. But either PyPI or NVIDIA's index at https://pypi.nvidia.com will work. Thanks!

mosabua (approved):
Looks great. Nice work.

PR #2857, opened 2025-12-19
Add EKS Auto Mode installation instructions

patrickjburke245:
Added instructions for installing on AWS EKS Auto Mode, including creating a storage class and modifying the helm install command.


ltagliaferri (approved):
Some formatting nits to fix please, but then you can merge

patrickjburke245 (commented):
Fixed 2 issues that Lisa pointed out.

PR #2878, opened 2026-01-05
Add documentation for AWS identities with IAM outbound identity federation

ribbybibby:
AWS recently added a new feature that allows you to generate an OIDC token for an AWS IAM role or user.

This finally allows customers to do assumable identities on AWS in the same way they would on other platforms, without our crazy custom implementation.

I've marked the existing method as 'Legacy' but left it around.

I've also removed the EC2 and Lambda docs. At this point, I don't think either page really adds anything beyond the other two pages and it just adds confusion as to which page to read.


content/chainguard/administration/assumable-ids/identity-examples/aws-identity-oidc.md
```diff
@@ -0,0 +1,194 @@
...
+        {
+            "Effect": "Allow",
+            "Action": "sts:GetWebIdentityToken",
+            "Resource": "*",
+            "Condition": {
+                "ForAnyValue:StringEquals": {
+                    "sts:IdentityTokenAudience": "https://issuer.enforce.dev"
+                },
+                "NumericLessThanEquals": {
+                    "sts:DurationSeconds": 300
```
SharpRake: this line gives me an ugly warning that I'm having trouble resolving:

```
Ln 13, Col 20
	
Type Mismatch: Use the operator type String instead of operator numeric for the condition key sts:DurationSeconds. Learn more 
```

I fed the warning into droid and it suggested removing the condition:

```
The cleanest solution is to remove the problematic condition:

   json
     {
         "Version": "2012-10-17",
         "Statement": [
             {
                 "Effect": "Allow",
                 "Action": "sts:GetWebIdentityToken",
                 "Resource": "*",
                 "Condition": {
                     "ForAnyValue:StringEquals": {
                         "sts:IdentityTokenAudience": "https://issuer.enforce.dev"
                     }
                 }
             }
         ]
     }

   If you absolutely need to limit the duration, you would typically do that in the assume role call itself (via the DurationSeconds parameter when calling the STS API), not in the IAM policy. The policy condition keys for STS operations are limited and
   sts:DurationSeconds may not be supported for GetWebIdentityToken.
```

SharpRake (commented):
I think this is great! Thanks @ribbybibby 

There are a couple things I think need further review: the IAM policy gives me a warning and an optional `aws` command is failing for me. Once those are squared away we can publish this.

s-stumbo (approved):

PR #2890, opened 2026-01-07
Fix typos and improve clarity in PCI DSS documentation

patrickjburke245:
1. PCI DSS, not PCI DCC.
2. industry-leading should have a hyphen
3. One of the hyperlinks isn't visible from the docs page since the text is bolded
4. One of the hyperlinks had a markdown issue so wouldn't render, and for the same reason as #2, shouldn't be embedded in the bold text.


matthewhelmke (approved):
Thank you! This LGTM.

PR #2899, opened 2026-01-13
clarify some wording in Java library docs

s-stumbo:

## Type of change
Docs wording update

### What should this PR do?
Make wording clearer in a few places, remove outdated info about manual browsing

Clearer wording, fix inaccurate content

### DIFF
--- a/content/chainguard/libraries/java/overview.md
+++ b/content/chainguard/libraries/java/overview.md
@@ -63,11 +63,11 @@ The URL for the repository is:
 https://libraries.cgr.dev/java/
 ```
 
-This site provides a directory browsing and file listing capability similar to the Maven Central repository, allowing you to find available libraries, library versions, and available files. Learn more below under [Manual access](#manual).
+This site provides a directory browsing and file listing capability similar to the Maven Central repository, allowing you to find available libraries, library versions, and available files. Learn more under [Manual access](#manual).
 
 This Chainguard Libraries for Java repository uses the Maven repository format
 and only includes release artifacts of the libraries built by Chainguard from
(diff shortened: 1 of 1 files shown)


content/chainguard/libraries/java/overview.md
```diff
@@ -53,27 +53,24 @@ Libraries for Java.
...
+This site provides a directory browsing and file listing capability similar to the Maven Central repository, allowing you to find available libraries, library versions, and available files. Learn more below under [Manual access](#manual).
+
 This Chainguard Libraries for Java repository uses the Maven repository format
 and only includes release artifacts of the libraries built by Chainguard from
-source. Snapshot versions are not available.
-
-The URL does not expose a browsable directory structure. However, if you know the
-location of any particular artifact you can use the login credentials and a set
-path URL to access a file.
+source. It does. not include all artifacts from the Maven Central repository or other repositories. Snapshot versions are not available.
```
mosabua:
```suggestion
source. It does not include all artifacts from the Maven Central repository or other repositories. Snapshot versions are not available.
```

also wrap

mosabua (changes requested):
Good to go after a bit of reformatting and tweaking

matthewhelmke (approved):
One minor edit. It started with me removing a typo character and then I ended up creating a longer edit suggestion. Whether you take the entire suggestion or not, I'm approving based on the assumption that you will get rid of the extra period after "It does".

mosabua (approved):
Looks good now. Ideally you can fix source wrapping before merge but essentially this is good to go.

PR #2900, opened 2026-01-14
Formatting updates to library access docs

s-stumbo:

## Type of change
Formatting update

### What should this PR do?
- Use a numbered list instead of bullet points for ordered steps
- Update a button label in the console that was different from the docs
- Slight updates to instructions to orient the reader before giving an action
- Remove now-unnecessary steps for adding library access to a pull token, updated instructions for creating a token in the console for libraries

Improved clarity, console update

The steps should be easier to follow

### DIFF
--- a/content/chainguard/libraries/access.md
+++ b/content/chainguard/libraries/access.md
@@ -134,38 +134,22 @@ Refer to the following resources for more specific information for your needs:
 
 ### Creating pull tokens with the Chainguard console
 
-Use the following steps to create a pull tokens for Chainguard Libraries in the
-Chainguard console.
+Follow these steps to create a pull tokens for Chainguard Libraries in the
+Chainguard console:
 
 1. Use your authentication details to access the console at
   [https://console.chainguard.dev/](https://console.chainguard.dev/).
 2. In the left-hand navigation, click **Overview**.
 3. Click the **Manage pull tokens** tab, then click **Create access token**.
-  * Alternatively, select **Access Tokens** from the menu at the top of the **Settings** page.
+    - Alternatively, select **Access Tokens** from the menu at the top of the **Settings** page.
 4. Configure the access token:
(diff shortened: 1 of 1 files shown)


content/chainguard/libraries/access.md
```diff
@@ -195,7 +194,7 @@ Use the `env` environment output option to create a snippet for a new token
 suitable for integration in a script.
 
 ```shell
-$ chainctl auth pull-token --output env --repository=java --parent=example
+chainctl auth pull-token --output env --repository=java --parent=example
```
mosabua: Removing this is wrong .. the $ shows that this first line is the command .. the rest is the output .. you could separate into two code blocks but thats bad and overkill

mosabua (changes requested):
For the formmating changes I suggest to NOT do them in only one page .. if you really want to go with the numbered lists (and they are super painful for long lists of steps) then you should make it consistent and apply the changes across the whole libraries docs at least .. 

Just for reference .. we decided on purpose not to use numbering for the reason that it gets painful really quick when lists span paragraphs and the render engine (hugo and friends) automatically resets numbers.. 

Ultimately however you have to decide how to proceed about this with @matthewhelmke and @matteo-visentin-chainguard and @SharpRake and it should be in line with the overall Chainguard style guide.

matthewhelmke (approved):
I see the convo with @mosabua. When he is happy, I think this is ready to go.

content/chainguard/libraries/access.md
s-stumbo: I changed this back to include it

mosabua (approved):
Looks good to me but keep in mind I did not test this. Maybe @matthewhelmke or @angela-zhang or someone else should confirm

matthewhelmke:
The console steps have been tested and there weren't any other changes that had an impact on the instructions. Merge away!

PR #2905, opened 2026-01-16
Add FAQ about eBPF compatibility with Chainguard

patrickjburke245:


matthewhelmke (approved):
I'm requesting one minor change, otherwise LGTM. Thank you!!

PR #2912, opened 2026-01-21
Update custom-assembly.md

jasonbishay:
Replace `build_history.list` capability (which doesn't exist) with the correct capability `build_report.list` for the custom-ca role.

[X ] Check if this is a typo or other quick fix and ignore the rest :)

## Type of change

Typo fix
### What should this PR do?

Yes when customers copy the CA custom role it tries to add a capability that doesn't exist. Replaced `build_history.list` with the correct capability `build_report.list`

creating a custom ca role based off of our docs.

review

creating a custom ca role following the docs.


zachhackett (approved):

matthewhelmke (approved):
LGTM and TYSM!!

PR #2916, opened 2026-01-23
Fix: stateless firewalls need symmetric rules, not stateful.

justinprince:


matthewhelmke (approved):
LGTM. Thank you!

For others reading, I looked this up and confirmed @justinprince is correct and we need this change.

PR #2917, opened 2026-01-24
update zscaler blurb with http/2 configuration details

cmdpdx:

Update the zscaler blurb with additional configuration information we received from a prospect to prevent http/2 traffic from being downgraded, to enable chainctl usage.


ltagliaferri (approved):
Some tiny things

PR #2927, opened 2026-01-27
2 Small Changes

patrickjburke245:
ORGANIZATION -> $ORGANIZATION. Justification: the highlighting looks better, and this allows users to copy and paste commands from the docs to their terminal more easily, assuming they set an `ORGANIZATION` env variable.

Wherever `bash` or `sh` are listed as the command after the image, this won't work. The [entrypoint](https://images.chainguard.dev/directory/image/python-fips/specifications) for `latest` and `latest-dev` are both `/usr/bin/python`. Therefore, the entrypoint needs to be changed via the `--entrypoint` flag.


matthewhelmke (approved):
LGTM, ty

PR #2928, opened 2026-01-27
Add Amplitude tracking to docs

s-stumbo:

## Type of change
Adding Amplitude script to the docs site to start tracking docs metrics

### What should this PR do?
Add Amplitude tracking to docs pages

To gain insight into docs metrics

- The site should continue to function normally
- The legal team verified that it is ok to proceed with this implementation

- Run the site locally. Open browser dev tools, look in the Network tab, and filter for `amplitude`. We should see requests like `api2.amplitude.com` 
- I'm not sure if viewing the deploy preview will create a page view event that gets sent to Amplitude


matthewhelmke (approved):
We can't really test whether this works and is sending telemetry outside of pushing this to prod. The changes look safe enough, in terms of I don't think they'll crash our site or anything. So, based on that, let's see if it works!!

Approved!

PR #2932, opened 2026-01-28
renumber page weights as maintenance

matthewhelmke:
[x ] Check if this is a typo or other quick fix and ignore the rest :)

As we've added pages over the years, the space between the weights of each page has been reduced to where in some cases it was impossible to add a page between two others. This spaces those weights out and leaves a spot for a soon-to-come new page.

To check, just make sure I don't have any duplicate weights. I've checked the order multiple times. Also, there is no `030` because that's where I'm going to put the new page in an upcoming PR.


s-stumbo (approved):
Thanks for making this change!

PR #2941, opened 2026-01-30
Reorganize content on verifying and downloading SBOMs

s-stumbo:

## Type of change
Small re-org of content and deduplication effort

### What should this PR do?
- Have the list of predicate types on only one page
- Clarify that you can use the full URI or the shorthand version in the `download` command
- From the "Verify containers" page, link out to info about downloading attestations
- Put attestation types into a table

Deduplicate info about predicate types and improve clarity

- Links should work
- Information should be accurate
- The table should be readable

### DIFF
--- a/content/chainguard/chainguard-images/how-to-use/retrieve-image-sboms/index.md
+++ b/content/chainguard/chainguard-images/how-to-use/retrieve-image-sboms/index.md
@@ -1,6 +1,6 @@
 ---
-title: "How to Retrieve SBOMs for Chainguard Containers"
-linktitle: "Retrieve SBOM"
+title: "How to Retrieve Attestations and SBOMs for Chainguard Containers"
+linktitle: "Retrieve Attestations and SBOMs"
 aliases: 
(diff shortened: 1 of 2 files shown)


content/chainguard/chainguard-images/how-to-use/retrieve-image-sboms/index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: "How to Retrieve SBOMs for Chainguard Containers"
-linktitle: "Retrieve SBOM"
+title: "How to Retrieve Attestations and SBOMs for Chainguard Containers"
+linktitle: "Retrieve Attestations and SBOMs"
```
matthewhelmke: Since we talk about SBOMs first, I wonder whether we should put it first in the title. I think it is also a more likely term people would expect in the sidebar, so especially in the linktitle.

matthewhelmke (approved):
One comment to consider, but it's not a blocker.

wlynch (approved):

Issue #2955, opened 2026-02-05
Docs update: Split helm charts EDU page into two pages


matthewhelmke:
Since the iamguarded Helm charts and the newly-provided community charts are being marketed differently and listed differently in the UI, [the single page](https://edu.chainguard.dev/chainguard/chainguard-images/how-to-use/use-chainguard-helm-charts/) where the instructions for both should be split. This will make it easy to add a link to the documentation page specific to each type of chart from the READMEs for each type of chart.

EDIT to add link to the page in question.

PR #2958, opened 2026-02-06
Docs: Split helm page

matthewhelmke:
When I added the documentation for the new community Helm charts last week, I inserted those details into an existing page that was created for the iamguarded charts. That wasn't wrong, but it wasn't as clear as we hoped. Also, it would be useful to be able to link from READMEs to a docs page specific to the type of chart being perused, community or iamguarded.

So, this splits the page into two, one for community charts and one for iamguarded charts.

At the moment, everything in each page works and customers can be successful using them. However, the two charts have some detail differences that would be cool to capture in these pages. We may do this during this PR's review cycle or in future PRs, probably both.

I'm starting this review with @joshrwolf as he is the best SME on the Helm charts and we've already talked about me creating this PR.

Docs Team: please hold off on reviewing until Josh and I have finished refining based on his review. Thanks!


matthewhelmke:
Adding that this fixes #2955

s-stumbo (approved):

PR #2962, opened 2026-02-06
Clarify Network Requirements for libraries

s-stumbo:

## Type of change
Clarify existing content, add section about library storage locations

### What should this PR do?
- Rephrases the content under "Access for development tools" section
- Add that Chainguard Libraries are stored and served from GAR

[Internal questions](https://chainguard-dev.slack.com/archives/C085187D8RE/p1770330816117169) about where libraries are stored and feedback that the current docs imply you have to use a repo manager

### DIFF
--- a/content/chainguard/libraries/network-requirements.md
+++ b/content/chainguard/libraries/network-requirements.md
@@ -31,14 +31,14 @@ the following domains:
 
 ## Access for development tools
 
-Whether using a repository manager or accessing libraries directly, you must have HTTPS access to:
+Whether using a repository manager or accessing libraries directly, you must allow HTTPS access to:
 
 * `libraries.cgr.dev` for library access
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
LGTM, thank you!

mosabua (changes requested):
We should NOT provide the info that we are internally hosting artifacts on GAR. This is an internal implementation detail and partially also not true. Technical info like that should be confirmed with @angela-zhang and engineers potentially.

mosabua (changes requested):
The wording changes actually make it less correct than it was .. formatting improvements are good .. so I think we need some updates .. happy to chat next week .. travelling now

mosabua (approved):
LGTM now

PR #2964, opened 2026-02-06
add Python hash integrity docs

s-stumbo:

## Type of change
Add new section about hash verification to Python Library docs

### What should this PR do?
Add content about checksum mismatches that could affect Python configurations

Customers have reached out about this issue during migration


matthewhelmke (approved):
LGTM, thank you

matthewhelmke (approved):
This is useful, thank you!!
LGTM

mosabua (changes requested):
Good start but I think we need to flesh out what actually needs to be done with actual steps for each packaging tool and also keep in mind specific uses like pip with and without lock file. 

Also .. we need to clarify more how to avoid using of cached packages .. not just in the repo manager but also locally on dev machines and in CI/CD clusters potentially.

mosabua (commented):
I am not reviewing or testing this further at this stage since I dont have time to do so sufficiently.

mosabua (approved):
I am not sure how to remove my "requested changes" setup to unblock merge so I am approving with this review although I did not really test or review this in more detail. Proceed as seems reasonable to you @s-stumbo and @matthewhelmke

PR #2967, opened 2026-02-09
fix: update noproxy zscaler blurb to use html rather than md

cmdpdx:

The markdown formatting is not rendering as expected. 

image

Instead, try html.


xnox (approved):

matthewhelmke (approved):
LGTM, hopefully it does what we want

PR #2971, opened 2026-02-10
Add new API Reference section

wlynch:
### What should this PR do?

This change adds a new API Reference section and moves the OpenAPI and authentication docs there.

API documentation has been nested under Administration, and it's not always obvious that's where it is - we've gotten feedback where people could not find our REST documentation, so this is an attempt to make it more obvious that it exists.

- Aliases from old links (should be backwards compatible)

### DIFF
--- a/config/_default/menus/menus.en.toml
+++ b/config/_default/menus/menus.en.toml
@@ -38,6 +38,12 @@
   weight = 30
 
 [[main]]
+  name = "API Reference"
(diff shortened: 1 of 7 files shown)


konradzapalowicz (approved):

matthewhelmke (approved):
Thank you!!
The preview render has been problematic lately. This LGTM

PR #2980, opened 2026-02-16
Moving API Ref docs lower in left nav

SharpRake:

## Type of change

@ltagliaferri thought the new location for the API Reference docs was a little out of place ([reference](https://chainguard-dev.slack.com/archives/C03H64P08UT/p1771181023564819)). I'm inclined to agree, and this PR moves it lower in the left-hand nav, while keeping it as a top-level item under `Product Docs`. 

This PR moves it below `Chainguard OS` but above `Administration`, which I think is a good compromise, but we can tweak this further as necessary.

### What should this PR do?

n/a, quick fix

Product docs should focus on flagship products. While we can and should start treating the API as a product of that level, the official products should probably come before it in the nav.

LMK if you think this location is okay or if we should put it somewhere else.

Check the deploy preview ([link](https://deploy-preview-2980--ornate-narwhal-088216.netlify.app/chainguard/api/)), and let me know what you think.

### DIFF
--- a/content/chainguard/api/_index.md
+++ b/content/chainguard/api/_index.md
@@ -7,5 +7,5 @@ date: 2020-10-06T08:48:23+00:00
 lastmod: 2020-10-06T08:48:23+00:00
 draft: false
 images: []
-weight: 005
+weight: 033
 ---


ltagliaferri (approved):

matthewhelmke (approved):
Thank you! I like this better.

PR #2990, opened 2026-02-17
Fix package mappings automation

SharpRake:
The package mappings were not being updated automatically because the GitHub workflows were missing the curl step to fetch the latest data from the upstream dfc repository before building.

This commit:
- Adds the missing fetch step to cloud-run.yaml workflow
- Adds the missing fetch step to autodocs-platform.yaml workflow
- Updates the local package-mappings.yaml with 1,543 new mappings
- Updates the lastmod date in the documentation

The workflows will now fetch fresh mappings on every build, ensuring the documentation stays in sync with upstream changes.

### DIFF
--- a/.github/workflows/autodocs-platform.yaml
+++ b/.github/workflows/autodocs-platform.yaml
@@ -54,6 +54,14 @@ jobs:
     - name: npm install
       run: npm install
 
+    - name: Fetch latest package mappings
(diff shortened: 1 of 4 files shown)


matthewhelmke (approved):
Thanks for the update!!
LGTM

PR #2992, opened 2026-02-18
update rumble bigquery to fetch 31 days only

jamonation:
The vulnerability comparison and CVE pages like https://edu.chainguard.dev/chainguard/chainguard-images/vuln-comparison/node/ are failing to update. It turns out we have 70,000,000+ rows of data and querying them all is.. not optimal..

Since the pages are broken, I think we should land this to get them updating again. I ran this locally to update the cloud storage buckets with fresh data and the charts are current again.

### DIFF
--- a/tools/rumble/cmd/vulns.go
+++ b/tools/rumble/cmd/vulns.go
@@ -78,13 +78,12 @@ func (v *vulnsJson) generateJSON() error {
 		log.Fatalf("error performing AllVulnsWithImagesQuery: %v", err)
 	}
 
-	// Group results by vulnerability ID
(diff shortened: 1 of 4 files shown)


SharpRake (approved):
Comparisons aren't loading in the preview but I think that's to be expected. We can merge and tweak things if they're still not working.

Food for thought:  I've wondered about whether we want to keep these comparisons around at all, as we have vulnerability comparisons live in both the Console and Directory, and those comparisons are a bit more actively maintained.

jamonation:
I am also not sure about keeping them. I think the one major difference is the list of CVEs, and being able to browse them. I'm not sure if that logic lives in the console or directory at the moment.

PR #3011, opened 2026-02-23
network-requirements: document ip version

xnox:
We currently have a mix of ipv4 & ipv6 endpoints, document this.

Especially since we are about to gain ability to greenlight ipv6.


matthewhelmke (approved):
Thank you for adding this!
LGTM

PR #3012, opened 2026-02-23
Document active tag status feature in images catalog

s-stumbo:

## Type of change
Document new feature

### What should this PR do?
Explain how to see tag statuses in the Console

Upcoming feature

With the active tags FF on, look in the Console at the images list. Ensure the content aligns with what's in the console

### DIFF
--- a/content/chainguard/chainguard-images/about/versions.md
+++ b/content/chainguard/chainguard-images/about/versions.md
@@ -278,6 +278,25 @@ lifecycle of your Chainguard Containers.
 
 ### List Active Tags
 
+**Active tags in the Chainguard Console**
+
+You can view tag statuses in the Chainguard Console. In the Console under
+**Images > Organization**, under the **Status** column, you can see:
+
+- `Active` status for images you are entitled to
+- `Expired` status for images you are no longer entitled to
+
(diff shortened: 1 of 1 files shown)


content/chainguard/chainguard-images/about/versions.md
```diff
@@ -278,6 +278,25 @@ lifecycle of your Chainguard Containers.
...
+You can view tag statuses in the Chainguard Console. In the Console under
+**Images > Organization**, under the **Status** column, you can see:
+
+- `Active` status for images you are entitled to
+- `Expired` status for images you are no longer entitled to
+
+If you use Chainguard Production Containers, it is also possible to view the
+statuses of different tag versions for an image. Click into an image to see the
+statuses alongside tag versions. If a tag is no longer being actively
+maintained, it displays in gray text and a **pause** icon appears next to its
```
SharpRake: I'm not sure I'd call that a pause icon. What I see is a circular arrow with a line through it.

Image

SharpRake (approved):
This is good! One non-blocking comment about what constitutes a pause icon, but don't let that stop you from publishing

PR #3029, opened 2026-02-27
Document fat JAR limitation for verification

s-stumbo:

## Type of change
Clarification 

### What should this PR do?
Document limitations around verification for fat JARs

Customer confusion, [internal feedback in Slack](https://chainguard-dev.slack.com/archives/C085187D8RE/p1772034146111659)

Make sure new content is clear and accurate

Verify the behavior of `chainctl libraries verify` on an individual JAR from the Maven mache vs. against a fat JAR

### DIFF
--- a/content/chainguard/libraries/verification.md
+++ b/content/chainguard/libraries/verification.md
@@ -133,9 +133,9 @@ pip3 install -r requirements.txt
 chainctl libraries verify --detailed ./venv/
 ```
 
-### Analyze a local Java JAR file
+### Analyze a Java JAR file
 
-Analyze a local Java `.jar` file:
+Analyze a Java `.jar` file:
 
 ```sh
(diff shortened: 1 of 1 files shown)


content/chainguard/libraries/verification.md
```diff
@@ -131,6 +133,8 @@ pip3 install -r requirements.txt
 chainctl libraries verify --detailed ./venv/
 ```
 
+### Analyze a local Java JAR file
```
matthewhelmke: In line 115 it's just a "a Python wheel file". Is there a distinction where we need "local" here? Do we need it in both?

matthewhelmke (approved):
This is useful, thanks!
LGTM, with just one question to consider before you merge.

mosabua (approved):
Some minor suggestions but essentially good to go

PR #3045, opened 2026-03-04
Clarify sources for Java libraries

s-stumbo:

## Type of change
clarify wording

### What should this PR do?
- Explains that Maven is the primary source, but other platforms are supported
- Add mention of Confluent and Gradle plugins to the list of other sources

Internal/customer feedback. [Slack thread here](https://chainguard-dev.slack.com/archives/C085187D8RE/p1767978789904909).


mosabua (changes requested):

mosabua (approved):
Two minor things to address but then good to go

PR #3051, opened 2026-03-09
Update wording on library access page

s-stumbo:

## Type of change
Wording update

### What should this PR do?
Make it clear that you "download" library artifacts 

Internal request

Content should be accurate


matthewhelmke (approved):
This LGTM, thank you! We can revise further in future PRs if needed.

PR #3064, opened 2026-03-11
Update chainctl example to remove redundant characters

s-stumbo:

## Type of change
Update chainctl command based on internal feedback that the `./` in `./venv/` was redundant


mosabua (approved):
Looks good

PR #3074, opened 2026-03-13
fixes H2 Header formatting issue

### LINKED ISSUE #2602: H2 Header formatting issue
Noticed a strange formatting issue in my [latest article](https://edu.chainguard.dev/chainguard/chainguard-images/tooling/building-go-containers-with-ko/):

Image

The source for these headers is: 
`## Step 2 — (Re) Authenticating to the Chainguard Registry`
`## Step 3 — Building Your First Container with Ko`

It seems to be related to the table of contents, as if the header in the tutorial is also applying the same modifications (CSS? JS?) to remove the word "Step" and format the number.

jijo-OO7:
## Bug fix (H2 Header formatting issue ) | Fixes: #2602

[✅] Fix ``assets/js/titles.js`` that strips "Step" from heading text

preview

### DIFF
--- a/assets/js/titles.js
+++ b/assets/js/titles.js
@@ -4,7 +4,7 @@ document.addEventListener("DOMContentLoaded", () => {
     if (match) {
       h2.innerHTML = `<div class="step-container">${h2.innerHTML.replace(
         match[0],
-        `<span class="step">${match[1]}</span>`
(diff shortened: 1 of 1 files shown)


SharpRake (approved):
Confirmed that the current styling doesn't look right. Thanks for the fix @jijo-OO7

PR #3084, opened 2026-03-16
JS libraries Artifactory integrity issue with R2 redirects

s-stumbo:

## Type of change
Add content on how to prevent JS checksum mismatches with Artifactory virtual repository

### What should this PR do?
Help customers prevent migration issues

Internal request / customer reported issue. [Slack thread here](https://chainguard-dev.slack.com/archives/C0962EGMS3F/p1773671606110999)

This content should be accurate. Working fix described here: https://github.com/chainguard-dev/customer-issues/issues/3217#issuecomment-4068348309

### DIFF
--- a/content/chainguard/libraries/javascript/global-configuration.md
+++ b/content/chainguard/libraries/javascript/global-configuration.md
@@ -201,16 +201,27 @@ virtual repository.
 
 ### Advanced settings for redirect handling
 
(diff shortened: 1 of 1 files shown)


dannf (approved):
Awesome, thanks! I can't vouch for the accuracy because I don't have a setup to reproduce/verify. But it all sounds reasonable to me..

matthewhelmke (approved):
I see your information source noted. Thank you. I think the solution should work.

I read through this and it LGTM, too.

PR #3085, opened 2026-03-16
Docs: Create a top level page for Chainguard Repository

matthewhelmke:
Closes https://github.com/chainguard-dev/internal/issues/5740

This is nowhere near ready to merge, but I wanted to get something in here for @s-stumbo and @SharpRake to comment on before I leave for the day.

I'm actually starting to think that the two child pages are extraneous. What do you two think? Can we make that argument?

I can add more content and detail tomorrow, but suggestions are very welcome given our quick turnaround time.


SharpRake (approved):
some minor suggestions

PR #3090, opened 2026-03-17
Clarify auth for Chainguard Repository

s-stumbo:

## Type of change
Clarify existing content around direct access for JS

### What should this PR do?
- Clarify that you do need to auth for Chainguard Repository, but it can be an existing pull token

- [Internal feedback](https://chainguard-dev.slack.com/archives/C03H64P08UT/p1773722836980079) 

- Make sure content is clear and accurate

Follow the docs to test out setting up Chainguard repo for JS


matthewhelmke (approved):
I did not test the new code, but it passes a sanity check. It should work. If you trust the source and/or have tested it yourself, I think this LGTM.

mosabua (changes requested):

content/chainguard/libraries/access.md
```diff
@@ -239,6 +239,10 @@ Chainguard Libraries for
 Java](/chainguard/libraries/java/build-configuration/#bazel) or for manual
 testing with curl.
 
+> Note: `.netrc` authentication is not supported by npm, pnpm, or Yarn. For
```
mosabua: I would remove this .. also the the auth is not always based on npmrc .. so I would leave that out as well. This section already talks about where it works .. maybe if you really think its required you could say that .netrc doesnt work with anything else .. but thats also too board .. for example in the Java ecosystem it doesnt work with Gradle or Maven .. but thats not worth mentioning .. or Apache Ant ..

content/chainguard/libraries/access.md
```diff
@@ -248,15 +252,15 @@ login YOUR_USERNAME_FOR_REPOSITORY_MANAGER
 password YOUR_PASSWORD
 ```
 
-For a direct connection to Chainguard Libraries, for example for testing with
+For a direct connection to Chainguard Libraries for Java or Python, for example for testing with
```
mosabua: Bazel is rare in Java so I dont think this is worth changing here .. if anything you could make it specific about Python

mosabua (changes requested):

mosabua (approved):
I have not tested this so will let you decide how to proceed.

PR #3093, opened 2026-03-17
Docs for console_viewer role and ability to manage custom roles in Console

s-stumbo:

## Type of change
Docs for product update

### What should this PR do?
- Explain the new console_viewer role
- Explain how to create/edit custom roles and bind roles in the Console 

To update the docs based on current product functionality

- Content should be accurate


matthewhelmke (approved):
This is useful!
LGTM

gracelarsen-star (changes requested):
- viewer - Read-only access across all resources, and **can** pull images
- console_viewer - Read-only access across all resources, and **cannot** pull images

We also need to add console_viewer to the Role Capabilities Comparison table

gracelarsen-star (approved):

PR #3097, opened 2026-03-17
Reorder regular and remediated repos in Python example

s-stumbo:

## Type of change
Clarify docs based on [internal feedback](https://chainguard-dev.slack.com/archives/C085187D8RE/p1773771854731199)/best practices

### What should this PR do?
Make it more obvious that you should list the remediated repo first to ensure it takes precedence

Internal feedback


wiedenmeier (approved):
Looks good - I think this is more correct, since pip may not always fallback to another index to look for additional versions if the first index it looks at has the requested package.

matthewhelmke (approved):
LGTM

PR #3099, opened 2026-03-18
Add content about what you can do with Custom Assembly

s-stumbo:

## Type of change
Update to make existing content easier to find

### What should this PR do?
- Rephrase the intro paragraph for Custom Assembly
- Add paragraph stating the ways you can customize an image

[Internal feedback](https://chainguard-dev.slack.com/archives/C07JZDR9HDG/p1773835451772619) that this info was hard to find 

- Content should be accurate


matthewhelmke (approved):
This is clear and useful! Thank you!!

LGTM

PR #3101, opened 2026-03-18
Add caveats from SA to unique tags

matthewhelmke:
Unique tags have caused many customer problems and the SA and SE teams would love for us to put stronger information in docs to prevent most people from using them. This adds those caveats.


SharpRake (approved):
One non-blocking suggestion

PR #3103, opened 2026-03-19
Add docs for private npm packages with Chainguard Repository

s-stumbo:

## Type of change
Add guidance for private npm packages alongside Chainguard Repository

### What should this PR do?
Explain how to configure your .npmrc to hit public npm for packages with a `@prefix/` and Chainguard Repository for everything else

A customer ran into issues, [reported internally in slack](https://chainguard-dev.slack.com/archives/C09MK1VTHL6/p1773874717040079)

- Content should be clear and accurate
- Note that this draft does not address other configurations (like artifact managers), as we still need to do additional testing

Follow Dylan's video in Slack to see how to recreate this issue

### DIFF
--- a/content/chainguard/libraries/javascript/build-configuration.md
+++ b/content/chainguard/libraries/javascript/build-configuration.md
@@ -78,7 +78,7 @@ with a specific prefix to a different registry:
 ```
 # .npmrc
 registry=https://libraries.cgr.dev/javascript/
-//libraries.cgr.dev/javascript/:_auth=<base64-encoded-token>
+//libraries.cgr.dev/javascript/:_auth={$token}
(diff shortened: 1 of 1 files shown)


mosabua (changes requested):
The generic info applies but the steps have to be broken up and tested separately for npm, pnpm, yarn classic, and yarn berry and then documented in these section .. it might not even work for all of them but it will certainly be different ...

mosabua (approved):
I cant do a full review/test of this but directionally this seems ok.

matthewhelmke (approved):
Oh, wow. This feels like important info to get out there!
LGTM, merge away!!

PR #3111, opened 2026-03-20
Strengthen language about pinning for Helm charts

matthewhelmke:
I was asked to add language about pinning to **digest** to the Helm docs, similar to language we use about pinning for images.

Turns out, that was already the recommendation for -iamguarded charts, but not the newer community charts, which recommended to pin to tag like upstream.

This modifies both.


s-stumbo (approved):
LGTM!

PR #3116, opened 2026-03-23
Update access.md

kallewesterling:
Wrong `--repository` if the user is to list expired pull tokens for Chainguard Libraries for Python


SharpRake (approved):
:+1:

PR #3120, opened 2026-03-24
Restructure Chainguard Repository docs slightly

matthewhelmke:

This changes from using just a directory and `_index.md` file by moving the content into an `overview.md` file and creating a new base `_index.md` that matches other sections (I copy/pasted/edited the one from Chainguard OS), just so that Hugo has things working consistently.


s-stumbo (approved):
LGTM!

PR #3128, opened 2026-03-27
Insert guidance for apk add

matthewhelmke:
Fixes https://github.com/chainguard-dev/internal/issues/5738

While looking through our docs, I feel like these are the two most likely places people using APK packages from Chainguard will look. During review, if you think of other places or better ways to guide, lmk.

Custom assembly won't let the selection of conflicting packages happen, but I put a note in there just as an example of what someone might try and why a fail would occur. I put the other in notes about building Docker containers with APK packages, where users definitely could encounter this blindly. Hopefully the note is enough.


SharpRake (approved):
One styling point/question, otherwise looks good!

PR #3132, opened 2026-03-30
Document support of npm ping

s-stumbo:

## Type of change
Add mention of using `npm ping` as it is now a supported way to verify auth

### What should this PR do?
Add information about using `npm ping` to the npm minimal project example. It seemed like the best location to add this information since users following the example would be likely to need an auth check before adding dependencies.

Content should be clear and accurate and on a page/section where it makes sense.

In a directory with a .npmrc configured to point to `https://libraries.cgr.dev/javascript/` with valid credentials, run `npm ping --userconfig .npmrc` and confirm you get a `PONG` response


ajayk (approved):

matthewhelmke (approved):
LGTM

PR #3141, opened 2026-03-31
Fix typos on Java page

s-stumbo:

## Type of change
apparently trying to style `</` as code turns it into a sideways arrow? Rephrasing something to work around that


SharpRake (approved):
:+1:

PR #3148, opened 2026-04-01
add how to config Docker credential helper on Windows

timm13:


SharpRake (approved):
I moved the new section down to the existing Docker credential helper section and made a few tweaks to the headings. Thanks for the addition!

PR #3153, opened 2026-04-02
Remove cuda130 from Python overview

s-stumbo:

## Type of change
Removing cuda130 repo from the list in Python overview because it is currently empty and returns 403s https://github.com/chainguard-dev/customer-issues/issues/3093


matthewhelmke (approved):
Thank you!

PR #3208, opened 2026-04-09
Update chainctl info in Quickstart and Access

s-stumbo:

## Type of change
Cross-linking to chainctl docs for the `libraries entitlements` command.

### What should this PR do?
- On both the "Quick start" and "Access" pages, link to the chainctl docs 
- Update the chainctl entitlements info on the "Access" page
- Make it clearer that repo manager is the recommended approach
- Update formatting of the "Configure your build tools" section

To provide additional info on the commands, make recommendations clearer


SharpRake (approved):
Left some thoughts, but overall these changes look good!

PR #3237, opened 2026-04-13
Add chainctl entitlement command to JS library overview

s-stumbo:

## Type of change
Update to JS library overview page

### What should this PR do?
- Remove the instruction to contact Chainguard to enable upstream fallback, replace it with how to use chainctl to create an entitlement and update the upstream policy

Because customers can now self-serve this policy change

- Accurate command
- Accurate instructions

Make sure the content appears as expected on the JS overview page


SharpRake (approved):
:+1:

PR #3245, opened 2026-04-14
Add USER root disclaimer to porting sample app guide

SharpRake:
## Type of change
**Documentation Update**
- Added disclaimer to the `USER root` key point clarifying it is a temporary escalation only, and that users should switch to a dedicated non-root user after completing root-level operations

## What should this PR do?

resolves https://github.com/chainguard-dev/internal/issues/4783

## Why are we making this change?

A customer migrating to Chainguard Images interpreted the `USER root` guidance as a permanent setting rather than a temporary step for software installation. The updated wording makes clear that containers should revert to a dedicated non-root user after root-level operations, preserving the security benefits of minimal images.

## What are the acceptance criteria?

- The "Porting Key Points" bullet on `USER root` explicitly states it is a temporary escalation
- The bullet advises creating and switching to a dedicated non-root user (or using the image's built-in non-root user) after root-level operations
- The disclaimer is clear and actionable without being overly verbose

## How should this PR be tested?

1. Check the preview link and verify the updated bullet point renders correctly in the "Porting Key Points" section
2. Confirm the disclaimer reads naturally within the existing list

### DIFF
--- a/content/chainguard/migration/porting-apps-to-chainguard/index.md
+++ b/content/chainguard/migration/porting-apps-to-chainguard/index.md
@@ -21,7 +21,7 @@ toc: true
 ### Porting Key Points
 
 * Chainguard's distroless Containers have no shell or package manager by default. This is great for security, but sometimes you need these things, especially in builder images. For those cases we have `-dev` variants (such as `cgr.dev/chainguard/python:latest-dev`) which do include a shell and package manager.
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
LGTM
I'm surprised we didn't already say this! Good call to add!!

PR #3248, opened 2026-04-15
Fix: Link to `configurable upstream fallback`

garrying:
- Fix: Link to `configurable upstream fallback`


SharpRake (approved):
thanks @garrying!

PR #3252, opened 2026-04-17
Cleanup: Updated content link style

garrying:
- Underline content links
- Updated color for links
- Reduce codeblock header height
- Fix `Manual access` anchor link in `Java` overview


SharpRake (approved):
thanks @garrying!

PR #3260, opened 2026-04-21
Add JS library SBOM retrieval command

s-stumbo:

## Type of change
Add SBOM retrieval command to JS library docs, also note a troubleshooting step to update chainctl

### What should this PR do?
Explain how to retrieve SBOMs for JS libraries

Product update

Ensure the commands work as described and the content appears in the right location

Review the deploy preview

### DIFF
--- a/content/chainguard/libraries/javascript/build-configuration.md
+++ b/content/chainguard/libraries/javascript/build-configuration.md
@@ -214,7 +214,7 @@ directory:
 chainctl auth configure-npm --pull-token
 ```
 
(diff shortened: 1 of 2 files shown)


angela-zhang (approved):

matthewhelmke (approved):
This is fine. Consider maybe linking to https://edu.chainguard.dev/chainguard/chainctl-usage/how-to-install-chainctl/#updating-chainctl or giving the command to update so users don't have to figure it out themselves.

PR #3268, opened 2026-04-23
Update docs on AI bundle: clarify GitHub release assets are static, update Quick Start to use container

matthewhelmke:
## Summary

  - Adds a note to the GitHub Release section warning that release assets are static and not automatically updated
  - Renames the release link from "Latest Release" to "Archived Bundle" to avoid implying freshness
  - Replaces the Quick Start curl command (which downloads from the frozen release) with the docker run ... extract
  command, which pulls from the live container image

Preview: https://deploy-preview-3268--ornate-narwhal-088216.netlify.app/developer-resources/#download-options

 ## Context

  The ai-docs-bundle GitHub Release is covered by an org ruleset that makes release assets immutable. The workflow attempts to update them on each build but fails silently (the step has continue-on-error: true), meaning users who download from the release get stale documentation. The container (ghcr.io/chainguard-dev/ai-docs:latest) is updated correctly on each build and is the right path for current docs.

  This PR updates the page to reflect that reality while a longer-term fix for the release update workflow is
  investigated.

### DIFF
--- a/content/ai-docs-security.md
+++ b/content/ai-docs-security.md
@@ -88,6 +88,8 @@ Example patterns we redact:
 
 ### Direct Download Verification
 
+> **Note:** The assets in this release are static and not updated automatically. These verification steps apply to the archived snapshot. For current documentation, use the container distribution.
+
 ```bash
 # 1. Download files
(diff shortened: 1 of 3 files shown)


matthewhelmke:
Added a commit that updates places where static (outdated) files were being used with clarifying notes.

s-stumbo (approved):

PR #3269, opened 2026-04-23
AI docs bundle fix: switch AI docs releases to dated tags, and more

matthewhelmke:
fix: switch AI docs releases to dated tags, consolidate release creation in GCS workflow

Summary

  - Replaces the frozen ai-docs-bundle release tag with dated tags (ai-docs-YYYYMMDD-HHMMSS), which are not subject to
   the org immutable releases ruleset
  - Adds cosign blob signing (.bundle files) to the GCS workflow, matching the signing that previously only existed in
   the other two workflows
  - Removes release creation from compile-public-docs.yml and compile-docs.yml — the GCS workflow is now the sole
  release creator, as it's the only workflow with complete content from all source repositories
  - Adds a cleanup step to retain only the 10 most recent dated releases
  - Updates all user-facing download URLs in three content pages from hardcoded /releases/download/ai-docs-bundle/ to
  /releases/latest/download/, which GitHub redirects to the most recent --latest release

  Context

  The ai-docs-bundle GitHub Release tag was covered by an org immutable releases ruleset. gh release upload --clobber
  silently failed on every run (the step had continue-on-error: true), leaving release assets permanently frozen while
   the container and GCS continued updating correctly. The security team confirmed there is no way to create a per-tag
   exemption without disabling the ruleset entirely.

  Testing (has been done)

  - Compilation pipeline validated: triggered workflow_dispatch on this branch against chainguard-dev/edu — completed
  successfully, confirming GCS authentication, all-repo download, compilation, and GCS upload are unaffected by these
  changes
  - Cleanup logic validated: tested the jq filter locally with a simulated 12-release + ai-docs-bundle input —
  correctly identified the 2 oldest dated releases for deletion, left the 10 most recent untouched, and ignored
  ai-docs-bundle entirely
  - Release creation and signing steps are gated on if: github.ref == 'refs/heads/main' and cannot be fully exercised
  pre-merge; the logic uses standard gh release create with a guaranteed-unique timestamped tag, which cannot
  encounter the immutability error

### DIFF
--- a/.github/workflows/compile-ai-docs-from-gcs.yaml
+++ b/.github/workflows/compile-ai-docs-from-gcs.yaml
@@ -331,39 +331,51 @@ jobs:
         
         echo "✓ Bundle uploaded to GCS"
 
-    - name: Update GitHub Release
+    - name: Sign bundle files
       if: github.ref == 'refs/heads/main'
-      continue-on-error: true  # Org ruleset makes release assets immutable; container/GCS are primary distribution
+      run: |
+        cd static/downloads
+        cosign sign-blob chainguard-ai-docs.tar.gz \
+          --yes \
+          --bundle=chainguard-ai-docs.tar.gz.bundle
+        cosign sign-blob chainguard-ai-docs.md \
+          --yes \
+          --bundle=chainguard-ai-docs.md.bundle
+        echo "Bundle files signed"
+
+    - name: Create GitHub Release
+      if: github.ref == 'refs/heads/main'
       env:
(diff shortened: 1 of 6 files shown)


matthewhelmke:
Post-merge action required

  After the first new dated release is created, update the frozen ai-docs-bundle release description:
  gh release edit ai-docs-bundle \
    --repo chainguard-dev/edu \
    --title "Archived Bundle (superseded)" \
    --notes "This release is archived and no longer updated. See the [latest
  release](https://github.com/chainguard-dev/edu/releases/latest) for current documentation."

s-stumbo (approved):

PR #3270, opened 2026-04-23
fix(docs) correct npm env var substitution syntax in .npmrc example

indexzero:
## What

Fixes the `.npmrc` env var syntax in the scoped registry example in `content/chainguard/libraries/javascript/build-configuration.md` and the static docs bundle. `{$token}` → `${token}`.

## Why

npm's `.npmrc` env var substitution requires the dollar sign before the brace (`${VAR}`). The previous `{$token}` syntax is not recognized by npm and would be passed as a literal string, silently breaking authentication for any user who copied the snippet.

## References

- [npm docs: Environment Variables](https://docs.npmjs.com/cli/v10/configuring-npm/npmrc#environment-variables)

### DIFF
--- a/content/chainguard/libraries/javascript/build-configuration.md
+++ b/content/chainguard/libraries/javascript/build-configuration.md
@@ -196,7 +196,7 @@ with a specific prefix to a different registry:
 ```
 # .npmrc
 registry=https://libraries.cgr.dev/javascript/
-//libraries.cgr.dev/javascript/:_auth={$token}
+//libraries.cgr.dev/javascript/:_auth=${token}
 
 @your-org:registry=https://registry.npmjs.org/
 ```


static/downloads/chainguard-complete-docs.md
```diff
@@ -31769,7 +31769,7 @@ with a specific prefix to a different registry:
 ```
 # .npmrc
 registry=https://libraries.cgr.dev/javascript/
-//libraries.cgr.dev/javascript/:_auth={$token}
+//libraries.cgr.dev/javascript/:_auth=${token}
```
mosabua, line 31772: I think this one is generated so I dont think you should fix here. In fact @s-stumbo @SharpRake @matthewhelmke .. maybe this file should not even be checked in..

content/chainguard/libraries/javascript/build-configuration.md
```diff
@@ -196,7 +196,7 @@ with a specific prefix to a different registry:
 ```
 # .npmrc
 registry=https://libraries.cgr.dev/javascript/
-//libraries.cgr.dev/javascript/:_auth={$token}
+//libraries.cgr.dev/javascript/:_auth=${token}
```
mosabua, line 199: @s-stumbo seems like this is the only place .. the original setups had this correct. Might need to check your example project

mosabua (approved):

static/downloads/chainguard-complete-docs.md
matthewhelmke: Agreed. The `scripts/compile_docs.py` and `scripts/compile_docs_secure.py` scripts each create/update this file at different times. I would want to spend some time investigating further before changing anything directly.

I can probably look into this next week and figure out making the syntax changes in the best locations. That might be preferable to merging this PR.

mosabua: I would just back this change here out and then merge the PR because its a valid wrong content at the moment. Also from merging you would find out what happens potentially.. but will leave that to you all

matthewhelmke: I'd be okay with that on a M-Th, but maybe not on a Friday afternoon. Only because I won't be around starting in about an hour until Monday.

matthewhelmke: Okay, all. I spent the time this morning doing proper due diligence. This static file is built by scripts that use the actual content as the source. So fixing here won't matter long-term, because it will be overwritten with the (corrected) content in the other file. But it does make sense from now to when this is next triggered. So, correcting both files is the proper thing to do, which is what @indexzero was doing. Thanks for the patience as I got caught up!! It was just too much for my tired Friday brain to process.

I'm going to approve and merge.

matthewhelmke (approved):

PR #3275, opened 2026-04-27
Fix typo in versions.md regarding floating tags

addyess:
This looks like a mistake in the floating tags explanation.


matthewhelmke (approved):
Nice catch! Thank you!! I'll merge this right away.

PR #3276, opened 2026-04-27
Cursor docs improvements

s-stumbo:

## Type of change
Batch of small updates to improve the Cursor docs

### What should this PR do?
- Clarify that Cursor expects an organization name that looks like a domain
- Recommend injecting creds via env variables
- Add note to review changes before running a build or deploying
- Note that recommends not using `latest` tag in production
- Note that Chainguard doesn't mirror every package available on public registries
- Add FAQ for when a package is not found (instructions for enabling upstream fallback)

[Internal questions from AEs](https://docs.google.com/document/d/16XKcHgGt4QAExFc9QnhNTly8w7S9jUe7pd8B8NWNzeg/edit?tab=t.0)

Review the deploy preview

### DIFF
--- a/content/chainguard/integrations/cursor.md
+++ b/content/chainguard/integrations/cursor.md
@@ -20,9 +20,10 @@ This page explains how to start using Chainguard artifacts in Cursor.
 
 Before you begin, you'll need:
 * A [Cursor](https://cursor.com/) account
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
LGTM

PR #3290, opened 2026-04-30
Fix: allow release-assets.githubusercontent.com in AI docs workflow

matthewhelmke:
GitHub now redirects release asset downloads to release-assets.githubusercontent.com, which was blocked by harden-runner's egress policy, causing cosign install to fail with curl exit code 6 (could not resolve host).

This fixes that issue.


matthewhelmke:
Fixes https://github.com/chainguard-dev/internal/issues/5830

PR #3298, opened 2026-05-05
Remove page on updating bash on macOS

s-stumbo:

## Type of change
Remove page from public docs

### What should this PR do?
Remove the "Updating bash on macOS" page, which is public/findable but not visible in the site menu

It is a remnant of Enforce docs and is no longer relevant. Described in issue here: https://github.com/orgs/chainguard-dev/projects/166/views/14?pane=issue&itemId=175432058&issue=chainguard-dev%7Cinternal%7C5790
There is a PR open to move this content to the mono repo: https://github.com/chainguard-dev/mono/pull/38700

In deploy preview, the page should not be accessible


matthewhelmke (approved):
Thank you

PR #3308, opened 2026-05-08
Update pnpm verification command

s-stumbo:

## Type of change
Update existing Library Verification page

### What should this PR do?
Update the command for verifying pnpm stores

[Feedback](https://chainguard-dev.slack.com/archives/C085187D8RE/p1778206497044769) that the command doesn't work as-is for pnpm; you need to add `--store-dir` and point to that

Review the deploy preview

### DIFF
--- a/content/chainguard/libraries/verification.md
+++ b/content/chainguard/libraries/verification.md
@@ -243,16 +243,32 @@ chainctl libraries verify "$(npm config get cache)"
 
 #### Verify a pnpm store
 
-Verify your pnpm store:
+Verify your pnpm store. Use `--store-dir` to install to an explicit path and verify that location:
 
 ```sh
-chainctl libraries verify "$(pnpm store path)"
+pnpm install --store-dir /tmp/my-pnpm-store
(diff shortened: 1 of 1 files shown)


content/chainguard/libraries/verification.md
```diff
@@ -243,17 +243,33 @@ chainctl libraries verify "$(npm config get cache)"
...
 
+Note that in `pnpm-lock.yaml`, packages resolved from Chainguard have only a
+`resolution:` entry with an integrity hash:
+
+```yaml
+supports-color@7.2.0:
+  resolution: {integrity: sha512-LPhWJX...}
+```
+
+Packages that fell back to the upstream registry include an explicit `tarball:` URL pointing to `javascript-upstream`. For example:
```
angela-zhang:
```suggestion
Packages that are pulled from the upstream fallback include an explicit `tarball:` URL pointing to `javascript-upstream`. For example:
```

matthewhelmke (approved):

angela-zhang (approved):

PR #3311, opened 2026-05-11
Add notifications troubleshooting FAQ

matthewhelmke:
Adding info directly from customer success to help prevent support calls/tickets.


matthewhelmke:
I'm merging without review today, but let's iterate next week if there are refinements. This is useful enough to just get it out there.

PR #3329, opened 2026-05-15
Update Guardener role requirements on capabilities page

s-stumbo:

## Type of change
Update to Built-in roles & capabilities page

### What should this PR do?
Add info about the new Guardener roles

New roles were recently added for Guardener

Content should be clear and accurate. We should check whether this information needs to be added anywhere else.

Review the deploy preview


matthewhelmke (approved):
LGTM

PR #3339, opened 2026-05-19
Add step to select charts org for Helm use

matthewhelmke:
Fixes https://github.com/chainguard-dev/internal/issues/5841

Adds a note that users need to select the `charts` source when trying to pull Helm charts. Provides a sample of what they see in the terminal as well.


s-stumbo (approved):
LGTM

PR #3341, opened 2026-05-19
Update JS build config pnpm instructions

s-stumbo:

## Type of change
Update to build config and JS overview

### What should this PR do?
- Correct the npm instructions - they mentioned the npmrc but didn't have the correct command to set the npmrc. It now specifies using the `--location=project `flag to set a project-level configuration, and mentions including auth
- Removed content from the JS overview because it is no longer relevant. The `update-hashes` command functionality has been updated so we don't need to mention the upstream package issue.

The pnpm docs didn't work as written

Review the deploy preview


matthewhelmke (approved):
LGTM

angela-zhang (approved):

PR #3346, opened 2026-05-20
update-hashes support for repo managers

s-stumbo:

## Type of change
Update existing Python and JS docs: overview, build config, global config

### What should this PR do?
- Reorganize content about the update-hashes command
- For Python, add update-hashes content to "Build config" to make it consistent with the JS docs
- Add new content explaining how to use it for repo managers 

We are introducing new functionality

The content should appear where it makes sense, and should be clear and accurate

Review the deploy preview


matthewhelmke (approved):
Very useful!! LGTM

ppalucha (commented):
This seems ok for me, but the same would also apply to https://github.com/chainguard-dev/edu/blob/main/content/chainguard/libraries/python/overview.md

PR #3347, opened 2026-05-21
chore: remove rust FIPS FAQ entry

mbarretta:
## Summary
- Removes the FAQ entry explaining why Chainguard does not build Rust FIPS images, since we now do build them.

## Test plan
- [ ] Confirm the FAQ page renders correctly without the removed section


matthewhelmke (approved):
LGTM, thank you!!

PR #3354, opened 2026-05-26
Image Matcher API - migration docs

s-stumbo:

## Type of change
New pages in the Migration section

### What should this PR do?
- Explain how to use the Image Matcher API

[Docs update request](https://github.com/orgs/chainguard-dev/projects/166/views/14?pane=issue&itemId=192327572&issue=chainguard-dev%7Cinternal%7C5854)

- Content should be clear and accurate, appear in the expected location

Review the deploy preview

### DIFF
--- a/content/chainguard/migration/image-matcher/_index.md
+++ b/content/chainguard/migration/image-matcher/_index.md
@@ -1,0 +1,12 @@
+---
+title: "Image Matcher"
+linktitle: "Image Matcher"
+aliases:
+description: "Language- or platform-specific tutorials outlining how you can migrate applications to Chainguard Containers."
+type: "article"
+date: 2026-05-26T00:00:00+00:00
+lastmod: 2026-05-26T00:00:00+00:00
+draft: false
+images: []
+weight: 050
+---
--- a/content/chainguard/migration/image-matcher/_index.md
+++ b/content/chainguard/migration/image-matcher/_index.md
@@ -1,8 +1,7 @@
 ---
 title: "Image Matcher"
 linktitle: "Image Matcher"
-aliases:
-description: "Language- or platform-specific tutorials outlining how you can migrate applications to Chainguard Containers."
+description: "Use the Chainguard Image Matcher API to find the closest Chainguard equivalent for an existing container image."
 type: "article"
 date: 2026-05-26T00:00:00+00:00
 lastmod: 2026-05-26T00:00:00+00:00
--- a/content/chainguard/migration/image-matcher/find-match.md
+++ b/content/chainguard/migration/image-matcher/find-match.md
@@ -1,0 +1,174 @@
+---
+title: "Find a Matching Chainguard Image Using the API"
+linktitle: "Find a Matching Image"
+description: "How to call the Chainguard Image Matcher API with an existing SBOM to find the closest Chainguard image equivalent."
+type: "article"
+date: 2026-05-26T00:00:00+00:00
+draft: false
+tags: ["Chainguard Images", "Migration", "SBOM", "API"]
+---
+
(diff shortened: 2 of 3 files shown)


matthewhelmke (approved):
This is great stuff! Let's have @wojciechka take a look as well, if for no other reason than to know his content is going to make it in and we are grateful.

matthewhelmke:
Nice work on this — content is clear, the worked example is concrete, and the structure (concept page + how-to) is what I'd expect. A few notes from a style/clarity pass against the [Google developer docs style guide](https://developers.google.com/style) and our [Style Guide wiki](https://github.com/chainguard-dev/edu/wiki/Style-Guide).

## `_index.md`

- **`description`** is the boilerplate from the parent `migration/_index.md` ("Language- or platform-specific tutorials...") and doesn't describe this section. Suggest something like: `"Use the Chainguard Image Matcher API to find the closest Chainguard equivalent for an existing container image."`
- Empty `aliases:` key — either drop the line or populate it.
- `weight: 050` is consistent with sibling section indexes, but the child pages use `01` and `02`. Pick one convention (three-digit is what the rest of `/migration/` uses).

## `find-match.md`

- **Typo:** "Learn more about the **probably score**" → "probability score" (under *Low `coverage` with a high `probabilityScore`*).
- "Before getting started, you **will need**:" → "you need:" (omit needless words).
- "looks like the following:" → "looks like this:" or just "For example:" (Strunk: definite, concrete).
- Syft is linked in the SBOM-tooling bullet but Trivy, `docker sbom`, and cdxgen aren't. Either link all four or none.
- "This narrows the candidate set and improves results, but is optional." reads more naturally as "Passing it is optional but narrows the candidate set and improves results." — or lead with "Optionally, pass..."
- The conditions list under *Reviewing multiple candidates* has verb-containing items without terminal periods. Google style: add periods, or rewrite as fragments.
- The trailing paragraph ("To request more or fewer candidates...") is glued to the bullet list without a blank line; add one so it renders as its own paragraph.

## `overview.md`

A few sentences drift into passive voice or wordy constructions where active/direct phrasing is tighter:

- "It **is designed to support** migration workflows where you know what you are running today and want to find the best Chainguard equivalent." → "It **supports** migration workflows: when you know what you run today, it finds the closest Chainguard equivalent."
- "The number of candidates returned by the API **is configurable**." → "You can configure the number of candidates the API returns."
- "It **is calculated from** three weighted components:" → "The matcher calculates it from three weighted components:"
- "This translation **is where most of the recommendation logic is concentrated**." → "Most of the recommendation logic lives in this translation step."
- "**The image catalog is refreshed daily.**" → "**Chainguard refreshes the image catalog daily.**"

Substantive question, not just style: the *Understanding the probability score* section mentions an **`overallScore`** field for tie-breaking, but the sample response in `find-match.md` doesn't show that field. Either add it to the example or clarify the relationship between `probabilityScore` and `overallScore` so readers aren't left guessing.

## Cross-cutting

- Both files have stray trailing whitespace and blank lines containing a single space. Worth a quick pass with `sed -i 's/[[:space:]]*$//'` or your editor's trim-on-save.
- Bulleted list items that contain verbs should end with periods (per Google style); right now usage is mixed across both pages — pick one and apply it consistently.
- Light anthropomorphism in a few spots ("the API expects", "the matcher's confidence") — generally fine in this domain, flagging only because Google style calls it out. I'd leave "the matcher's confidence" since it's idiomatic for ML/scoring systems.

None of this is blocking — the content reads well and the worked example is the strongest part. Happy to push a follow-up commit with the mechanical fixes (typo, trailing whitespace, list punctuation) if that's helpful.

wojciechka (approved):

PR #3358, opened 2026-05-28
Add tab functionality to docs

s-stumbo:

## Type of change
Adding content functionality to docs site

### What should this PR do?
Add the ability to use tabs within content

To make content easier to follow

The code should work as expected and not have any UI issues

In the deploy preview, go to the Libraries Overview page. Review the example tabs to ensure they work as expected. I will delete these before merging the PR; they are only here as a test


matthewhelmke (approved):
LGTM

PR #3361, opened 2026-05-29
Fix broken links

s-stumbo:

## Type of change
Fix two broken links to Kyverno docs

Fixes the 404s identified in https://github.com/chainguard-dev/edu/issues/3342


matthewhelmke (approved):
TYSM

PR #3365, opened 2026-06-01
Update Policy Gates docs page

matthewhelmke:
This changes the sample code to match the reference code in `--help` and in the [chainctl Reference](https://edu.chainguard.dev/chainguard/chainctl/chainctl-docs/chainctl_policy-gate/) pages. Instead of `policy-gates` (which is a valid alias) the page now uses `policy-gate` in the commands.

It adds information about the `--policy=` switch that was recently added.

It adds `chainctl policy-gate describe`.

It adds ` --parent=$ORGANIZATION` to match reference pages.

It separates the command from the output into two code boxes when output is included.

### DIFF
--- a/content/chainguard/administration/policy-gates/index.md
+++ b/content/chainguard/administration/policy-gates/index.md
@@ -38,37 +38,37 @@ Policy gates are managed using `chainctl`. System policies are shipped with the 
 See which policies are available to your organization:
 
 ```shell
-chainctl policy-gate list
+chainctl policy-gate list --parent=$ORGANIZATION
 ```
 
 Inspect a policy to see its full definition and configurable parameters before enabling it:
 
 ```shell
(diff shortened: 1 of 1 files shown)


matthewhelmke:
Fixes https://github.com/chainguard-dev/internal/issues/5859

angela-zhang:
@danak-cg @matthewhelmke I think this should be renamed from policy gates to policies?

matthewhelmke:
@angela-zhang Thank you. I haven't seen that come through as official yet (the reference docs and `--help` still use `policy-gate`, but I'm ready to make the change the moment I'm told it's time! @danak-cg LMK!

s-stumbo (approved):

matthewhelmke:
I'm going to merge this now and we'll update the docs when the command is changed.

PR #3372, opened 2026-06-02
Note about config change delays

s-stumbo:

## Type of change
Update to JS overview

### What should this PR do?
Add note that says fallback/cooldown changes can take 30 minutes to go into effect

Requested in slack: https://chainguard-dev.slack.com/archives/C085187D8RE/p1780362513344569?thread_ts=1780299371.513859&cid=C085187D8RE


matthewhelmke (approved):
LGTM

PR #3375, opened 2026-06-02
Rename policy gates to policies

matthewhelmke:
Depends on https://github.com/chainguard-dev/mono/pull/41398, which is merged but not yet in a build. Wait for the build. When that happens, we'll also see an automated platform docs PR that includes updates to the `chainctl` reference docs. This should land at the same time as that PR.


s-stumbo (approved):

matthewhelmke:
The automated docs PR with the changes merged: https://github.com/chainguard-dev/edu/pull/3382.
So, this is now ready to merge.

PR #3381, opened 2026-06-03
Updates across Build Config and Migration

s-stumbo:

## Type of change
Updates to Build Config and JS migration guide

### What should this PR do?
Update the framing of the pages and condense the lockfile info 

To improve clarity of pages and reduce duplicated content


matthewhelmke (approved):
LGTM

PR #3409, opened 2026-06-10
fix(how-to): Examples for adding apks in distroless fail post-install

JoeNorth:
The examples we give for adding apks in distroless environments fails due to the apk's post-install script `ldconfig-commit.sh` as a result of not having a shell in the environment to run the script. As a workaround, we run `apk add` with the `--no-scripts` flag and then run `ldconfig -r /base-chroot` to regenerate the cache.

## Type of change

Fixing example code
### What should this PR do?

Fixes errors surfaced during customer testing
ref: https://chainguardhelp.zendesk.com/agent/tickets/9823

The examples are currently incorrect and will fail when following the instructions

```
 => ERROR [build 6/6] RUN apk add --root /base-chroot so:libcurl.so.4                                                                                                                                           1.3s
------
 > [build 6/6] RUN apk add --root /base-chroot so:libcurl.so.4:
0.069 fetch https://apk.cgr.dev/chainguard/aarch64/APKINDEX.tar.gz
1.159 ERROR: ldconfig-commit.sh: script exited with error 127
------
Dockerfile.dynamic:29
--------------------
  27 |
  28 |     # Customize base image chroot
  29 | >>> RUN apk add --root /base-chroot so:libcurl.so.4
  30 |
  31 |     # Create customized production image from scratch
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c apk add --root /base-chroot so:libcurl.so.4" did not complete successfully: exit code: 99
```

Run through the build steps for both Appendix A and B to ensure they build correctly and that the updated wording makes sense.

### DIFF
--- a/content/chainguard/chainguard-images/how-to-use/install-apks-in-distroless-variants.md
+++ b/content/chainguard/chainguard-images/how-to-use/install-apks-in-distroless-variants.md
@@ -21,9 +21,9 @@ This page documents workflows for installing APK packages in [distroless variant
 
 The distroless variants of Chainguard Containers do not contain shells or package managers by design. This reduces attack surface and exploitability for these images. In cases where additional packages are required, we typically recommend the following:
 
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
Nice catch! Thank you!!

PR #3413, opened 2026-06-11
Update digest extraction

matthewhelmke:

## Type of change

Compile AI docs from GCS builds [started failing](https://github.com/chainguard-dev/edu/actions/runs/27344970712) last night. 

The job is failing at the **"Build and push container image"** step with the error:

```
unknown blob
##[error]Process completed with exit code 1.
```

### Root Cause

The issue occurs when extracting the image digest from the `docker push` output. The regex pattern on line 244 is failing to match the expected digest format:

```bash
DIGEST=$(docker push "ghcr.io/$REPO_OWNER/ai-docs:latest" | grep -oP 'sha256:[a-f0-9]+')
```

The docker push output format may have changed, or the grep pattern isn't correctly capturing the digest. When `DIGEST` is empty, the subsequent `cosign sign` command on line 252 fails with "unknown blob".

### Fix

Update the digest extraction to be more robust and handle different docker output formats:

```bash
# Build container with GitHub Container Registry
docker build -f scripts/Dockerfile.ai-docs -t "ghcr.io/$REPO_OWNER/ai-docs:latest" scripts/
docker tag "ghcr.io/$REPO_OWNER/ai-docs:latest" "ghcr.io/$REPO_OWNER/ai-docs:$COMMIT_SHA"

# Push images and capture digest for signing
PUSH_OUTPUT=$(docker push "ghcr.io/$REPO_OWNER/ai-docs:latest")
echo "$PUSH_OUTPUT"

# Extract digest more reliably - handle both formats
DIGEST=$(echo "$PUSH_OUTPUT" | grep -oP 'sha256:[a-f0-9]{64}' | tail -1)
if [ -z "$DIGEST" ]; then
  # Fallback: use inspect to get the digest
  DIGEST=$(docker inspect --format='{{index .RepoDigests 0}}' "ghcr.io/$REPO_OWNER/ai-docs:latest" | grep -oP 'sha256:[a-f0-9]{64}')
fi

if [ -z "$DIGEST" ]; then
  echo "ERROR: Failed to capture image digest from docker push"
  exit 1
fi

echo "Captured digest: $DIGEST"
docker push "ghcr.io/$REPO_OWNER/ai-docs:$COMMIT_SHA" || { echo "ERROR: Failed to push SHA-tagged image"; exit 1; }

# Sign container image by immutable digest
cosign sign --yes "ghcr.io/$REPO_OWNER/ai-docs@$DIGEST"
```

**Key improvements:**
1. **Echo the push output** for debugging visibility
2. **Refine the regex** to match exactly 64 hex characters (full SHA256)
3. **Add a fallback mechanism** using `docker inspect` to extract the digest if the grep fails
4. **Validate before signing** to ensure we have a valid digest

Co-authored with Copilot in GitHub Actions running Claude Haiku 4.5

### DIFF
--- a/.github/workflows/compile-ai-docs-from-gcs.yaml
+++ b/.github/workflows/compile-ai-docs-from-gcs.yaml
@@ -241,11 +241,22 @@ jobs:
         docker tag "ghcr.io/$REPO_OWNER/ai-docs:latest" "ghcr.io/$REPO_OWNER/ai-docs:$COMMIT_SHA"
 
         # Push images and capture digest for signing
-        DIGEST=$(docker push "ghcr.io/$REPO_OWNER/ai-docs:latest" | grep -oP 'sha256:[a-f0-9]+')
+        PUSH_OUTPUT=$(docker push "ghcr.io/$REPO_OWNER/ai-docs:latest")
+        echo "$PUSH_OUTPUT"
+
+        # Extract digest more reliably - handle both formats
+        DIGEST=$(echo "$PUSH_OUTPUT" | grep -oP 'sha256:[a-f0-9]{64}' | tail -1)
+        if [ -z "$DIGEST" ]; then
+          # Fallback: use inspect to get the digest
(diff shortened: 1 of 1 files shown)


matthewhelmke:
Successful run: https://github.com/chainguard-dev/edu/actions/runs/27347851304/job/80801294331

PR #3418, opened 2026-06-12
Add trust setup for brew install

mosabua:
### What should this PR do?

Add necessary brew command. 

Required now with newer brew versions for install and updates.

Tested. 

Just try upgrading without first.


SharpRake (approved):
Nice catch, thank you @mosabua!

PR #3445, opened 2026-06-19
libraries/javascript: fix pnpm auth for upstream-fallback packages

angela-zhang:
## What

Scopes a small fix to the **pnpm** section of the JavaScript build-configuration docs so direct-access auth works for packages served through the upstream fallback.

- Configure pnpm direct-access credentials keyed to the whole `libraries.cgr.dev` host (`//libraries.cgr.dev/:_auth`) instead of scoping them to the `/javascript/` path.
- Fix the `username`/`_password` example: pnpm requires the `_password` value to be **base64-encoded** and rejects a raw token. The previous example used the raw token and fails.

## Why

pnpm authenticates against whatever tarball URL the registry returns for each package. Packages Chainguard has not yet rebuilt are served from `https://libraries.cgr.dev/javascript-upstream/`. With credentials scoped only to `//libraries.cgr.dev/javascript/`, pnpm sends **no auth header** for the `/javascript-upstream/` path, so any install that resolves an upstream-fallback package (e.g. `react`) fails with `ERR_PNPM_FETCH_401`.

A credential keyed to the host covers both `/javascript/` and `/javascript-upstream/`. The explicit two-path form is kept in the docs as an alternative.

## Testing

Verified against **pnpm 11.8.0** with a fresh project/lockfile against `libraries.cgr.dev/javascript`:

- Host-keyed `//libraries.cgr.dev/:_auth` → `react` (upstream-fallback) resolves and downloads. ✅
- Path-scoped `//libraries.cgr.dev/javascript/:_auth` only → `ERR_PNPM_FETCH_401` on `…/javascript-upstream/react/…tgz` ("No authorization header was set"). ❌
- `username`/`_password` with a raw token fails; base64-encoded `_password` succeeds.

The **npm** section is intentionally left unchanged: npm authenticates upstream-fallback tarball fetches even with path-scoped `/javascript/` credentials, so it is already correct.

## Scope

pnpm section only (2 hunks). No changes to npm, Yarn, Bun, or repository-manager sections.

### DIFF
--- a/content/chainguard/libraries/javascript/build-configuration.md
+++ b/content/chainguard/libraries/javascript/build-configuration.md
@@ -390,13 +390,31 @@ URL to point to Chainguard and set auth credentials. The following command write
 export token=$(echo -n "${CHAINGUARD_JAVASCRIPT_IDENTITY_ID}:${CHAINGUARD_JAVASCRIPT_TOKEN}" | base64 -w 0)
 
 pnpm config set registry https://libraries.cgr.dev/javascript/ --location=project
-pnpm config set //libraries.cgr.dev/javascript/:_auth "${token}" --location=project
-pnpm config set //libraries.cgr.dev/javascript-upstream/:_auth "${token}" --location=project
+pnpm config set //libraries.cgr.dev/:_auth "${token}" --location=project
 ```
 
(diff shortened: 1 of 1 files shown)


s-stumbo (approved):

jumoel (approved):
makes sense, good change!

PR #3471, opened 2026-06-25
fix(ci): GitHub Actions security hardening

stevebeattie:
Automated GitHub Actions security hardening (zizmor + actionlint + shellcheck,
plus manual review). All changes are workflow-config only — no application or
docs content is touched.

## Changes

- **Scope `GITHUB_TOKEN` permissions per job** — replace blanket workflow-level
  permissions with `permissions: {}` at the top and the minimum scopes granted
  on the individual jobs that need them, following least privilege.

- **Set explicit `persist-credentials` on checkout steps** — `false` on the
  checkouts that do no downstream git writes (the default-safe state), and an
  explicit `true` (with a comment) on the one job that performs a bare
  `git push`, so the intent is documented rather than implicit.

- **Move template expressions into `env`** — `${{ … }}` interpolations in
  `run:`/script bodies are hoisted into quoted `env:` aliases (built-ins use the
  `$GITHUB_*` variables), removing the shell script-injection surface from
  expanding expressions directly into inline scripts.

- **Correct a pinned-action version comment** — the digestabot `setup-gitsign`
  step's version comment didn't match its pinned SHA; corrected so the pin is
  self-describing.

- **Add a zizmor config + dependabot cooldown** — a `.github/zizmor.yml`
  disabling a few cosmetic pedantic rules, plus a 3-day cooldown on each
  dependabot ecosystem so brand-new (and potentially compromised) dependency
  releases aren't pulled the instant they ship.

- **Quote shellcheck-flagged sites (SC2086) that are provably safe** —
  runner-provided redirect sinks (`>> "$GITHUB_ENV"`), single-value command
  substitutions (`"$(pwd)"`), `$VAR` in single-bracket tests, `read -r`, and
  bare single-value path/number variables used as command arguments
  (`$TEMP_BUILD_DIR` from `mktemp -d`, `$GITHUB_WORKSPACE`, the issue number in
  `gh issue edit`). Only sites where quoting cannot change behavior are touched;
  genuinely ambiguous ones (e.g. `ls`-parsing, single-quoted literals) are left
  as-is.

## Note for reviewer

This PR is from a fork and touches only workflow configuration. Several of this
repo's workflows depend on cloud/registry credentials or OIDC identity; any such
checks that report failed or skipped on a fork PR are environmental and
unrelated to these changes.

Refs: PSEC-923

### DIFF
--- a/.github/actions/integrate-platform-docs/action.yaml
+++ b/.github/actions/integrate-platform-docs/action.yaml
@@ -14,21 +14,30 @@ runs:
   steps:
     - name: download openapi api.json spec from cloud storage
       shell: bash
+      env:
+        PROJECT_ID: ${{ inputs.project_id }}
+        STORAGE_BUCKET: ${{ inputs.storage_bucket }}
       run: |
-        gcloud --quiet storage cp --project="${{ inputs.project_id }}" \
-          "${{ inputs.storage_bucket }}/enforce-openapi/api*.json" static/
+        gcloud --quiet storage cp --project="${PROJECT_ID}" \
+          "${STORAGE_BUCKET}/enforce-openapi/api*.json" static/
 
     - name: download events.md from cloud storage
       shell: bash
+      env:
+        PROJECT_ID: ${{ inputs.project_id }}
+        STORAGE_BUCKET: ${{ inputs.storage_bucket }}
       run: |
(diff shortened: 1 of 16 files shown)


matthewhelmke (approved):
LGTM, and thank you!

egibs (approved):

PR #3477, opened 2026-06-26
Clarify cooldown settings

s-stumbo:

## Type of change
Update to Libraries overview and Java build config

### What should this PR do?
- Make it clearer how to manage cooldown policies, mention error that can occur if your build fails due to the cooldown policy
- Clarify wording around how to enable upstream fallback for Java/Python 

Internal request: https://chainguard-dev.slack.com/archives/C07GQF397DH/p1782489406990609

Test the commands to create and enforce policies, and the command to view policy bindings

### DIFF
--- a/content/chainguard/libraries/overview.md
+++ b/content/chainguard/libraries/overview.md
@@ -157,7 +157,7 @@ for JavaScript, and adds the npm upstream fallback policy. Enabling upstream fal
 chainctl libraries entitlements create --ecosystems=JAVASCRIPT --policy=CHAINGUARD_AND_UPSTREAM 
 ```
 
-For JavaScript, you can also enable upstream fallback in the Chainguard Console. For Java and Python, you cannot currently enable fallback or view upstream vs. Chainguard-built packages via the Chainguard Console.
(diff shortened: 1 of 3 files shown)


angela-zhang (approved):

content/chainguard/libraries/overview.md
```diff
@@ -207,22 +207,44 @@ Use Chainguard's [malware API endpoint](/chainguard/api/spec-api-v1/#tag/malware
...
 ```
-Note: These commands require `chainctl` v0.2.291 or newer.
 
 The default cooldown period is 7 days. The cooldown period provides an additional layer of defense on top of malware and greyware scanning, giving the broader security community time to surface threats that may not be immediately detectable. 
 
+**Disable cooldown**
+
+To disable the cooldown, set it to 0. In the example below, the policy is created, then it is enforced on the Java ecosystem:
+
+```bash
```
mosabua: I would have looked at this content in the access page. I think we should move it there and then also build more details there .. the whole thing is rather complex and hard to work with imho

s-stumbo: I moved this to the Access page, and linked to it from this section

matthewhelmke (approved):

PR #3488, opened 2026-06-30
fix(chainctl): pin version when downloading and verifying chainctl

imsurajsk:
Related ZD ticket - https://chainguardhelp.zendesk.com/agent/tickets/9913

## What

Updates the [chainctl install guide](https://edu.chainguard.dev/chainguard/chainctl-usage/how-to-install-chainctl/) so that the binary, signature, and certificate are all downloaded from the **same pinned release version** instead of from `/latest/` independently.

## Why

The guide currently resolves the version and downloads the binary, signature, and certificate from `/latest/` in separate steps. If a new `chainctl` release is published between any of those downloads, you end up with a binary from one release and a signature/cert from another, and `cosign verify-blob` fails with:

```
Error: failed to verify signature: could not verify message: invalid signature when validating ASN.1 encoded signature
```

A customer running download+verify at CI volume reported hitting this intermittently (~1% of runs) and confirmed that pinning every artifact to a single resolved version eliminated the failures.

## Change

- Resolve `VERSION` once from `metadata.json`, then pin the binary download to `/${VERSION}/` (Linux/macOS `curl`, Windows PowerShell).
- Pin the `--signature` and `--certificate` URLs in the Cosign verification step to the same `/${VERSION}/`.
- Add a note explaining why pinning matters and what the symptom looks like, plus guidance to reuse the same shell session.

### DIFF
--- a/content/chainguard/chainctl-usage/how-to-install-chainctl.md
+++ b/content/chainguard/chainctl-usage/how-to-install-chainctl.md
@@ -62,8 +62,14 @@ You are now ready to use the `chainctl` command. You can verify that it works co
 A platform-agnostic approach to installing `chainctl` is through using `curl`. We have [specific instructions for Windows users](/chainguard/chainctl-usage/how-to-install-chainctl/#installing-with-curl-in-windows-powershell) on installing `chainctl` with `curl`, but all others can run the following command:
 
 ```bash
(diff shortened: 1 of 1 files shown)


manojkrgupta (approved):
Looks good to me.

matthewhelmke (approved):
LGTM, thank you!

PR #3495, opened 2026-07-01
libraries/javascript/migration: fix pnpm auth for direct access and upstream-fallback

doismellburning:
Ports the pnpm credential fixes from #3445 to the migration guide, and adds a warning that `chainctl auth configure-npm` generates path-scoped credentials that break pnpm when upstream-fallback packages are involved.

- `chainctl auth configure-npm` (and `--pull-token`) now documented as npm/Bun only; pnpm users directed to the manual configuration section.
- Manual pnpm config updated to use host-level `//libraries.cgr.dev/:_auth` (and `username`/`_password`) instead of path-scoped `//libraries.cgr.dev/javascript/`, matching the fix in #3445.
- `_password` example now base64-encodes the token; pnpm rejects a raw token.
- Short explanation added of why host-level credentials are required (upstream-fallback tarballs are served from `/javascript-upstream/`, which path-scoped credentials don't cover).

### DIFF
--- a/content/chainguard/libraries/javascript/migration.md
+++ b/content/chainguard/libraries/javascript/migration.md
@@ -170,7 +170,7 @@ credentials short-lived and avoids passing around static tokens.
 
 #### Local development
 
-For npm, pnpm, and Bun, run the following command to write a project-level `.npmrc` with the
+For npm and Bun, run the following command to write a project-level `.npmrc` with the
 registry URL and credentials from your current Chainguard session:
 
 ```shell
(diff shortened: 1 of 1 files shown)


content/chainguard/libraries/javascript/migration.md
```diff
@@ -180,7 +180,10 @@ chainctl auth configure-npm
 Because [this command](/chainguard/chainctl/chainctl-docs/chainctl_auth_configure-npm/) uses a session-backed bearer token, you will need to re-run it when the token expires. If the command returns an error, ensure you are
 using the [latest version of chainctl](/chainguard/chainctl-usage/how-to-install-chainctl/#updating-chainctl).
 
-For Yarn, [configure the .npmrc manually](#manual-registry-configuration).
+For pnpm and Yarn, [configure the .npmrc manually](#manual-registry-configuration).
```
doismellburning, line 183: I don't love the language Claude and I have used here, but I'm struggling to produce better

matthewhelmke: It's fine. The language is clear and direct, which is what matters most.

matthewhelmke (approved):
Thank you for this, @doismellburning!
LGTM

PR #3410, opened 2026-06-10
Note about pnpm support

s-stumbo:
[ ] Check if this is a typo or other quick fix and ignore the rest :)

## Type of change
Update to JS build config and overview

### What should this PR do?
Note that we have tested against pnpm v11 and recommend that version

### Why are we making this change?
Issue reported in Slack: https://chainguard-dev.slack.com/archives/C085187D8RE/p1780919639161649

### What are the acceptance criteria? 
Content should be accurate and appear in the expected location

### How should this PR be tested?

Any documentation published to Chainguard Academy is reviewed carefully for accuracy. GUI procedures, API commands, and CLI code snippets in a draft are run and tested thoroughly — by both the author and the reviewer — to confirm they work exactly as written. This helps ensure that readers can follow along and get the same results. See the [`edu` repo's README](https://github.com/chainguard-dev/edu#testing).

PR #3496, opened 2026-07-01
Note pnpm v11 support in JS migration guide

doismellburning:
Mirrors the notes added to `overview.md` and `build-configuration.md` in #3410 — upstream fallback has been tested against pnpm v11, so we recommend v11 or newer.

Adds the note in the pnpm section of the manual registry configuration block, which is where the pnpm-specific upstream fallback auth instructions live.

### DIFF
--- a/content/chainguard/libraries/javascript/migration.md
+++ b/content/chainguard/libraries/javascript/migration.md
@@ -236,6 +236,8 @@ First create a pull token as described in the [prerequisites](#pull-token), then
 
 **pnpm**
 
+> Note: The Chainguard Repository [upstream fallback](/chainguard/libraries/javascript/overview/#upstream-fallback-policy-and-controls) has been tested with pnpm v11. We recommend using pnpm v11 or newer.
+
 ```shell
 export token=$(echo -n "${CHAINGUARD_JAVASCRIPT_IDENTITY_ID}:${CHAINGUARD_JAVASCRIPT_TOKEN}" | base64 -w 0)
 


matthewhelmke (approved):
tysm!

doismellburning:
@matthewhelmke Do you think it's worth adding something Claude-y or similar to ensure the migration guide doesn't get overlooked in changes? I appreciate we may not want to duplicate everything there, and maybe I should have referred to other docs more along the way, but it felt like it should be a fairly in-depth standalone guide I could just follow and be done with

matthewhelmke:
@doismellburning it's always hard to ensure consistency across separate guides and main docs, but it's worth trying to achieve. Your instincts are correct that we want the standalone guide to be in depth enough to do everything in the guide without clicking out to other docs (but we also want to keep it focused on just the one task, so it's a balance).

Did you have a specific idea in mind for helping ensure the guide is not overlooked?

matthewhelmke:
@doismellburning I think I'll play with this idea locally, maybe have Claude run something across the local clone of the repo during commit time to check whether other changes are needed in the docs repo. Thank for the idea!!

doismellburning:
@matthewhelmke Agreed and appreciated - I'm definitely not sitting on any concrete bright ideas myself, and I definitely appreciate the tradeoffs and challenges...! Cheers :)

matthewhelmke:
@doismellburning I have a new skill file that we are going to test as a team for a bit. That takes care of part of the issue. At some point soon, I think I'll also try to create something that runs on the GitHub side when a docs PR is created, but that will take me longer. :)

This is a good idea, thank you!

PR #3514, opened 2026-07-06
Add Java verification approach for Maven

s-stumbo:

## Type of change
Update to Verification docs

### What should this PR do?
- Explain a recommended workflow for verifying Java runtime dependencies, not including everything locally cached
- This update does not include similar workflows for Gradle/Bazel - I will investigate those separately. 

Requested: https://linear.app/chainguard/issue/FEA-516/feature-request-submission

In a test Java project with dependencies:
1. Confirmed settings.xml resolves against the Chainguard Libraries repo before testing the real workflow (mvn -s settings.xml help:effective-settings in a test dir — surfaced a missing-file error until pointed at a project with a working settings.xml, which caught a config issue early rather than during the actual dependency copy).
2. Run: `mvn -U -q -s settings.xml dependency:copy-dependencies \
     -DincludeScope=runtime \
     -DoutputDirectory=target/chainguard-verify`
3. Run `ls target/chainguard-verify` to see if only runtime-scoped jars are present
4. Run `chainctl libraries verify -o json --parent=chainguard.edu --detailed target/chainguard-verify/*.jar > provenance-report.json` (replace with whatever org you want to use)
    - I first ran `chainctl libraries verify -o json --detailed target/chainguard-verify/*.jar > provenance-report.json` and it failed with `not running in an interactive terminal`. It needed the org specified.

It should return output summarizing the total artifacts and how many were verified.

### DIFF
--- a/content/chainguard/libraries/verification.md
+++ b/content/chainguard/libraries/verification.md
@@ -197,7 +197,7 @@ chainctl libraries verify -o json --detailed target/chainguard-verify/*.jar \
 
 > Note: If you belong to multiple Chainguard organizations, include the `--parent=<org>` flag in the command.
 
-This returns output similar to the following:
+It can take up to 5 minutes for this command to return results. It returns output similar to the following:
 
 ```bash
 {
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
LGTM

mosabua:
We should add a note that this can take a long time - like well over a minute, maybe even 5 or more.

PR #3523, opened 2026-07-07
chore: update CODEOWNERS and allow GitHub verified signatures in source policy

jmeridth:
### What should this PR do?
- Replace the retiring `@chainguard-dev/deved-team` in `CODEOWNERS` with `@chainguard-dev/education-team-write` and `@chainguard-dev/devrel-team-write`.
- Add a `github.verified: true` block to `.chainguard/source.yaml` so GitHub verified ssh, gpg, and smime signatures are accepted.

The deved team is being deleted, so ownership needs to move to the two teams that will carry it forward. The source policy update aligns signature verification with the `chainguard-dev/mono` source policy.

- `CODEOWNERS` no longer references `@chainguard-dev/deved-team`.
- Both replacement teams have write access and are valid owners (confirmed with owners).
- `.chainguard/source.yaml` accepts GitHub verified signatures at the `spec` level.

Confirm the two replacement team slugs are correct and resolve to teams with write access to the repository, and that the `github.verified` block sits at `spec` level (sibling of `authorities`). No content or code changes.

### DIFF
--- a/.chainguard/source.yaml
+++ b/.chainguard/source.yaml
@@ -10,3 +10,6 @@ spec:
     - key:
         # Allow commits signed by GitHub.
         kms: https://github.com/web-flow.gpg
+  # Allow Github verified ssh, gpg, and smime signatures
+  github:
+    verified: true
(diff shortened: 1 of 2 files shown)


jmeridth:
A followup PR will breakdown the CODEOWNERS file for each team so both are not a catch-all.

jmeridth:
`Enforce - commit signing` check will pass after this merges to main

cpanato (approved):

Issue #66, opened 2022-09-01
Resource | How to connect your Kubernetes clusters to Enforce for Kubernetes


nsmith5:
**What topic are you requesting a resource about?**

* Chainguard product -- Enforce for Kubernetes

**Proposed title:**

How to connect your Kubernetes clusters to Enforce for Kubernetes

**Description:**

There are three different approaches for connecting your Kubernetes clusters to Enforce for Kubernetes:

- Agentful -- Runs the Chainguard agent on your cluster
  - Without `--private` -- this is good for clusters with a publically accessible OIDC endpoint (GCP and EKS)
  - With `--private` -- good for must other clusters but some limitations
- Agentless -- We run the Chainguard agent for you ourside your cluster, but you need to set up cloud account associations (point to https://github.com/chainguard-dev/edu/issues/65 article)

This article would walk through the different options and describe when each should be used combined with a small how-to.

PR #3534, opened 2026-07-08
fix(rumble): bump grpc to v1.79.3 and drop obsolete stats/opentelemetry module

jmeridth:
## Type of change

Dependency / security fix for `tools/rumble`.

### What should this PR do?

Bump `google.golang.org/grpc` from v1.67.2 to v1.79.3 in `tools/rumble` and remove the standalone `google.golang.org/grpc/stats/opentelemetry` require. This resolves Dependabot alert #66 (CVE-2026-33186, critical gRPC authorization bypass).

Two reasons, both stemming from the same root cause:

1. **Security**: CVE-2026-33186 is a critical (CVSS 9.1) authorization bypass in gRPC-Go, patched in v1.79.3.
2. **Dependabot is stuck**: The automated `go_modules` update for `tools/rumble` keeps failing. I manually re-ran the gomod Dependabot update again and it failed again with the same error. In grpc >= 1.78 the `stats/opentelemetry` package was folded into the main grpc module, so with the bump the import path exists in **two** modules (the main module and the old standalone pseudo-versioned module), producing an ambiguous import. Dependabot reports this as `dependency_file_not_resolvable` and abandons the whole run, which also blocks `cloud.google.com/go/storage`, `google.golang.org/api`, and `cloud.google.com/go/bigquery`.

Dependabot cannot drop the obsolete standalone module on its own, so this needs a manual fix. This PR does that, which unblocks future Dependabot runs.

- `tools/rumble` builds, vets, and tests clean.
- Dependabot alert #66 is resolved once merged.
- The next Dependabot `go_modules` run for `tools/rumble` no longer fails with `dependency_file_not_resolvable`.

No published documentation is affected. This only touches the `tools/rumble` Go module. Verified locally:

```
cd tools/rumble
go build ./...   # exit 0
go vet ./...     # exit 0
go test ./...    # exit 0 (no test files, compiles and vets clean)
```

The failure was reproduced locally before the fix (`ambiguous import ... found package google.golang.org/grpc/stats/opentelemetry in multiple modules`), and confirmed resolved after dropping the standalone module and running `go mod tidy`.

---

**Risk**: low. Transitive dependency bumps in an internal tool that does not run a gRPC server, so the CVE's exploit conditions do not apply here; this is dependency hygiene plus unblocking Dependabot.

**Review focus**: confirm the dropped `google.golang.org/grpc/stats/opentelemetry` require is intended (no rumble code imports it directly) and that the transitive bumps in `go.sum` look reasonable.


matthewhelmke (approved):
I had noticed dependabot was stuck and hadn't yet had time/energy to research why. Thank you for noticing and jumping in! This is helpful.

Approved.

PR #3554, opened 2026-07-09
Updating Chainguard Agent Skills docs to fix commands, clarify public status

smythp:
Updating Chainguard Agent Skills docs to fix commands, clarify public status


smythp:
@SharpRake Pretty small fixes, should be ready to merge

SharpRake:
We've got some fancy new checks that are blocking this PR. Rather than dealing with lots of back and forth to get them to pass in this PR, I just created a new one with your same changes: https://github.com/chainguard-dev/edu/pull/3571

Thanks for jumping on this @smythp!

PR #3558, opened 2026-07-09
fix(js): resolve eslint errors in theme JavaScript

jmeridth:
## Type of change

Lint / bug fix for the theme's JavaScript.

### What should this PR do?

Clear all eslint findings under `assets/js` and `config`:
- Autofix formatting (double-quotes, comma-dangle)
- Bump `eslint.config.js` `ecmaVersion` 2018 → 2022 (the code uses optional chaining and top-level await, which 2018 can't parse)
- Fix a real bug in `titles.js`
- Behavior-neutral cleanups: `switch`-case braces and optional catch binding in `copypage.js`, an unused-variable removal in `cookie-banner-override.js`, and a `var` redeclare in `rumble/base.js`
- Add an eslint override for `assets/js/rumble/**`

`assets/js` carried the eslint backlog. Most was mechanical; a few were genuine issues that only surfaced once the parser was modernized.

- `eslint assets/js config` reports zero errors.
- No behavior regressions in the affected UI.

Verified locally in the Hugo dev server:
- **`titles.js` (real fix):** step-numbered headings referenced undefined `stepNumber`/`dash` and threw a `ReferenceError`, so the step styling never applied. Now they use the regex captures — confirmed the "Step **N** — …" badge renders (e.g. the policy-controller how-to pages).
- **Rumble tool:** the d3 comparison chart renders and the table filter works (image-comparison + vulnerability pages).
- **Cookie banner** and **copy-page-as-markdown** behave as before.
- `hugo` builds all 1563 pages.

### Notes

- **`rumble/**` override** disables `no-unused-vars` / `no-undef` for those files: Hugo concatenates `base.js` with `comparison.js` / `vulnerability.js` at build time (`resources.Concat`), so symbols defined in one fragment (`d3`, `fetchAdvisories`, `severityColours`) are used in another. eslint lints each file in isolation and can't see that, so the "unused"/"undefined" reports are false positives — the symbols are live, and deleting them would break the tool.

---

**Risk**: low-medium. Mostly mechanical/formatting plus one genuine bug fix and behavior-neutral cleanups; rendering verified locally.

**Review focus**:
1. `titles.js` regex-capture fix (renders step numbers — verified in browser).
2. The `rumble/**` eslint override vs deleting the "unused" symbols.

### DIFF
--- a/assets/js/cookie-banner-override.js
+++ b/assets/js/cookie-banner-override.js
@@ -2,36 +2,36 @@
 // This is a workaround for HubSpot's aggressive inline styles that CSS can't override
 
 (function() {
-  'use strict';
+  "use strict";
 
   // Create or update backdrop overlay
   function createBackdrop() {
-    let backdrop = document.getElementById('cookie-banner-backdrop');
+    let backdrop = document.getElementById("cookie-banner-backdrop");
     if (!backdrop) {
-      backdrop = document.createElement('div');
-      backdrop.id = 'cookie-banner-backdrop';
+      backdrop = document.createElement("div");
+      backdrop.id = "cookie-banner-backdrop";
       document.body.appendChild(backdrop);
     }
 
-    backdrop.style.setProperty('position', 'fixed', 'important');
-    backdrop.style.setProperty('top', '0', 'important');
-    backdrop.style.setProperty('left', '0', 'important');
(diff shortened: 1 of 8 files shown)


matthewhelmke (approved):
Preview looks good. I didn't find any issues.

PR #3561, opened 2026-07-09
style(markdown): lint-clean libraries docs (batch 2); disable MD036

jmeridth:
## Type of change

Markdown lint cleanup — batch 2 (`content/chainguard/libraries`).

### What should this PR do?

Make 13 `libraries` docs fully markdownlint-clean and disable `MD036` (emphasis-as-heading) in `.markdownlint-cli2.jsonc`.

- Autofixes: trailing whitespace, list style/indent, blanks around lists/headings/fences, EOF
- Manual: heading-increment (`MD001`) promotions and one list-indent (`MD005`) fix
- Disable `MD036`: it flags intentional `**bold**` labels used throughout the reference/config docs; converting the ~68 repo-wide instances to real headings would be invasive (changes structure, TOC, and heading anchors) for little benefit

Continuing the markdown lint backlog cleanup in reviewable, per-section batches.

- Every changed file is markdownlint-clean.
- No rendering changes.

- `markdownlint-cli2` reports zero errors on every changed file.
- `hugo` builds all pages; no hard line breaks lost (literal `<br>` counts match `main`, incl. `quickstart.md`).
- `MD001` fixes change only heading levels, so heading anchors are unchanged (derived from text).

### Notes / scope

- **Deferred to a follow-up** (reverted here): `java/build-configuration.md` and `javascript/migration.md`, whose only remaining errors are `MD051` link-fragments. Those need verification against Hugo's actual anchor IDs (its slugger differs from markdownlint's), so I'm handling the anchor fixes separately rather than guessing.

---

**Risk**: low. Mechanical autofix + heading-level/list-indent corrections + one rule disable; rendering verified unchanged.

**Review focus**:
1. Disabling `MD036` (keep intentional bold labels vs convert to headings).
2. The `MD001` heading-level promotions read correctly.

### DIFF
--- a/content/chainguard/libraries/java/global-configuration.md
+++ b/content/chainguard/libraries/java/global-configuration.md
@@ -40,10 +40,10 @@ optional measures:
 * Remove all cached artifacts in the proxy repository of Maven Central and other
   repositories. This step allows you to validate which libraries are not
   available from Chainguard Libraries and proceed with potential next steps with
-  Chainguard and your own development efforts. 
+  Chainguard and your own development efforts.
 * Remove any repositories that are no longer desired or necessary. Depending on
   your library requirements this step can result in removal of some proxy
-  repositories or even removal of all proxy repositories. 
+  repositories or even removal of all proxy repositories.
 
 Adopting the use of a repository manager is the recommended approach, however if
 your organization does not use a repository manager, you can still use
@@ -53,7 +53,7 @@ configure and control. Refer to the [direct access documentation for build
 tools](/chainguard/libraries/java/build-configuration/#direct-access) for more
 information.
 
-### Considerations for fallback approach
+## Considerations for fallback approach
 
 Before configuring your repo manager, consider how you want to handle packages that aren't
 yet available in the Chainguard Libraries repository. If you configure a fallback to Maven Central, packages sourced from that registry are not covered by Chainguard's
@@ -74,7 +74,7 @@ by defining multiple upstream repositories.
 ### Initial configuration
 
 Use the following steps to add a repository with the Maven Central Repository
-and the Chainguard Libraries for Java repository as Maven upstream repositories. 
+and the Chainguard Libraries for Java repository as Maven upstream repositories.
 
 Configure a `java-all` repository:
 
@@ -83,10 +83,10 @@ Configure a `java-all` repository:
 1. On the **Repositories** page, click the **+ New repository** button.
 1. Enter the name `java-all` for your new repository. The name should
    include `java` to identify the ecosystem. This convention helps
-   avoid confusion since repositories in Cloudsmith are multi-format. 
+   avoid confusion since repositories in Cloudsmith are multi-format.
 1. Select a storage region that is appropriate for your organization and
    infrastructure.
-1. Click **+Create Repository**. 
+1. Click **+Create Repository**.
 
 Configure an upstream proxy for the Maven Central Repository:
 
@@ -175,7 +175,7 @@ Before configuring the repositories, you must create a secret with the [password
 value as retrieved with chainctl](/chainguard/libraries/access/):
 
 1. Navigate to the **Secret Manager**
-1. Click **Create secret**. 
+1. Click **Create secret**.
 1. Set the **Name** to `chainguard-libraries-java`.
 1. Set the **Secret** value to the password from your `chainctl` output.
 1. Click **Create secret**.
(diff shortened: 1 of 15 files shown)


content/chainguard/libraries/javascript/global-configuration.md
```diff
@@ -405,10 +404,10 @@ If you use Chainguard Repository with upstream fallback enabled, npm clients can
...
 
 * An AWS account with the following IAM permissions:
-    * `codeartifact:PublishPackageVersion`
-    * `codeartifact:PutPackageMetadata`
-    * `codeartifact:GetAuthorizationToken`
-    * `codeartifact:GetRepositoryEndpoint`
+  * `codeartifact:PublishPackageVersion`
+  * `codeartifact:PutPackageMetadata`
+  * `codeartifact:GetAuthorizationToken`
+  * `codeartifact:GetRepositoryEndpoint`
```
matthewhelmke: @jmeridth I think these are intended to be nested, hence the indents. We should not remove them here.

matthewhelmke (dismissed):
One comment, otherwise I think this looks good.

content/chainguard/libraries/javascript/global-configuration.md
jmeridth: @matthewhelmke Good catch. You are right that the 4-space indentation is intentional. The cause was markdownlint MD007 (ul-indent) defaulting to 2-space and the autofixer normalizing to it. I have set `MD007: { indent: 4 }` in `.markdownlint-cli2.jsonc` so the rule now expects the repo's 4-space nesting, and re-ran the fix so these lists are back to 4 spaces. Addressed in https://github.com/chainguard-dev/edu/commit/20446b8c90c8d8a718209689a3dc4a791e770789

content/chainguard/libraries/java/global-configuration.md
```diff
@@ -53,7 +53,7 @@ configure and control. Refer to the [direct access documentation for build
 tools](/chainguard/libraries/java/build-configuration/#direct-access) for more
 information.
 
-### Considerations for fallback approach
+## Considerations for fallback approach
```
mosabua, line 56: How is this determined to be right or wrong?

content/chainguard/libraries/java/global-configuration.md
```diff
@@ -321,21 +321,22 @@ curl -sSf -L \
   | sha256sum
 ```
 
-2. Fetch the same artifact through the Artifactory remote repository and compute its checksum:
+1. Fetch the same artifact through the Artifactory remote repository and compute its checksum:
```
mosabua, line 324: This rule seems like a bad idea @jmeridth @s-stumbo @SharpRake @matthewhelmke

mosabua (approved):
I am approving but I am a bit worried about this causing a lot of friction for the ongoing PRs from @s-stumbo

content/chainguard/libraries/java/global-configuration.md
jmeridth: 🤖 @mosabua Good question. Two things determine it here.

First, the markdownlint MD001 (heading-increment) rule: this was the first heading in the body and it was an h3 (`###`) directly under the page title (h1), which skips h2. Heading levels are expected to increase one step at a time, so an h3 as the first heading gets flagged.

Second, it reads as a top-level section. It sits before the repository-manager sections (Cloudsmith, Google Artifact Registry, JFrog Artifactory, Sonatype Nexus Repository), which are all h2, and there is no h2 parent above it. Promoting it to h2 makes it a peer of those sections and satisfies the rule.

Changing the level does not change the anchor (it is derived from the heading text), so no in-page or inbound links break. If you would prefer it stay h3, the alternative would be to add a new h2 parent heading above it, but there is no natural parent section there, so h2 seemed the cleaner fix.

content/chainguard/libraries/java/global-configuration.md
jmeridth: 🤖 @mosabua This is the markdownlint MD029 (ordered-list prefix) rule, and it comes down to implicit vs explicit numbering.

In the source, this list was inconsistent: most items were written as `1.` (implicit, where the renderer auto-numbers them 1, 2, 3) and one was written as `2.` (explicit). MD029 default (`one_or_ordered`) accepts either style but requires the list to be internally consistent, so it normalized the stray `2.` to `1.`. Both styles render as 1, 2, 3, so this is a source-only change.

If the concern is that `1. 1. 1.` reads poorly for numbered steps, we can retune the rule rather than remove it. Three options:

1. `MD029: "ordered"`: enforce explicit `1. 2. 3.`. Re-running the fix renumbers these lists sequentially. Most readable for step-by-step docs.
2. `MD029: "one_or_ordered"` (current default): accept either style, require consistency. This is what produced the `2.` to `1.` normalization.
3. `MD029: false`: do not enforce ordered-list numbering at all.

I would lean toward option 1 so numbered steps read as 1, 2, 3 in the source, but happy to go with whatever the team prefers.

matthewhelmke: I've used 1, 1, 1 for many years in markdown.

jmeridth: @matthewhelmke same. I prefer it so I don't have to remember the order. 😄 it's preference/explicit vs implicit.

matthewhelmke (approved):

PR #3566, opened 2026-07-09
fix(actions): remove reference to a removed SHA-digest section

jmeridth:
## Type of change

Docs fix — removes a broken in-page link in `content/chainguard/actions/overview.md`.

### What should this PR do?

Remove the `[find the correct SHA digest](#find-the-sha-for-a-specific-release)` reference. That anchor pointed to a section that existed in an early draft; the Actions reviewers chose not to include it, so the link now points nowhere. This also lint-cleans the file (it was the last markdown doc with outstanding markdownlint errors, deferred from the batch cleanup).

Per Matthew Helmke: the SHA-digest section was intentionally dropped during initial review, so the leftover reference should go.

- No dangling `#find-the-sha-for-a-specific-release` link.
- The file is markdownlint-clean.
- No rendering changes beyond the removed clause.

`hugo` builds all pages with no warnings and no broken-anchor reference. The inline SHA-lookup instructions under "Replace the `uses:` line in each workflow" are unaffected — only the pointer to the removed dedicated section is gone.

---

**Risk**: low. One removed clause plus markdown autofixes; build verified.

**Review focus**: confirm dropping the SHA-digest link reference matches the Actions team's intent (per Matthew Helmke).

### DIFF
--- a/content/chainguard/actions/overview.md
+++ b/content/chainguard/actions/overview.md
@@ -89,7 +89,7 @@ Action names often have the upstream organization appended to the action name fo
 
 Search the Chainguard Actions repository, find the action you want to use, and then use the name you find there.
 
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
Thank you!

PR #3585, opened 2026-07-13
Add org-level Guardener configuration via .github repository

wlynch:
## Summary

Documents organization-level configuration for Chainguard Guardener via the special `.github` repository, so operators can define `.chainguard/` config once for the whole org instead of per repository.

- Adds an **Organization-level configuration with the `.github` repository** section covering the directory layout and setup steps.
- Adds a **How repository and organization configuration combine** subsection documenting precedence: per-repo config overrides the org-level default per feature; repos without a feature file fall back to the org default; features configured in neither place stay disabled.
- Updates the intro to note config can live per-repo or once at the org level.

## Notes

The precedence/merge behavior is described to match the standard GitHub `.github`-repo convention and the existing docs' tone. If the actual Guardener implementation does a deeper key-level merge rather than file-level override, the "How ... combine" subsection should be tightened to match.

### DIFF
--- a/content/chainguard/guardener/configuration.md
+++ b/content/chainguard/guardener/configuration.md
@@ -15,7 +15,7 @@ weight: 020
 toc: true
 ---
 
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
LGTM and cool!

PR #3593, opened 2026-07-15
docs(policies): update availability section

Rodrigo-Munoz-Salas:
Updating availability section to reflect that all customers have access to policies.

Refs: CON-2034


matthewhelmke (approved):
Thank you!

PR #3602, opened 2026-07-16
docs(actions): document Chainguard Actions telemetry and privacy

priyawadhwa:
Adds a dedicated **Telemetry and privacy** page for Chainguard Actions and links it from the overview.

## Why

A beta customer (Baseten) flagged that the hardened-action phone-home telemetry was only discoverable by reading the code, and asked for public docs on the endpoint and payload. This page answers that.

## What the page covers

- The phone-home endpoint (`actions.enforce.dev/actions/v1/record`) and that the hook is fire-and-forget / fails open.
- **What we collect, split by permission:**
  - Without `id-token: write`: repository name, timestamp, and an unverified flag.
  - With `id-token: write`: verified OIDC claims (repository, actor, ref, sha, workflow path, repository visibility, run identifiers).
- The token is audience-locked to `actions.chainguard.dev` and cannot be used elsewhere; `ACTIONS_ID_TOKEN_REQUEST_TOKEN` never leaves the runner.
- How to limit what is sent (withhold `id-token: write`).

Content mirrors the internal telemetry statement. Suggested reviewers: Actions product + PMM/enablement (Tyler Paxton, Nevin Coco, Elsie Phillips) and legal (Sara Haneef / Sean Pan) for the data-collection language.

Opened as a **draft** pending review.

### DIFF
--- a/content/chainguard/actions/overview.md
+++ b/content/chainguard/actions/overview.md
@@ -233,5 +233,6 @@ If an action isn't working as expected, [open an issue](https://github.com/chain
 
 ## Learn more
 
+- [Chainguard Actions telemetry and privacy](/chainguard/actions/telemetry/)
 - [Chainguard Actions product page](https://www.chainguard.dev/actions)
 - For other questions, [contact Chainguard](https://www.chainguard.dev/contact?utm=docs).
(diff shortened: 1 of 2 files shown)


wlynch (approved):

SharpRake (approved):
thank you @priyawadhwa!

PR #3434, opened 2026-06-17
docs: add Custom Assembly custom runtime repositories documentation

tomgeorge:
## Summary

- Documents the new custom runtime repositories feature for Custom Assembly, which allows customers to replace default `virtualapk.cgr.dev` entries in `/etc/apk/repositories` with their own internal mirror URLs
- Adds overview section in `custom-assembly.md` covering the feature, compatibility limitations (Artifactory remote repos work; virtual repos and Nexus do not due to signing keys), and a note about future keyring support
- Adds `chainctl` usage section in `custom-assembly-chainctl.md` with YAML config examples, declarative `build apply` workflow, and the full set of URL validation rules (HTTPS-only, no private/loopback/link-local IPs, no internal hostnames)

## References

- Linear: [CON-1759](https://linear.app/chainguard/issue/CON-1759/document-custom-apk-repositories-on-educhainguarddev)
- Project: [Custom Assembly - Custom APK Repositories](https://linear.app/chainguard/project/custom-assembly-custom-apk-repositories-456a3a1e770c/overview)
- Design doc: https://docs.google.com/document/d/1esti3mf8PHl-GWlGK69K0QheJhZNQUdUl3KEm7mw_H8
- Platform PRs: chainguard-dev/mono#41940, chainguard-dev/mono#42101

## Test plan

- [ ] Verify Hugo renders both pages without errors
- [ ] Review YAML examples match the actual `chainctl` behavior
- [ ] Confirm validation rules match the implementation in mono PRs
- [ ] Check all internal links resolve correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3635, opened 2026-07-23
Document Custom Assembly custom runtime keys

tomgeorge:
## Summary

Follow-up to #3434 (custom runtime repositories). Documents the custom runtime keyring feature shipped in chainguard-dev/mono#43731, chainguard-dev/mono#43749, chainguard-dev/mono#46359, and chainguard-dev/mono#45294: customers can embed APK signing public keys into `/etc/apk/keys` so `apk` trusts packages from mirrors that re-sign them.

### `custom-assembly-chainctl.md`
- New **Custom runtime keys** section under Custom runtime repositories: the `--with-runtime-keyring` flag, the `contents.runtime_keyring` YAML field (`name` + PEM `content`), key naming against the APKINDEX signature filename (`.SIGN.RSA256.<name>`), SHA-256 digests in confirmation diffs, editor keep/replace/delete semantics, and removal.
- **Key validation rules** list: single PEM RSA public key per entry, unique non-empty names, no leading dots or path separators, reserved `chainguard`/`wolfi` prefixes, 512 KB total cap.

### `custom-assembly.md`
- Adds a **Custom runtime keys** feature bullet and concept section linking to the chainctl guide.
- Rewrites **Limitations and compatibility**: re-signing mirrors (Artifactory virtual repositories, Sonatype Nexus Alpine repositories) were previously documented as not compatible with a future-release caveat; they now work by adding the mirror's public signing key. Drops the `--allow-untrusted`/Dockerfile workarounds.

The new content sits under the existing beta-flagged Custom runtime repositories section, matching the shared feature gate on the platform side.

Refs: CON-1887

### DIFF
--- a/content/chainguard/chainguard-images/features/ca-docs/custom-assembly-chainctl.md
+++ b/content/chainguard/chainguard-images/features/ca-docs/custom-assembly-chainctl.md
@@ -264,6 +264,49 @@ Custom runtime repository URLs are validated when the configuration is applied. 
 * `localhost` isn't allowed.
 * Hostnames that resolve to any of the blocked IP ranges listed earlier aren't allowed either.
 
+### Custom runtime keys
+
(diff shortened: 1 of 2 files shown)


matthewhelmke (approved):
LGTM, thank you!

PR #3643, opened 2026-07-24
Rename --with-runtime-keyring flag to --with-runtime-keys

tomgeorge:
Follow-up to #3635. The chainctl flag was renamed from `--with-runtime-keyring` to `--with-runtime-keys` in chainguard-dev/mono#45294 before merging; this updates the two references in the Custom runtime keys section. The YAML field name (`contents.runtime_keyring`) is unchanged.

Refs: CON-1887


matthewhelmke (approved):
TYSM!

PR #3646, opened 2026-07-24
Explain more auth configure-docker details

max-allan-cgr:

This misunderstanding caused a customer bug report.

### DIFF
--- a/content/platform/chainctl-usage/how-to-install-chainctl.md
+++ b/content/platform/chainctl-usage/how-to-install-chainctl.md
@@ -222,7 +222,7 @@ This will:
 * Update your Docker config file to call `docker-credential-cgr` when an auth token is needed. A browser window will open when the token needs to be refreshed.
 
 {{< note >}}
-If you choose to use `sudo` when creating the `chainctl` symlink, and then complete login in that same `sudo` session, the Docker credentials will be written to `root`'s `.docker/config.json` instead of your user's Docker config. In that case, run `configure-docker` again outside of `sudo` so the credentials are also available to your user account.
+If you installed `chainctl` to a directory that needs root permissions and run as root (eg with `sudo`) when creating the `docker-credential-cgr` symlink, and then complete login in that same session, the Docker credentials will be written to `root`'s `.docker/config.json` instead of your user's Docker config. In that case, run `configure-docker` again as the correct user.
 {{< /note >}}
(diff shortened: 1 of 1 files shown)


max-allan-cgr:
NB I removed the reference to pull tokens because explaining pull tokens doesn't belong in a section about doing the exact opposite of using a pull token!

Maybe someone else would like to add a section on creating pull tokens.

content/platform/chainctl-usage/how-to-install-chainctl.md
```diff
@@ -216,9 +216,14 @@ You can configure a Docker credential helper with chainctl by running:
...
-This will update your Docker config file to call chainctl when an auth token is needed. A browser window will open when the token needs to be refreshed.
+This will:
 
-For guidance on pull tokens, please review [authenticating with a pull token](/chainguard/chainguard-registry/authenticating/#authenticating-with-a-pull-token).
+* Create a link for `docker-credential-cgr` in the same directory as `chainctl`.
+* Update your Docker config file to call `docker-credential-cgr` when an auth token is needed. A browser window will open when the token needs to be refreshed.
+
+{{< note >}}
+If you installed `chainctl` to a directory that needs root permissions and run as root (eg with `sudo`) when creating the `docker-credential-cgr` symlink, and then complete login in that same session, the Docker credentials will be written to `root`'s `.docker/config.json` instead of your user's Docker config. In that case, run `configure-docker` again as the correct user.
+{{< /note >}}
```
matthewhelmke, line 226:
```suggestion
This will:

* Create a link for `docker-credential-cgr` in the same directory where `chainctl` is installed.
* Update your Docker config file to call `docker-credential-cgr` when an auth token is needed. A browser window will open when the token needs to be refreshed.

{{< note >}}
If you installed `chainctl` to a directory that needs root permissions and you then create the symlink for `docker-credential-cgr` while running as root (for example, with `sudo`) , and then complete your login in that same root session, the Docker credentials will be written to the root-owned `.docker/config.json` instead of your user's Docker config.

To correct this, run `configure-docker` again as the correct user.
{{< /note >}}
```

matthewhelmke (approved):
This still isn't perfect, but I think it's better. That's quite the chain of conditions to follow. I'm approving, because I can't think of a better way to say this yet.

PR #3647, opened 2026-07-24
Revert "Revise major change reporting process for contributors"

matthewhelmke:
Reverts chainguard-dev/edu#3644

This is a public repo. Linear is an internal tool. Reverting to invite issue filing here in the public repo.


mosabua (approved):
Seems fine with me. It might be worth mentioning that we are tracking internally.

PR #3663, opened 2026-07-29
add note to solve DOCS-87

matthewhelmke:

## Type of change
Adds a new entry in the **Be aware that the following are not covered by Chainguard's EOL grace period:** list

> * Installing updated versions of packages in instances where the new package version no longer supports the EOL version of the image's main package (see note).

And a new note with further clarification

> **Installing updated versions of packages:** Chainguard won't remediate CVEs by installing updated versions of packages in instances where the new package version no longer supports the EOL version of the image's main package. If we did, this would cause test failures and a broken image forcing an end of support for that image, so we're sticking with the old version of the package to avoid that and give you a little longer to migrate.

You can attempt to use `apk upgrade` at you own risk to try to update those packages.

### What should this PR do?
Fixes https://linear.app/chainguard/issue/DOCS-87/clarification-on-eol-grace-period

Addresses customer confusion

Is the wording clear and does it follow the style guide (Docs Team) and is the new content accurate (Max)?

### DIFF
--- a/content/chainguard/chainguard-images/features/eol-gp-overview/index.md
+++ b/content/chainguard/chainguard-images/features/eol-gp-overview/index.md
@@ -4,7 +4,7 @@ linktitle: "EOL Grace Period"
 type: "article"
 description: "Understanding Chainguard's end-of-life (EOL) grace period."
 date: 2025-05-14T08:49:31+00:00
-lastmod: 2024-06-11T08:49:31+00:00
+lastmod: 2026-07-29T12:53:45+00:00
 draft: false
 tags: ["Chainguard Containers"]
 images: []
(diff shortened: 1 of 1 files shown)


max-allan-cgr (approved):

content/chainguard/chainguard-images/features/eol-gp-overview/index.md
```diff
@@ -46,10 +46,15 @@ Be aware that the following are not covered by Chainguard's EOL grace period:
 
 * Updating an image’s EOL primary package.
 * Backporting or cherry-picking individual commits or patches to the EOL primary package.
+* Installing updated versions of packages in instances where the new package version no longer supports the EOL version of the image's main package (see note).
 * Any package labeled end-of-life for more than 6 months by its open-source creators or maintainers.
 
 Additionally, if a container image fails to build because underlying dependencies conflict with the primary package, it will no longer be supported. A failed build signals the end of support for that image.
 
+{{< note >}}
+**Installing updated versions of packages:** Chainguard won't remediate CVEs by installing updated versions of packages in instances where the new package version no longer supports the EOL version of the image's main package. If we did, this would cause test failures and a broken image forcing an end of support for that image, so we're sticking with the old version of the package to avoid that and give you a little longer to migrate. You can attempt to use `apk upgrade` at your own risk to try to update those packages.
```
s-stumbo: I might slightly reword. Something like: When an image's primary package reaches EOL, newer versions of some dependency packages may no longer support it. In these cases, Chainguard does not update those packages solely to remediate CVEs, because doing so could introduce compatibility issues or result in a broken image. Instead, the existing package versions are retained to preserve image functionality while providing additional time to migrate to a supported image. You can attempt to install newer package versions at your own risk, using `apk upgrade`.

s-stumbo (approved):
Thanks for adding this! I left one suggestion.

content/chainguard/chainguard-images/features/eol-gp-overview/index.md
matthewhelmke: Oh, I like that! Will update.

PR #3668, opened 2026-07-29
docs(policies): align Container Pull Policies page with shipped CLI

konradzapalowicz:
Match the docs to current chainctl behavior: state the policy composition rule, document decision-list paging, stop promising a decision reason the CLI does not surface, and remove the Console mention this page does not cover.


matthewhelmke (approved):
LGTM, thank you

PR #3682, opened 2026-07-31
Move library and container policy pages

s-stumbo:

## Type of change
Move policy content for libraries and containers

### What should this PR do?
Moves policy content to appear under the Chainguard Repository section

We have gotten internal feedback that the policy content is hard to find. This change brings both the library and container policy content together in one section and makes it more visible

Content should appear as expected


matthewhelmke (approved):
I see an alias for the container policies page. Does anything else need one?
Outside of that question to consider, LGTM and approved.

PR #3692, opened 2026-08-03
update list of container policies and add menu entry

matthewhelmke:

## Type of change

* Adds a missing container policy to the list, `support-window`
* Adds a crosslink menu entry
* 

Fixes requests made in slack by the container policies PM

The additional content should be clearly written and the new menu entry for `Container Pull Policies` should appear under `Products > Chainguard Containers > Staying Secure`

### DIFF
--- a/content/chainguard/chainguard-images/staying-secure/_index.md
+++ b/content/chainguard/chainguard-images/staying-secure/_index.md
@@ -6,7 +6,7 @@ aliases:
 description: "Best practices for container security with Chainguard: vulnerability management, SBOM verification, supply chain protection, and enterprise-grade hardening techniques"
 type: "article"
 date: 2024-12-19T08:49:15+00:00
-lastmod: 2026-08-03T12:43:21+00:00
+lastmod: 2026-08-03T13:37:02+00:00
 draft: false
 images: []
 weight: 035
--- a/content/chainguard/chainguard-images/staying-secure/_index.md
+++ b/content/chainguard/chainguard-images/staying-secure/_index.md
@@ -6,8 +6,11 @@ aliases:
 description: "Best practices for container security with Chainguard: vulnerability management, SBOM verification, supply chain protection, and enterprise-grade hardening techniques"
 type: "article"
 date: 2024-12-19T08:49:15+00:00
-lastmod: 2024-12-19T08:49:15+00:00
+lastmod: 2026-08-03T12:43:21+00:00
 draft: false
 images: []
 weight: 035
+crosslinks:
+- title: "Container Pull Policies"
+  url: "/chainguard/chainguard-repository/container-policies/"
 ---
--- a/content/chainguard/chainguard-images/_index.md
+++ b/content/chainguard/chainguard-images/_index.md
@@ -4,10 +4,13 @@ linktitle: "Chainguard Containers"
 description: "Chainguard provides the most secure container images with zero known CVEs, minimal attack surface, SBOMs, and daily updates - the enterprise choice for container security"
 type: "article"
 date: 2022-09-05T08:49:15+00:00
-lastmod: 2025-07-10T08:49:15+00:00
+lastmod: 2026-08-03T12:54:49+00:00
 draft: false
 images: []
 weight: 020
+crosslinks:
+- title: "Container Pull Policies"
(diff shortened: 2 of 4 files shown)


content/chainguard/chainguard-images/staying-secure/_index.md
```diff
@@ -6,8 +6,11 @@ aliases:
...
 type: "article"
 date: 2024-12-19T08:49:15+00:00
-lastmod: 2024-12-19T08:49:15+00:00
+lastmod: 2026-08-03T12:43:21+00:00
 draft: false
 images: []
 weight: 035
+crosslinks:
+- title: "Container Pull Policies"
+  url: "/chainguard/chainguard-repository/container-policies/"
```
matthewhelmke, line 15: @SharpRake, this does not appear in the sidebar in the preview. It didn't render locally or here.

However, if I cut/paste these three lines into the index one level higher (`/chainguard/chainguard-images/_index.html`) the new link appears as expected in my local build. Any ideas? I'll add an edited version of that file alongside this one so we can see if it renders here in the netlify preview.

matthewhelmke: And, the netlify build has been stuck now for 20 minutes since I did that. So... `¯\_(ツ)_/¯`

matthewhelmke: I pushed a trivial commit to try to trigger a new netlify build since the last one was stuck for about an hour.

matthewhelmke: Screenshot_2026-08-03_08-40-52

Working in parent, not in child. Hmm.

matthewhelmke: I think I've got this figured out. I'm going to update the PR with a bit of code that should allow the new menu entry deeper in the hierarchy. Then we can discuss whether that's okay with the team before we merge. I'll have to update the PR either way to remove one of the added entries, so if I need to remove the code, no biggie.

matthewhelmke: @SharpRake — traced why the crosslink wasn't showing under Staying Secure. The partial only iterates .Params.crosslinks at two depths in [the code](https://github.com/chainguard-dev/edu/blob/main/layouts/partials/sidebar/auto-collapsible-menu.html#L121-L127): the top-level section (line 141) and the first-level node (line 122). A second-level subsection like chainguard-images/staying-secure/ is never checked, so its crosslinks frontmatter is read but silently dropped.

This commit mirrors your existing subsection block one level deeper, inside the level-2 node's child list. Same iteration, same markup and classes — just at the next nesting level. It renders correctly in local preview; checking the Netlify preview next.

Open question for you and @s-stumbo is should crosslinks work at arbitrary depth, or is two levels the intended ceiling? We can stop where it already was and I'll remove what I added for testing here. We can keep what I added and have it available at three levels. Now, if we want it deeper still, the hand-copied nesting is getting repetitive and might be worth pulling into a recursive partial rather than adding another near-identical block each level, but that may be deeper than we care about.

In any case, I would like to suggest we keep the third level I added here for testing so that I can put the link where the page used to be, but I'm open to further discussion.

matthewhelmke: It renders correctly in the PR preview (but still has the addition in two places, so once we decide I need to remove one).

SharpRake: IMO they should work everywhere, though I wonder how often we should use them. I could see it getting confusing over time

SharpRake (approved):
lgtm!

matthewhelmke:
I removed the extra crosslink, so once I know @s-stumbo is okay with us enabling the one-deeper level menu option, this is ready to merge.

content/chainguard/chainguard-images/staying-secure/_index.md
s-stumbo: I agree that they should work everywhere

matthewhelmke: This PR just adds that third level. I'll create a separate PR for "everywhere" because it's a little more involved.

PR #3696, opened 2026-08-03
Recursive sidebar partial: cross-links at any depth

matthewhelmke:
## What

Refactors the sidebar menu into a single recursive partial so cross-links work at any depth in the content hierarchy.

Previously the sidebar hard-coded three near-identical nesting levels, so a `crosslinks` entry defined below the third level was read by Hugo but never rendered. This came out of the discussion on #3692, where the docs team decided cross-links should work at any level rather than a fixed depth.

## How

- New `layouts/partials/sidebar/menu-item.html` renders a page as either a collapsible node or a leaf link, and calls itself for each child — so the menu and its cross-links work at arbitrary depth.
- `auto-collapsible-menu.html` keeps the section frame and section-level cross-links; the ~125-line triplicated block becomes a ~17-line loop.

## Behavior preserved

- The two styling tiers — a section's direct children ("top") versus everything nested below — render exactly as before.
- The top-level-only `Videos` guard is unchanged.
- Existing pages (levels 1–3) render the same markup.

## One deliberate change beyond the depth fix

Bootstrap collapse ids now key on `md5(RelPermalink)` instead of `md5(LinkTitle)`. Two subsections that share a title previously collided — expanding one toggled the other — and recursion made that more likely, so this closes it. No JS or SCSS depends on the old id scheme.

## Verification

- Full `hugo` build is clean (no template errors).
- Existing cross-links render unchanged.
- A temporary cross-link placed on a level-3 section rendered correctly at the previously-impossible depth (reverted before commit).

---
Created in collaboration with Claude Code running Claude Opus 4.8 on 2026-08-03.

### DIFF
--- a/content/chainguard/chainguard-images/staying-secure/_index.md
+++ b/content/chainguard/chainguard-images/staying-secure/_index.md
@@ -6,7 +6,7 @@ aliases:
 description: "Best practices for container security with Chainguard: vulnerability management, SBOM verification, supply chain protection, and enterprise-grade hardening techniques"
 type: "article"
 date: 2024-12-19T08:49:15+00:00
-lastmod: 2026-08-03T13:37:02+00:00
+lastmod: 2026-08-03T12:43:21+00:00
 draft: false
 images: []
 weight: 035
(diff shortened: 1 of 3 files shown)


SharpRake (approved):
Thank you!

PR #3751, opened 2026-08-11
Rename Chainguard Images docs path to /chainguard/containers/

SharpRake:
The product is Chainguard Containers; the URL still said `chainguard-images`. This moves `content/chainguard/chainguard-images` to `content/chainguard/containers` and rewrites the 616 internal links that pointed at the old path.

## Redirects go through Hugo aliases, not nginx

These were never two independent systems. `layouts/index.aliases` renders every page's `aliases:` frontmatter into nginx map syntax, which lands in `public/_aliases`, which `nginx.conf:46` pulls in with `include aliases;`. Both end up as regex entries in the same `map` block, and nginx tests them in declaration order — so a Hugo alias always beats a hand-written rule matching the same URL.

Aliases won on three counts:

- **They work on deploy previews.** Netlify never reads `nginx.conf`. An nginx catch-all would 404 every old URL on every PR preview.
- **Destinations self-heal.** The template emits `$p.RelPermalink`, not a literal, so targets follow pages that move again later.
- **Touching nginx destinations is what broke things.** An earlier attempt repointed them and broke 8 URLs that had been surviving on a second hop. Leaving those rules alone lets the alias layer fix them automatically.

## Why every legacy alias has a `/containers/` twin

An alias generates `"~^<path>(.+)?$"` with `$1` appended, so it acts as a subtree redirect. A section alias therefore swallows everything beneath it and can outrank a deeper page alias depending on Hugo's iteration order. The twin makes the result correct either way — one hop if the specific alias wins, two if the greedy one does.

Moving the pages also activated 32 aliases Hugo had been suppressing. The guard at `layouts/index.aliases:3` drops an alias equal to its own page's permalink; once the pages moved, those old-path aliases stopped matching and began emitting.

## The three nginx changes

Two URLs cannot be expressed as aliases, because a section-index alias would shadow every rule beneath it. The bare section root and the `getting-started` index use rules anchored with `/?$`, matching only themselves. Separately, `/vulnerabilities` is served by an internal `rewrite` to a file path, which no alias can express.

The `reference/*/getting-started-*` rules now point at their real destinations instead of stale URLs that a second alias had to rescue, cutting them from two hops to one.

## Verification

`main` was built in a worktree and served alongside this branch, then 592 legacy URLs were swept against both:

| | |
|---|---|
| Regressions | **0** |
| Already-broken URLs now fixed | 90 |
| Pre-existing failures, untouched | 41 |
| Resolve in 1 hop / 2 hops / 3+ | 576 / 16 / 0 |

All 155 canonical pages return 200, all 301 legacy asset URLs still resolve, `nginx -t` passes, and markdownlint is clean.

## Notes for review

- `/chainguard/chainguard-images/getting-started/istio/` still redirects to the image directory, matching `main`. The local `istio.md` is real content but unreachable on `main` too. Delete `nginx.conf:82` if that page should be live.
- The menu `identifier` stays `chainguard-images`. It is an internal key referenced by `parent:` in 32 files and never rendered; renaming it would mean 33 more edits with a chance of silently dropping pages from the sidebar.
- `ai-metadata.json` counts pages by a `section` param no page sets, so its subcategory counts render 0 — including Libraries and VMs, which this PR does not touch. Pre-existing, left alone.
- The second commit fixes `README-PACKAGE-MAPPINGS.md` (its filesystem paths pointed at a directory that no longer exists) and `static/downloads/chainguard-complete-docs.md`, which is a committed artifact read by `compile-ai-docs-from-gcs.yaml` rather than a build output, so it does not pick up the rename on its own.

The 41 pre-existing broken redirects are mostly `/chainguard/administration/*` → `/platform/*` and are worth a separate pass.

## Why two lint configs changed

pre-commit runs on changed files only, so files nobody had edited had never been checked. The rename touched several for the first time and they failed on violations that already existed on `main`. Verified by linting each file at `main` and on this branch — identical results, so the rename introduced none of them.

Fixed in place, since these are hand-maintained:

- `layouts/shortcodes/blurb/why_ca.html` and `layouts/index.aimetadata.json` did not end in a newline.
- `docs/PACKAGE_MAPPINGS_UPDATE.md` had 7 blank-line violations (autofixed, no prose changed).

Excluded, since it is generated and any fix is discarded on the next run — `static/downloads/chainguard-complete-docs.md`, built by `scripts/compile_docs.py`:

- **markdownlint** — inherits ~35k style violations from the pages it bundles.
- **trailing-whitespace** — same inheritance.
- **detect-private-key** — bundles `cosign-manual-way.md`, whose example PEM key is already excluded on its own. `.github/gitleaks.toml` already allowlists this file for the same reason.

Each exclusion sits alongside the existing generated-file exclusions (`chainctl-docs`, `static/api*.json`) and follows the precedent in `c3bd2dc63`.


matthewhelmke (approved):
TYSM!
I affirm the decisions made as well. Good calls.

PR #3754, opened 2026-08-11
Remove old and unused contributors info

matthewhelmke:

## Type of change

In content/contributors we have a series of outdated pages with short bio type pages for four people who have contributed content in the past. These pages are not used anymore anywhere. Let's remove them and the section.

### What should this PR do?

Fixes https://linear.app/chainguard/issue/DOCS-101/remove-old-contributors-pages-from-edu-repo

We have not used these pages in years, the info was stale, and all of the content here is attributed to Chainguard in general, not individual creators.

Are the right things removed?

Why I say that: In my local environment, the `content/contributors` directory and individual contributors directories for each of these files were also removed. I always have to re-learn how Git deals with empty directories, but I think those will just disappear here, too.

Check whether the correct things were removed and nothing else. There should be no other changes in this PR.


SharpRake (approved):
Thank you!

PR #3759, opened 2026-08-12
Override semver to patched 7.x under hugo-installer (DOCS-107)

matthewhelmke:
## What

Adds a scoped npm `overrides` entry that forces `semver` to a patched 7.x (`^7.5.2`) **within `hugo-installer` only**, clearing the last dev-scope Dependabot finding tracked in DOCS-107 ([GHSA-c2qf-rxjj-qqgw](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw), high-severity semver ReDoS).

## Why the ticket's original plan didn't apply

DOCS-107 was written assuming `hugo-installer` needed a breaking bump to 2.0.0. It doesn't — DOCS-103 already moved us to `hugo-installer@^4.0` (currently `4.0.1`, the latest published release). But that bump **did not clear the vuln**: `hugo-installer@4.0.1` still hard-pins `semver: "7.3.x"`, resolving to `semver@7.3.7`, which sits inside the vulnerable range (`>=7.0.0 <7.5.2`). There is no upstream `hugo-installer` fix, so a version bump can't resolve it — an override is the only path.

## Why scoped, not global

A second `semver` exists in the tree — `execa/node_modules/semver@5.7.2` — which is **not** vulnerable (5.x is outside the affected range). A global `semver` override would needlessly drag execa from 5.x to 7.x and risk an API break for zero security benefit. The scoped override patches exactly the vulnerable node and leaves execa alone.

A `//overrides` comment key in `package.json` documents this for future maintainers.

## Verification

| Check | Result |
|---|---|
| `postinstall` Hugo download | Hugo 0.152.1 extended, unchanged |
| hugo-installer's semver | now `7.8.5` (patched) |
| execa's semver (untouched) | still `5.7.2`, non-vulnerable |
| `npm audit` | **0 vulnerabilities** (was 2 high) |
| `npm run build` | 1681 pages, builds clean |

`hugo-installer`'s own version and postinstall behavior are unchanged, so regression risk is low. The Netlify deploy preview on this PR is the final build-environment confirmation.

Resolves DOCS-107.

---
Created in collaboration with Claude Code running Claude Opus 4.8 on 2026-08-12.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4732,21 +4732,6 @@
         "node": ">=20"
       }
     },
-    "node_modules/hugo-installer/node_modules/semver": {
-      "version": "7.3.7",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.3.7.tgz",
-      "integrity": "sha512-QlYTucUYOews+WeEujDoEGziz4K6c47V/Bd+LjSSYcA94p+DmINdf7ncaUinThfvZyu13lN9OY1XDxt8C0Tw0g==",
-      "dev": true,
-      "dependencies": {
-        "lru-cache": "^6.0.0"
-      },
-      "bin": {
-        "semver": "bin/semver.js"
-      },
-      "engines": {
-        "node": ">=10"
-      }
-    },
     "node_modules/hugo-installer/node_modules/strip-dirs": {
       "version": "3.0.0",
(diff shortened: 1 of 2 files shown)


SharpRake (approved):
:+1:

PR #3760, opened 2026-08-12
Removing 2 internal images that were addded to the Changelog

SharpRake:
[ x ] Check if this is a typo or other quick fix and ignore the rest :)


matthewhelmke (approved):
TY

PR #3763, opened 2026-08-12
Add Microsoft Entra ID SCIM provisioning guide

eslerm:
New procedural tutorial for provisioning users into Chainguard from Microsoft Entra ID over SCIM 2.0, and linking each provisioned user to their SSO login by the Entra `oid` claim. Companion to the existing Entra ID SSO guide.

Validated end to end against a live Entra tenant, including the current provisioning UX (Connectivity / Attribute mapping / Provisioning pages).

Before un-drafting:
- [ ] Confirm the production SCIM base-URL shape/host that `scim token generate` prints (doc keeps it a placeholder; observed only in dev).


eslerm:
Superseded by #3781.

PR #3766, opened 2026-08-13
docs: Remove --blocked-license flags from Libraries policy reference

doismellburning:
🤖 The AutoDocs run in #3626 published the `--blocked-license`, `--remove-blocked-license` and `--replace-blocked-licenses` flags onto the `chainctl libraries policy create` and `update` reference pages. The feature is not generally available yet, so the reference should not document it.

Only the license flags and their examples are removed. The block/allow list changes that landed in the same AutoDocs run are shipped and are left alone.

The Options tables are re-rendered using the generator's own column-padding rule for the reduced flag set rather than hand-trimmed, so the result matches what a future AutoDocs run will produce. `npm run build` passes.

A follow-up change upstream of AutoDocs is needed for this to persist past the next run.

### DIFF
--- a/content/platform/chainctl/chainctl-docs/chainctl_libraries_policy_create.md
+++ b/content/platform/chainctl/chainctl-docs/chainctl_libraries_policy_create.md
@@ -20,15 +20,8 @@ Create a CUSTOM Libraries policy for an organization.
 A policy configures the gates applied when your organization pulls upstream
 packages. Use --cooldown-days to quarantine newly published versions for N
 days (0 disables the cooldown, 1-30 sets an explicit window, omit to inherit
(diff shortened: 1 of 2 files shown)


matthewhelmke (approved):
Thanks for the PR when you noticed the issue!

matthewhelmke:
@doismellburning keep me posted via Slack if you need me to hold off merging any new AutoDocs PRs over the next couple of days. It's easy enough for me to gate temporarily for a short time.

doismellburning:
@matthewhelmke Thank you! I think I've figured out how to exclude this from future runs - will let you know...! 😅

doismellburning:
Also kudos to Andrew Dean who found it!

PR #3781, opened 2026-08-14
Add SCIM provisioning docs: concept page plus Okta and Entra ID guides

eslerm:
Customer docs for SCIM user provisioning early access (DOCS-130, CUS-1081), structured as general documentation plus per-IdP onboarding:

- **Provision Chainguard Users with SCIM** (`custom-idps/scim-provisioning/`) — the provider-neutral page: how provisioning behaves (records vs. accounts, externalId matching, deactivation semantics), common prerequisites including the two-directly-assigned-owners requirement, token lifecycle, enable/disable, and limits.
- **Okta SCIM provisioning** (`idp-providers/okta-scim/`) — Okta procedure only: token, a dedicated app integration to carry provisioning (SCIM can't attach to the OIDC login app), enablement, verify, troubleshooting.
- **Microsoft Entra ID SCIM provisioning** (`idp-providers/ms-entra-id-scim/`) — carried over from #3763, which this PR supersedes, with small fixes: the two-owner prerequisite, token replacement pointed at `regenerate` (a second `generate` is not the replacement path), and links into the concept page.

All three carry a beta banner, and group provisioning is explicitly scoped out (leave Okta group push off; disable the Entra ID groups mapping) — EA is `/Users` only.

The `chainctl` commands and platform behavior are verified against source: token expiry defaults and flags, rotation overlap bounds, revoke semantics, the two-owner enable gate, rate-limit responses, and the request body cap.

Two things to weigh during review:

1. Banner wording — `beta` shortcode with `feature="SCIM user provisioning"`; tune or drop as house style dictates.
2. The Okta Admin Console steps come from our draft doc and need a live verification pass. In particular the guide states the HTTP Header value is sent exactly as typed and must be `Bearer <token>` — the opposite of Entra ID's bare-token field — so that's the claim most worth confirming against a real Okta org.


eslerm:
Follow-up, not for this PR: the existing Entra SSO guide creates the identity provider [without a correlation rule](https://github.com/chainguard-dev/edu/blob/56041a76b75ba42ad89f35198ffa73d56e0b9841/content/platform/administration/custom-idps/idp-providers/ms-entra-id/index.md#L111-L134), and that setting is [immutable](https://github.com/chainguard-dev/edu/blob/6aa8cec1ca68515d61f14500ea4cde3dba00c42d/content/platform/administration/custom-idps/idp-providers/ms-entra-id-scim/index.md#L85) — anyone who follows the SSO guide first and later wants SCIM has to delete and recreate their identity provider. A short pointer in the SSO guide ("if you plan to use SCIM provisioning, follow the SCIM guide instead — the correlation rule can't be added later") would prevent that.

SharpRake (approved):

PR #3791, opened 2026-08-17
docs(agent-skills): publish Public Skills MCP page

priyawadhwa:
## What

The [Public Skills MCP page](https://github.com/chainguard-dev/edu/blob/main/content/chainguard/agent-skills/public-skills-mcp.md) was merged with `draft: true` in its frontmatter, so Hugo skips it in the production build and it never renders on the public site.

This flips it to `draft: false` so it publishes under the Agent Skills menu, matching its sibling pages (`public-registry.md`, `skills-registry.md`) which are all `draft: false`.

## Why

The page is complete and merged but invisible to users. No other frontmatter changes are needed — `menu`, `parent`, `weight`, and `type` already match the siblings.


matthewhelmke (approved):

matthewhelmke:
Thank you, @priyawadhwa! That was my failure and I appreciate you finding it.

PR #3793, opened 2026-08-17
Note Entra group Object IDs in the Entra SCIM guide

eslerm:
Adds a short section to the Microsoft Entra ID SCIM guide on group-to-role mapping, with the Entra-specific detail that Entra emits group Object IDs (GUIDs), not display names, so a mapping shows a GUID in `chainctl` and the console's group-mapping tab. Covers how to identify the group by its Object ID, or configure Entra to emit display names instead.

Cross-links the existing Grant Chainguard roles from identity provider groups guide rather than duplicating it.

Context: surfaced during eslerm-cus SCIM/GBRM console testing.


SharpRake (approved):
thank you!

PR #3800, opened 2026-08-19
Make Chainguard the only library source in the JFrog instructions, and apply consistent Advanced settings

andrewd-cg:
## Type of change

Documentation correction and restructure, covering the JFrog Artifactory sections of the Python, JavaScript, and Java Libraries pages.

### What should this PR do?

**1. Make Chainguard the only library source.** Each JFrog section now opens by stating the recommended setup: Chainguard as the single upstream, with the Chainguard Repository's own [upstream fallback](/chainguard/libraries/overview/#upstream-fallback-and-controls) serving anything Chainguard does not build. Readers are told to disable or remove any existing Artifactory remote pointing at Maven Central, PyPI, or the public npm registry, and to remove it from the virtual repository their builds resolve against.

All self-managed fallback guidance is gone from the three JFrog sections, along with the fallback repositories it introduced into the virtual repository steps. The validation sections previously noted that Artifactory silently falls back to the public registry when a remote fails; that warning is retained but reframed to support the new guidance instead of assuming a fallback remote exists.

**2. Apply consistent Advanced settings.** The three pages previously prescribed different subsets of the same Advanced tab settings:

| Setting | Python | JavaScript | Java |
| --- | --- | --- | --- |
| Disable URL Normalization | documented | documented | documented |
| Block Mismatching Mime Types | uncheck | not mentioned | uncheck |
| Bypass HEAD Requests | not mentioned | check | not mentioned |
| Lenient Host Authentication | not mentioned | uncheck | not mentioned |
| Cookie Management | not mentioned | optional | not mentioned |

All three now document the same five settings, in the same order, with the same explanation. The JavaScript page's standalone **Advanced settings for redirect handling** section is folded into **Initial configuration**, so readers apply the settings while creating the repository rather than after hitting corruption. The cache-clearing remediation moves to the validation section, where a reader discovers they need it.

**3. Java section cleanup.**

* The remediated-libraries setup moves out of the initial configuration list into its own **Remediated libraries** section beside the virtual repository steps. It was one step buried in a list, which made the following "Combine the repositories" instruction read as though it always applied.
* The JFrog and Nexus intros said "Chainguard Java Libraries" — the only two uses in the Libraries docs, against 91 uses of "Chainguard Libraries for Java". Both also ended with a period while introducing a numbered list.
* Removed the "Use this setup for initial testing" paragraph and the JFrog Curation bypass note.

The validation troubleshooting bullets now point back to the configuration steps instead of restating the settings.

The Advanced settings divergence read as though each ecosystem needs different handling. It does not — all three are served by the same backend and hit the same redirect path. I verified this against `libraries.cgr.dev`:

* **Redirects are not npm-specific.** Artifact requests return `302` to a pre-signed Cloudflare R2 URL in all three ecosystems. `.whl`, `.jar`, `.pom`, `.sha1`, and `.tgz` all redirect.
* **The redirect body is `text/html`** in every ecosystem, so an artifact fetch presents `text/html` on the first hop. That is what makes **Block Mismatching Mime Types** risky, and it applies to npm as much as to the other two.
* **Credential forwarding breaks the fetch.** Replaying the pre-signed URL with the `Authorization` header forwarded returns `400`; without it, `200`. **Lenient Host Authentication** matters in all three ecosystems, though it was only documented for JavaScript.
* **URL mutation breaks the signature.** Stripping the query returns `400`, removing `X-Amz-Signature` returns `400`, dropping one parameter returns `403`, and adding a trailing slash returns `403`. This is what **Disable URL Normalization** protects against.

The one genuine per-ecosystem difference is HEAD handling:

| | GET | HEAD |
| --- | --- | --- |
| npm `.tgz` | 302, `text/html`, 478 bytes | **200, `application/octet-stream`, 4187 bytes** |
| Maven `.jar` | 302, `text/html` | 302, `text/html` |
| PyPI `.whl` | 302, `text/html` | 302, `text/html` |

For npm, Artifactory's pre-cache HEAD probe is told the artifact is a 4187-byte octet-stream, then the GET returns a redirect — which is how a redirect response ends up cached as the tarball and fails npm's integrity check. Python and Java answer both verbs identically, so enabling **Bypass HEAD Requests** there is safe and removes a redundant round trip.

* No JFrog section describes configuring your own fallback to a public registry.
* Each JFrog section states that Chainguard should be the only source and that existing public remotes should be disabled or removed.
* The three pages document the same Advanced settings with consistent wording.
* The JavaScript page has no separate advanced-settings section.
* No broken in-page anchors.

1. `npm install && npm run build` — builds clean (1623 pages; only pre-existing Dart Sass deprecation warnings).
2. I verified programmatically that no in-page link on any of the three pages is broken, including the two JavaScript links that previously pointed at the now-removed `#advanced-settings-for-redirect-handling`. The JFrog **Initial configuration** anchors are `initial-configuration-1` on JavaScript and `initial-configuration-2` on Python and Java.
3. The commands in the validation sections are unchanged, but the redirect and header behaviour they depend on was re-tested against `libraries.cgr.dev` as described above.

### Notes for the reviewer

Two judgement calls worth checking:

* I set **Block Mismatching Mime Types** to unchecked for npm, matching Python and Java. It was previously checked only because it is Artifactory's default and the JavaScript page never mentioned it. Unchecking removes a guard against caching an HTML error page, with the documented checksum validation as the compensating control. If the team prefers keeping it checked, the alternative is adding `text/html` to **Override Default Blocked Mime Types** on all three instead — the two approaches should not be mixed.
* The page-level **Manually managing fallback** sections are left in place, since they are not JFrog-specific and already advise against self-managed fallback. The Cloudsmith, Google Artifact Registry, and Nexus sections still describe the manual pattern; aligning those would be a larger change and is not attempted here.

### DIFF
--- a/content/chainguard/libraries/java/global-configuration.md
+++ b/content/chainguard/libraries/java/global-configuration.md
@@ -4,7 +4,7 @@ linktitle: "Global configuration"
 description: "Configuring Chainguard Libraries for Java in your organization"
 type: "article"
 date: 2025-03-25T08:04:00+00:00
-lastmod: 2026-06-24T14:42:00+00:00
+lastmod: 2026-08-20T15:07:13+00:00
 draft: false
 tags: ["Chainguard Libraries", "Java"]
 images: []
@@ -51,6 +51,10 @@ manager. If your organization does not use a repository manager, you can pull
 directly from Chainguard. Refer to the [direct access documentation for build
 tools](/chainguard/libraries/java/build-configuration/#direct-access) for more
 information.
+
+## Configure access with a repository manager
+
+See the Build configuration docs for instructions on configuring repository manager access for [Maven](/chainguard/libraries/java/build-configuration/#repo-manager-maven), [Gradle](/chainguard/libraries/java/build-configuration/#repo-manager-gradle), and [Bazel](/chainguard/libraries/java/build-configuration/#repo-manager-bazel).
 
 ## Manually managing fallback
 
@@ -275,19 +279,13 @@ for proxying and hosting, and virtual repositories to combine them. Refer to the
 Artifactory](https://docs.jfrog.com/artifactory/docs/maven-repositories)
 for more information.
 
-Configure Chainguard as the only source of Java libraries in Artifactory, and
-rely on the Chainguard Repository's [upstream
-fallback](/chainguard/libraries/overview/#upstream-fallback-and-controls) to
-serve anything Chainguard does not build. Chainguard retrieves those packages
-from Maven Central on your behalf, applying malware scanning and the cooldown
-policy on the way through.
-
-This means you should disable or remove any existing Artifactory remote
-repository that points at Maven Central, and remove it from the virtual
-repository your builds resolve against. A remote pointing directly at Maven
-Central bypasses those protections, and because Artifactory resolves through the
-virtual repository in order, a misconfiguration can result in Artifactory serving
-an unprotected package.
+If you follow the recommended approach to rely on Chainguard Repository's
+upstream fallback, disable or remove any existing Artifactory remote repository
+that points at Maven Central, and remove it from the virtual repository your
+builds resolve against. A remote pointing directly at Maven Central bypasses
+those protections. Since Artifactory resolves through the virtual repository in
+order, a misconfiguration can result in Artifactory serving an unprotected
+package.
 
 ### Initial configuration
 
@@ -325,6 +323,10 @@ pre-signed URL, forward your credentials across the redirect, or cache the
 redirect response in place of the POM, JAR, or checksum file. A cached redirect
 response fails checksum verification at build time.
 
+If you are manually managing fallback, you can configure an additional remote
+repository for Maven Central with lower priority. Make sure to deactivate
+**Maven Settings - Handle Snapshots** in the remote repository.
+
 Create a virtual repository to give your build tools a single access point:
 
 1. Click **Create a Repository** and choose the **Virtual** option.
@@ -335,7 +337,7 @@ Create a virtual repository to give your build tools a single access point:
 1. Add the `java-chainguard` repository.
 1. Click **Create Virtual Repository**.
 
-### Remediated libraries
+### Add remediated libraries
 
 This section applies only if you use the separate repository of [remediated
 Java libraries](/chainguard/libraries/cve-remediation/). It requires a second
@@ -355,7 +357,7 @@ that remediated versions resolve first.
 
 ### Validate the remote repository
 
-After creating the `java-chainguard` remote repository, validate that Artifactory is successfully proxying through to Chainguard before proceeding. A misconfigured remote repository fails quietly: if any remote pointing at Maven Central is still present, Artifactory resolves through it instead and the build succeeds with no visible error, having pulled an unprotected package. This is the main reason to remove those remotes.
+After creating the `java-chainguard` remote repository, validate that Artifactory is successfully proxying through to Chainguard before proceeding. A misconfigured remote repository fails silently; if any remote pointing at Maven Central is still present, Artifactory resolves through it instead and the build succeeds with no visible error. This can result in pulling an unprotected package.
 
 Common sources of misconfiguration include invalid or expired credentials, or an incorrect or incomplete repository URL. The Artifactory **Test** button on the repository configuration screen is not a reliable indicator; it may fail for a correctly configured repository, and may pass for an incorrectly configured one. Instead, use the following steps to verify that fetching an artifact through Artifactory produces the same checksum as fetching it directly from `libraries.cgr.dev`.
 
@@ -435,11 +437,14 @@ Libraries for Java repository for production use.
 
 ### Initial configuration
 
-Use the following steps to add Chainguard Libraries for Java as a proxy repository:
+Use the following steps to add Chainguard Libraries for Java as a proxy repository.
 
-If you are configuring your own fallback in your repo manager, for initial testing it is advised to create a separate proxy
-repository for the Maven Central Repository, a separate proxy repository
-Chainguard Libraries for Java repository, and a separate repository group.
+The recommended approach is to use the Chainguard Repository's [upstream
+fallback](/chainguard/libraries/overview/#upstream-fallback-and-controls). If
+you are instead configuring your own fallback in your repo manager, for initial
+testing it is advised to create a separate proxy repository for the Maven
+Central Repository, a separate proxy repository Chainguard Libraries for Java
+repository, and a separate repository group.
 
 1. Log in as a user with administrator privileges.
 1. Click the gear icon in the top navigation bar to access **Server administration**.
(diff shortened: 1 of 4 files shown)


content/chainguard/libraries/java/global-configuration.md
```diff
@@ -293,31 +307,55 @@ Configure a remote repository for the Chainguard Libraries for Java repository:
...
-1. Add the `java-chainguard` repository. If you are using the remediated repository, add the `java-chainguard-remediated` repository and ensure it is the first in the displayed list.
+1. Add the `java-chainguard` repository.
 1. Click **Create Virtual Repository**.
 
-Use this setup for initial testing with Chainguard Libraries for Java. For
-production usage add the `java-chainguard` repository to your production virtual
-repository.
+### Remediated libraries
+
+Skip this section unless you are using the separate repository with [remediated
```
s-stumbo:
```suggestion
This section applies only if you use the separate repository of [remediated
```

content/chainguard/libraries/java/global-configuration.md
```diff
@@ -293,31 +307,55 @@ Configure a remote repository for the Chainguard Libraries for Java repository:
     * **User Name** and **Password / Access Token**: Set to the [values as retrieved with chainctl](/chainguard/libraries/access/).
     * Deactivate **Maven Settings - Handle Snapshots**.
 1. Optionally click **Test** to verify connection and authentication.
```
s-stumbo:
```suggestion
1. Note: The **Test** button is not a reliable indicator; to verify your setup, see the [validation steps](#validate-the-remote-repository) later on this page.
```

content/chainguard/libraries/javascript/global-configuration.md
```diff
@@ -199,52 +213,40 @@ repository:
 1. Set the **URL** to `https://libraries.cgr.dev/javascript/`.
 1. Set **User Name** and **Password / Access Token** to the [values as retrieved
    with chainctl](/chainguard/libraries/access/).
-1. Click the **Advanced** configuration tab, then check the box next to **Disable URL Normalization**.
+1. Optionally click **Test** to verify connection and authentication.
```
s-stumbo:
```suggestion
Note: The **Test** button is not a reliable indicator; to verify your setup, see the [validation steps](#validate-the-remote-repository) later on this page.
```

content/chainguard/libraries/python/global-configuration.md
```diff
@@ -247,30 +260,51 @@ Configure a remote repository for the Chainguard Libraries for Python index:
...
 1. Select *PyPI* as the **Package type**.
 1. Set the **Repository Key** to `python-chainguard`.
-1. Set the **URL** to `https://libraries.cgr.dev/`. Do not include `/python` in the URL.
+1. Set the **URL** to `https://libraries.cgr.dev/`. Do not include `/python` in
+   the URL. Two things make this necessary. Python's [Simple Repository
+   API](https://peps.python.org/pep-0503/) keeps the package index on its own
+   path, so Artifactory's PyPI repository type takes that location separately, in
+   the two **PyPI Settings** fields below, rather than from the base URL. The base
+   URL also has to cover the `/python-upstream/` paths that the index advertises
+   for packages served through upstream fallback, which sit outside `/python/`.
```
s-stumbo:
```suggestion
   the URL. Python's [Simple Repository
   API](https://peps.python.org/pep-0503/) keeps the package index on its own
   path which goes in the **PyPI Settings** fields below, and the base
   URL also needs to cover the `/python-upstream/` paths for upstream fallback packages.
```

mosabua (changes requested):
I agree in principle with the change, but I strongly disagree with doing this change only for Artifactory. The docs for each repo manager should be consistent in our approach, so if we do this change to remove upstream fallback in the default instructions for Artifactory, we must also do that for Nexus, Cloudsmith, and whatever else we have documented.

The other aspect is that we do not have separate URLs for upstream and from source built artifacts and all repo managers have policy engines that are often used. This makes it difficult to adopt and switch to Chainguard when different artifacts with different trust and metadata are served, so organizations might correctly choose NOT to use both, so we need to make sure we keep the docs for upstream use around as well. 

cc @angela-zhang @katie-crabtree @justinbyo

andrewd-cg:
> I agree in principle with the change, but I strongly disagree with doing this change only for Artifactory. The docs for each repo manager should be consistent in our approach, so if we do this change to remove upstream fallback in the default instructions for Artifactory, we must also do that for Nexus, Cloudsmith, and whatever else we have documented.

Agree that those also should be updated, but selfishly I have a POV in flight that complained to me about the inconsistent docs in regards to JFrog setup, so my goal is to get these updated to help me move on with the POV.

Maybe you could open a separate PR for the other Artifact managers so this one doesn't get blocked while that happens @mosabua (Point claude at this PR and ask it to draft one for you based on the same changes and then update from there?)

mosabua:
> > I agree in principle with the change, but I strongly disagree with doing this change only for Artifactory. The docs for each repo manager should be consistent in our approach, so if we do this change to remove upstream fallback in the default instructions for Artifactory, we must also do that for Nexus, Cloudsmith, and whatever else we have documented.
> 
> Agree that those also should be updated, but selfishly I have a POV in flight that complained to me about the inconsistent docs in regards to JFrog setup, so my goal is to get these updated to help me move on with the POV.
> 
> Maybe you could open a separate PR for the other Artifact managers so this one doesn't get blocked while that happens @mosabua (Point claude at this PR and ask it to draft one for you based on the same changes and then update from there?)

I have to leave the decision on how to message this in the docs to the docs team (@s-stumbo ) and the PMs (@angela-zhang @katie-crabtree @justinbyo ). Maybe its fine to get this PR ready and then do a fast follow PR as you suggest. Unfortunately I wont be able to allocate time for that.

s-stumbo:
Thanks for the work on this and reviews @andrewd-cg @mosabua. I just made some updates to this PR to add back in the notes on how to set up manual fallback, but in each mention I make it clear that it is not the recommended approach. I updated the wording in the Artifactory section intros to mention the recommended approach again, and added a mention of it to the Sonatype intros. The global config page intros also describe the recommended approach.

justinbyo:
I'm comfortable with how @s-stumbo framed the upstream fallback recommendation

mosabua (approved):
Looks good to me as long as we also make sure the other repo managers get the consistent messaging in a follow-up PR.

s-stumbo (approved):

andrewd-cg:
@s-stumbo thanks for updating and merging. One minor point is that the java version doesn't have the 'this is not recommended' for using the virtual grouping with fallback to public from Jfrog.... could we add something in the Java page to make it consistent that it's not recommended to manually manage fallback in Jfrog etc?


* Python "rather than using the recommended Chainguard Repository built-in fallback approach"                                  
* JavaScript "rather than following the recommended approach to use Chainguard's upstream fallback"                                
* Java "If you are manually managing fallback, you can configure an additional remote repository for Maven Central with lower priority."

PR #3832, opened 2026-08-24
Containers networking requirements

xnox:
We have now deployed PQC X25519MLKEM768 universally on all our
endpoints.

Document that X25519MLKEM768 is now sufficient to communicate with all
our services. This enables customers to only use PQC-safe cryptography
if they so choose.

Also update all settings to follow the IANA standard names as listed on https://www.iana.org/assignments/tls-parameters


SharpRake (approved):
Looks good to me, thank you @xnox!

PR #3836, opened 2026-08-25
docs(container-policies): document input.main_package_license and input.packages fields

Rodrigo-Munoz-Salas:
Add two new PolicyInput fields to the input-fields table and the FAQ bullet list. Both fields are declared in the SDK (mono PR [#52482](https://github.com/chainguard-dev/mono/pull/52482)) but not yet populated at runtime, so every mention carries an explicit warning that custom policies referencing them today evaluate against empty values.

Ref: CON-2558


matthewhelmke:
Holding this until I hear from @Rodrigo-Munoz-Salas that the other PR has merged. Then I'll review/approve and we can merge this.

matthewhelmke (approved):
Thank you! I'll leave this to @Rodrigo-Munoz-Salas to merge so he can time it to go live when the code merges.

PR #3838, opened 2026-08-25
Add a workflow to open the weekly changelog rotation issue

SharpRake:
Someone on the Docs team updates the [Containers changelog](https://edu.chainguard.dev/chainguard/changelog/) by hand every Monday, but nothing tracks whose turn it is. This adds a workflow that opens a reminder issue each Monday and assigns it to whoever is on shift.

The workflow only opens the issue. It does not touch the changelog page.

## What it does

Every Monday it works out the current owner, checks whether an issue for that week is already open, and if not opens one assigned to them and labeled `automated`. The body links the source spreadsheet and points at the `update-changelog` skill.

`workflow_dispatch` is enabled too. A manual run on any weekday names the same owner the schedule would, and exits without opening a duplicate if the week's issue already exists.

## Three things worth a reviewer's attention

**The roster and cadence.** `OWNERS = ["SharpRake", "matthewhelmke", "s-stumbo"]`, four weeks each. All three have write access on this repo, so the assignment will stick.

**`EPOCH` is backdated on purpose.** It anchors the rotation and must be a Monday. Setting it to 2026-08-10, three weeks before the first scheduled run, starts the roster partway through the first shift so I take a single week and Matthew picks up a full four:

| Week of | Owner |
| --- | --- |
| 2026-08-31 | SharpRake |
| 2026-09-07 through 09-28 | matthewhelmke |
| 2026-10-05 through 10-26 | s-stumbo |
| 2026-11-02 through 11-23 | SharpRake |

A date in the past that matches no particular event reads like a typo, so there's a comment above it explaining the offset. Moving `EPOCH` reshuffles who is up when.

**The cron drifts an hour in winter.** GitHub accepts UTC only and does not follow DST, so `7 17 * * 1` is 10:07 PT now and 09:07 PT once the clocks go back. Both land after the source spreadsheet's Monday refresh (06:00 to 06:30 ET) and before Monday triage, and I picked the side that drifts earlier for that reason. The odd minute is deliberate: runs scheduled on the hour queue behind everyone else's and can be delayed well past it.

## Testing

- `actionlint` v1.7.12 with `shellcheck --severity=error --exclude=SC2129`, matching the `SHELLCHECK_OPTS` in `actionlint.yaml` — clean, and clean across every workflow in the repo rather than this file alone.
- `zizmor` v1.29.0 against `.github/zizmor.yml` — no findings.
- The rota arithmetic executed directly over the next 15 Mondays to produce the table above.

Nothing runs until this merges. The job is guarded by `if: github.repository == 'chainguard-dev/edu'`, and scheduled workflows only fire from the default branch. The first issue would open on Monday 2026-08-31.

### DIFF
--- a/.github/workflows/changelog-rotation.yaml
+++ b/.github/workflows/changelog-rotation.yaml
@@ -1,0 +1,133 @@
+# Copyright 2026 Chainguard, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# Opens a weekly issue reminding the current owner to update the Containers
+# changelog, and rotates that owner every four weeks.
+#
+# To change the roster or the cadence, edit OWNERS, EPOCH, or SHIFT_WEEKS in the
+# "Work out whose shift it is" step. Read the comments there first: EPOCH anchors
+# the whole rotation, so moving it reshuffles who is up when.
+
+name: changelog rotation
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Mondays at 17:07 UTC: 10:07 PT in summer, 09:07 PT once the clocks go
+    # back. GitHub only accepts UTC, so the local time drifts by an hour twice a
+    # year. Either way it lands after the source spreadsheet's Monday refresh
+    # (06:00-06:30 ET) and before the triage meeting. The odd minute is
+    # deliberate; runs scheduled on the hour queue behind everyone else's.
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
LGTM, thanks for setting this up!

Note: I'm approving joyfully, but did not test the code, I just read it.

PR #3851, opened 2026-08-27
Clarify FIPS usage with helm charts

max-allan-cgr:

I've tried to tighten up the language with a more explicit "set the image in the values" rather than "adjust the charts" (the chart is not adjusted, the values are. And it doesn't say what to adjust)

Removed the comment about building a single version for non-iamguarded because we clearly have 2 versions now!

And made it clearer after a user complained about the elasticsearch-fips image not working with elasticsearch-iamguarded chart, that they can't swap them out.


matthewhelmke (approved):
Thank you!

PR #3854, opened 2026-08-27
Move container migration docs under Chainguard Containers

SharpRake:
## Type of change

**Documentation Update**
- Move the container migration docs from `/get-started/migration/` to `/chainguard/containers/migration/`, with Hugo aliases on every page
- Merge the two duplicate Guardener pages into one
- Rebuild `/get-started/migration/` as a thin cross-product router covering Containers, Libraries, the API, and Guardener
- Trim `migrations-overview.md` and repoint the Containers landing banner at the migration section

### What should this PR do?

resolves https://linear.app/chainguard/issue/DOCS-117/migration-docs-cleanup-reorg-fixes

Move the container-specific migration docs under Chainguard Containers, and turn the Get Started migration page into a product-neutral router that serves Libraries and API readers as well.

`content/get-started/migration/` was the only Get Started subsection that owned deep content — its siblings are thin hubs. All 22 of its files were container-specific despite the product-neutral name, so readers arriving for Libraries found nothing relevant. Guardener was documented twice across two sections, with each page pointing at the other and neither a superset of the other.

- Every legacy migration URL redirects to its new home — 73 were verified resolving 200 through nginx against the built site
- `/get-started/migration/` stays a live page (it is never aliased) and routes to all four products
- The Guardener merge drops no content from either source page
- Migration appears under Chainguard Containers between About and Features

The `/chainguard/migration/` section root uses a hand-written `nginx.conf` rule anchored with `/?$` rather than a Hugo alias. A generated alias is a greedy prefix that would misroute the flat legacy URLs whose targets moved into subfolders (`alpine-compatibility` into `compatibility/`, `migrating-python` into `migration-guides/`) and would swallow `/chainguard/migration-guides/` as well. This is a deliberate exception to the alias-first convention and is worth a reviewer's eye.

1. Check the preview link and ensure the changes look right
2. Visit `/get-started/migration/` and confirm the hub links out to Containers, Libraries, API, and Guardener
3. Spot-check a few legacy URLs — `/chainguard/migration/`, `/chainguard/migration/alpine-compatibility/`, `/chainguard/migration-guides/migrating-python/` — and confirm each redirects rather than 404s
4. Open `/chainguard/containers/` and confirm the "Migrate to Chainguard Containers" banner now lands on the migration section


matthewhelmke (approved):
Another step in the right direction! The site IA keeps getting better, tysm!!

PR #3863, opened 2026-08-28
containers: document tarballs.cgr.dev as a required endpoint (OS-490)

8none1:
Closes [OS-490](https://linear.app/chainguard/issue/OS-490/document-tarballscgrdev-as-a-required-endpoint-in-network-requirements)

> [!WARNING]
> **This change was written by an AI agent (Claude) working through the Q3 OS PMAI backlog, and has not been checked by a human yet.** The wording in particular is a judgement call about what we tell customers, so please read it properly rather than waving it through.

## What and why

INC-137 follow-up. `tarballs.cgr.dev` serves the upstream source archives that Chainguard SBOMs point at via the SPDX `downloadLocation` field, but it was not listed in the containers network requirements table, so customers had no way to know they needed egress to it.

Adds the table row, plus a short note that this host is unlike the others on the page: it is only needed by tooling that follows those SBOM URLs (source-provenance checks, license and compliance scanners, air-gapped source mirroring). Container and package pulls do not touch it.

## Where the claim comes from

`mono/build/internal/sbom/spdx/upstream.go` constructs the URL when a package's upstream has no stable downloadable archive:

```go
const tarballMirror = "https://tarballs.cgr.dev"
...
downloadLocation = fmt.Sprintf("%s/%s/%s-%s.tar.gz", tarballMirror, name[:1], sha256Hex(name), ref)
```

## Please check

- **Is the distinction worth drawing at all?** I called it out as "not needed for pulls" so people do not over-scope their firewall rules. The opposite argument is that a caveat invites customers to skip it and then hit the exact INC-137 confusion later. Happy to drop the paragraph and just add the row.
- Marked `v4` only, matching the other `*.cgr.dev` rows. I did not verify whether the backend bucket has AAAA records.
- Should the same row go on the Libraries network requirements page? I only touched the Containers one, since the tarball mirror is populated from APK SBOMs.

### DIFF
--- a/content/chainguard/containers/network-requirements.md
+++ b/content/chainguard/containers/network-requirements.md
@@ -10,7 +10,7 @@ lead: "Using Chainguard Containers with firewalls, access control lists, and pro
 type: "article"
 description: "Using Chainguard Containers with firewalls, access control lists, and proxies."
 date: 2023-09-08T08:49:31+00:00
-lastmod: 2026-08-25T13:24:30+00:00
+lastmod: 2026-08-28T00:00:00+00:00
 draft: false
 tags: ["Chainguard Containers", "Reference"]
 images: []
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
LGTM, thank you!

Checking the v6 and libraries items before merging.

matthewhelmke:
Confirmed no AAAA record for `tarballs.cgr.dev` (A record only, `34.107.162.32`), so the `v4` marking is correct.

On the open question about Libraries: no equivalent row is needed. `tarballs.cgr.dev` is populated only by the APK push-event handler, which uploads source archives for packages whose SPDX `downloadLocation` points at that host — an APK/container path with no Libraries codepath behind it. Libraries handles the same "no stable upstream archive" case by serving fallback tarballs from `libraries.cgr.dev` itself (e.g. `https://libraries.cgr.dev/javascript-upstream/tar-fs/-/tar-fs-2.1.4.tgz`), and that host is already listed on the Libraries network requirements page. So this change is correctly scoped to Containers.

PR #3867, opened 2026-08-28
Updating docs to reflect setup for Entra and add info for App Roles

ewmcleod:
Adding in instructions for customers looking to use Entra App Roles as the mapping is different than Entra Groups. Also adding in a clarifying note that when using entra they should not be setting a scope as this will cause failures.


SharpRake (approved):
Thank you @ewmcleod!

PR #3884, opened 2026-09-01
Updating Python Remediated instructions for JFrog Setup

ewmcleod:
Customer reported that our Python Remediated JFrog instructions are incorrect.
I can confirm that we update the normal ones but did not update the Remediated ones.


matthewhelmke (approved):
Thank you!

PR #3885, opened 2026-09-01
docs: add a Get support page

matthewhelmke:
Adds `/get-started/get-support/`. edu had no page about how to get help from Chainguard — no file under `content/` matched `*support*`, `*ticket*`, or `*help*`.

In July's Kapa data, "Submitting Support Tickets" ran 8 threads from 6 unique users. Kapa cited no edu page in 7 of the 8 and hedged or deflected in 7 of 8, the worst ratio of any cluster in the dataset, falling back to two Zendesk articles instead. This gives Kapa and site search an edu page to answer from, and gives customers the portal's prerequisites before they hit an error trying to reach it.

## What the page covers

- **Where to send your request.** The support portal; your own identity provider administrator for SSO sign-in problems; `support@chainguard.dev` when you can't sign in at all; community resources for Catalog Starter, which doesn't include ticket access; and your account team for subscription questions.
- **Before you open a ticket.** Site search, Ask AI, and the knowledge base.
- **Open a ticket.** The two prerequisites — a paid plan, and an identity linked to an organization — then the steps, and what to do when the portal still returns an error.
- **What to include in your request.**

The page documents the process for contacting support rather than indexing individual issues. DOCS-156 listed recurring cases as candidate content (2FA recovery, organization rename, domain change, SLA reporting); those are support-owned procedures that would go stale on a docs page. Two examples survive, both because they disambiguate a routing boundary rather than document a procedure.

## Also in this PR

- Adds "unsubscribe" wording to the *Manage email notifications* procedure and routes readers without the `owner` role to an owner, then to the new page. The Kapa data included three consecutive follow-ups on a single unresolved unsubscribe ticket.
- Adds a "Get help" entry to the Get started landing page.

No menu configuration was needed. The sidebar entry comes from the section weight.

## Verification

`npm run build` completes clean, with only the pre-existing Sass deprecation warnings. Checked in the local dev server: the page renders, the `note` shortcode renders, the TOC builds, and every internal link resolves in `public/`.

## Notes for reviewers

- The page says click **Ask AI**, which matches the widget's actual button label on this site (`data-button-text="Ask AI"` in `layouts/partials/head/script-header.html`). The support announcement article calls it "Ask Linky AI." If support prefers the branded name, that's a deliberate change rather than a correction.
- The knowledge base article on resetting MFA is access-restricted, so the page states the SSO boundary in its own words rather than linking to it.
- The account team routing line is the least-sourced statement on the page and is the one most worth a support review.

[...]

PR #3886, opened 2026-09-01
docs: point support links at Get support, and welcome attachments

matthewhelmke:
Two follow-ups to [#3885](https://github.com/chainguard-dev/edu/pull/3885).

## DOCS-166: route generic "contact support" links through Get support

Pages across the repo sent readers straight to `support.chainguard.dev` without mentioning what they need in place first. An identity that isn't linked to an organization produces an error at the portal, which [Get support](https://edu.chainguard.dev/get-started/get-support/) now explains.

The links fall into two kinds, and they get different treatment:

**Resource lists** — "Need help?" style lists where the reader is looking for somewhere to go. These now link Get support, because that is the orientation page.

- `content/developer-resources.md`
- `content/mcp-server-ai-docs.md`
- `content/chainguard/containers/features/ca-docs/custom-assembly-gitops.md`
- `content/platform/api/api-v2-migration.md`

**Inline calls to action** — these follow a specific problem ("if you need this package version, contact support"), so the reader has a concrete ask and wants the form. They keep their direct portal link and gain one sentence pointing at the prerequisites, so nobody loses a click.

- `content/platform/fips/non-approved-algorithms.md`
- `content/chainguard/containers/about/images-testing.md`
- `content/chainguard/containers/about/versions.md`
- `content/chainguard/containers/staying-secure/cve-risk.md`
- `content/chainguard/containers/features/unique-tags/index.md`

### Deliberately unchanged

Four instances match the same string but are not support calls to action. A find-and-replace would have broken them:

- `content/chainguard/containers/network-requirements.md:60` — `support.chainguard.dev` as a hostname in the required-endpoints table, not a link.
- `content/chainguard/containers/staying-secure/security-advisories/how-to-use/index.md:149` and `content/chainguard/containers/staying-secure/working-with-scanners/false-results.md:37` — deep links to a specific knowledge base article on scanner discrepancies.
- `content/chainguard/containers/about/catalog-starter.md:124` — links to the knowledge base root as a resource for Catalog Starter users, who cannot open tickets.

### One judgment call worth flagging

`non-approved-algorithms.md` has three support calls to action. Adding the same pointer to all three would read as noise on one page, so it appears once, at the first occurrence.

## Fan Yang's addition to Get support

Fan Yang asked in Slack, after #3885 merged, that the page invite supporting material — missing context costs support a round of back-and-forth before an investigation can start. Added as a final bullet under **What to include in your request**:

> - Anything else you think might help: logs, screenshots, or a short screen recording.

The section's opening line already says why this matters ("A support engineer can start on a problem instead of asking you for more information"), so the bullet does not repeat the rationale.

## Verification

`npm run build` completes clean. Link accounting after the change: 4 replaced, 5 pointers added, 7 direct portal links deliberately retained, 4 non-CTA instances untouched. Verified the new bullet and all rewritten links render, and that `/get-started/get-support/` resolves from every page that now points at it.

---
Created in collaboration with Claude Code running Opus 5 on 2026-09-01.

### DIFF
--- a/content/platform/fips/non-approved-algorithms.md
+++ b/content/platform/fips/non-approved-algorithms.md
@@ -6,7 +6,7 @@ linktitle: "FIPS and non-approved algorithms"
 type: "article"
 description: "Technical deep-dive into Chainguard FIPS images access to non-approved algorithms such as MD5 and SHA1"
 date: 2025-10-28T08:00:00+00:00
-lastmod: 2026-08-03T18:16:45+00:00
+lastmod: 2026-09-01T16:34:19+00:00
 draft: false
 tags: ["FIPS", "MD5"]
 images: []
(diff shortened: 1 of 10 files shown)


s-stumbo (approved):

content/platform/fips/non-approved-algorithms.md
```diff
@@ -111,7 +111,7 @@ Error setting context
...
 
-If there is C/C++ software that uses OpenSSL APIs and needs access to MD5 and doesn't support `-fips` property query string, please [open a support request](https://support.chainguard.dev/) for Chainguard engineering to look into adding support.
+If there is C/C++ software that uses OpenSSL APIs and needs access to MD5 and doesn't support `-fips` property query string, please [open a support request](https://support.chainguard.dev/) for Chainguard engineering to look into adding support. See [Get support](/get-started/get-support/) for the portal's prerequisites.
```
s-stumbo: we could change "see" to "refer to" or similar

PR #3890, opened 2026-09-02
docs: add container and version availability troubleshooting

matthewhelmke:
A customer who can't pull a container image, or a specific version of one, now has a page that tells them which of four situations they're in and which ones they can resolve without Chainguard. Three of the four are self-service — add the image from the Console, move to the stream tag, or pin a digest — and the page says plainly which single case needs a support ticket.

Per the Kapa.ai July 2026 export analyzed in DOCS-145, "Request for Older and Missing Image Tags" was the second-largest topic at 18 threads from 14 unique users, and the worst-served: Kapa hedged or deflected to support in 16 of 18 threads, and 10 of 18 cited no edu page at all. No page covered this ground, so Kapa answered from the support knowledge base instead.

## Changes

**New page** — `chainguard/containers/about/container-version-troubleshooting/`, "Troubleshoot container and version availability." It routes on what the Console actually shows, then covers four cases:

1. The image is in the Chainguard Containers Directory but not in your organization's catalog.
2. Chainguard doesn't build the image at all.
3. The image is there, but not the version you need.
4. The version has reached end of life.

Case 3 carries the weight, because it's where the question is usually mis-stated. Chainguard actively maintains the latest patch of each supported stream, and tags float, so a reader asking for an older patch tag is asking for an image that's no longer rebuilt and will accrue CVEs. Readers who want a specific patch for reproducibility are routed to digests. The page mirrors the Console's own guidance to try a supported tag before opening a request.

**Cross-links** from the six pages that describe the states the new page explains:

- Product release lifecycle, at the top and in the EOL section. That page covers the policy thoroughly but never says what to do about a specific missing version.
- Using the Chainguard Console, from both the **Add to org** column and the **Pull URL** access messages.
- Requesting new Chainguard resources, clarifying that the **Requests** section is for resources Chainguard doesn't build at all. Its form asks for a public project name and an upstream repository, so it has no notion of requesting a version of something we already build.
- Onboard your teams, from the failed-pull paragraph.
- Get support, as a new routing case under "Where to send your request."
- The About Chainguard Containers section index.

**Terminology** — standardized on "Chainguard Containers Directory" in Onboard your teams, matching the rest of that section. Left the other files alone to keep this reviewable.

**Path fixes** — two pre-rename links to `/chainguard/chainguard-images/about/catalog-starter/`, in the self-serve overview and the self-serve Helm charts page. Both redirect, but the canonical path is `/chainguard/containers/about/catalog-starter/`.

## Verification

`npm run build` clean at 1653 pages. All in-page anchors resolve to real heading IDs, and all nine outbound links from the new page hit real built pages with no redirects — two needed correcting to the canonical `/platform/administration/...` paths. `check-related-pages` found the second stale `catalog-starter` link, which is why that fix is here.

## Sources

Deliberately not linked from the page itself, since the docs should be the authoritative reference for this:

- The support knowledge base article "How Chainguard Image Tags, Versions, and EOL Work," which documents the request path and the no-retroactive-EOL-builds policy.
- The GTM support runbook's missing-version-tag procedure.
- Zendesk ticket traffic, including a case where a customer asked for a version released well before Chainguard began maintaining the image.

Console strings were read from `console-ui/app/src/pkg/images/details/Versions.tsx` and confirmed against the live Console: the **Looking for older tags?** dialog below the tag table, the pause icon tooltip `This tag is no longer considered active.`, and the **Pull URL** access labels.

Two things I did not put in the page, deliberately. The support knowledge base states that Chainguard doesn't prune digests from the registry; that reads as a durability commitment, and commitments belong in contracts, so the digest guidance stands on content-addressing alone. Support also occasionally adds unsupported older tags for specific customers; that's a rare, customer-initiated exception and not something to advertise.

## Follow-ups, not in this PR

- The two Console strings disagree with each other: the dialog says "made available to your organization" and the tooltip says "added to your organization." The page uses "added to." Worth settling in `console-ui`.
- "Chainguard Directory" versus "Chainguard Containers Directory" is still split across nine other files.

DOCS-155

---
Created in collaboration with Claude Code running Opus 5 on 2026-09-02.

### DIFF
--- a/content/chainguard/containers/about/container-version-troubleshooting.md
+++ b/content/chainguard/containers/about/container-version-troubleshooting.md
@@ -1,0 +1,149 @@
+---
+title: "Troubleshoot container and version availability"
+linktitle: "Availability troubleshooting"
+type: "article"
+description: "When a container or version isn't available to you: how to identify which situation you're in, what to do about each, and when to open a support request."
+date: 2026-09-02T00:00:00+00:00
+lastmod: 2026-09-02T13:22:28+00:00
+draft: false
+tags: ["Chainguard Containers"]
+images: []
+menu:
+  docs:
+    parent: "about"
+weight: 026
+toc: true
+---
+
+You need a container image, or a particular version of one, and you can't pull it. The right next step depends on why it's missing, and there are four distinct reasons. This guide helps you tell them apart and resolve each one.
+
+Start by separating the two surfaces involved, because they aren't the same thing:
+
+* The [Chainguard Containers Directory](https://images.chainguard.dev/) at `images.chainguard.dev` is public. It lists every container image Chainguard builds, along with the tags, sizes, and metadata for each. Anyone can browse it.
+* Your organization's registry at `cgr.dev/$ORGANIZATION/` holds only the images your organization has access to. This is what your teams pull from.
+
+Browsing an image in the Directory doesn't mean your organization can pull it. For a fuller explanation of that distinction and how it relates to your subscription, see [Onboard your teams](/get-started/onboard-your-teams/).
+
+## Find your situation
+
+Look up the image in the Directory, then match what you see to the following table:
+
+| What you find | Go to |
+| --- | --- |
+| The image is in the Directory, but the Console shows **Unavailable to organization**, **Add to organization for access**, or **Request image for access** | [The container isn't in your organization's catalog](#the-container-isnt-in-your-organizations-catalog) |
+| The image isn't in the Directory at all | [Chainguard doesn't build the container](#chainguard-doesnt-build-the-container) |
+| The image is in the Directory, but not the version you need | [The version you need isn't listed](#the-version-you-need-isnt-listed) |
+| The version is listed with a pause icon, an **Expired** status, or an end-of-life date that has passed | [The version has reached end of life](#the-version-has-reached-end-of-life) |
+
+## The container isn't in your organization's catalog
+
+Chainguard builds the image, but your organization hasn't added it yet. What you do next depends on your subscription.
+
+**Catalog customers.** A Catalog subscription covers the whole catalog, but only a subset of images is loaded into your organization's registry at any time. Administrators add more as teams need them. If you have the `owner` role, you can add the image yourself from the Console: go to **Images**, then click **Add image** on the **Organization** tab, or **Add to org** on the **Chainguard catalog** tab. The image appears in your catalog after a few minutes.
+
+Self-serve adds require a role with the `repo (create, list, update)` capabilities, and the `registry.entitlement (list)` capability is useful for understanding what your organization is entitled to. The `owner` role is the only built-in role that carries all of them. For the full walkthrough and the role details, see [Chainguard container catalog pricing](/chainguard/containers/about/pricing/) and [Overview of roles and role-bindings](/platform/administration/iam-organizations/roles-role-bindings/roles-role-bindings/).
+
+If you don't have the `owner` role, ask an administrator in your organization to add the image.
+
+**Per-image customers.** A per-image subscription covers a specific licensed set of images rather than the whole catalog. You can browse everything in the Directory, but only your licensed images are permitted for builds, deployments, and production workloads. Ask your administrators to start a request; once they approve it, they add the image to your organization's registry.
+
+Questions about what your organization is licensed for go to your account team, meaning the customer success manager or solutions architect assigned to your organization.
+
+## Chainguard doesn't build the container
+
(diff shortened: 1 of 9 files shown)


content/chainguard/containers/about/container-version-troubleshooting.md
```diff
@@ -0,0 +1,149 @@
...
+menu:
+  docs:
+    parent: "about"
+weight: 026
+toc: true
+---
+
+You need a container image, or a particular version of one, and you can't pull it. The right next step depends on why it's missing, and there are four distinct reasons. This guide helps you tell them apart and resolve each one.
+
+Start by separating the two surfaces involved, because they aren't the same thing:
```
s-stumbo: This part feels a little conversational. Could we try something more explicit like "First, distinguish between the public container image catalog and your organization’s registry"

content/chainguard/containers/about/container-version-troubleshooting.md
```diff
@@ -0,0 +1,149 @@
...
+Look up the image in the Directory, then match what you see to the following table:
+
+| What you find | Go to |
+| --- | --- |
+| The image is in the Directory, but the Console shows **Unavailable to organization**, **Add to organization for access**, or **Request image for access** | [The container isn't in your organization's catalog](#the-container-isnt-in-your-organizations-catalog) |
+| The image isn't in the Directory at all | [Chainguard doesn't build the container](#chainguard-doesnt-build-the-container) |
+| The image is in the Directory, but not the version you need | [The version you need isn't listed](#the-version-you-need-isnt-listed) |
+| The version is listed with a pause icon, an **Expired** status, or an end-of-life date that has passed | [The version has reached end of life](#the-version-has-reached-end-of-life) |
+
+## The container isn't in your organization's catalog
```
s-stumbo, line 38: should these sections be H3s?

s-stumbo (approved):

content/chainguard/containers/about/container-version-troubleshooting.md
matthewhelmke: Good question. That would push some things further down the page into H4s and I wanted to avoid that. I'll look at the page more closely and see if adjustments can be made to the organization.

matthewhelmke: I did a local build and using H4s dropped those headings from the side menu (and maybe from the search index?) and I don't want that. I'm choosing to leave these as H3s for now. It works.

max-allan-cgr (changes requested):
I think there's a gap on "inactive tags".
Support can add them on request.

It's also worth adding that if people want to request inactive tags, they should specify all the tags they need in a single request to save us some time!

content/platform/console/images-directory/index.md
```diff
@@ -82,7 +84,7 @@ Each container image details page has several tabs that provide information abou
 The default page for each image is the **Tags** tab which contains information about the version tags available for each image. This contains a table with columns:
 
 * **Tag**: this column lists each tag available for the container image
-* **Pull URL**: the URL you can use to download each version of the image. In the Console, Production containers you don't already have access to will show a message reading `Add to organization for access` if you're logged in under an organization with access to Production Containers; if you're logged in under an unverified organization, the message reads `Request image for access`.
+* **Pull URL**: the URL you can use to download each version of the image. In the Console, Production containers you don't already have access to will show a message reading `Add to organization for access` if you're logged in under an organization with access to Production Containers; if you're logged in under an unverified organization, the message reads `Request image for access`. For what to do about either message, refer to [Troubleshoot container and version availability](/chainguard/containers/about/container-version-troubleshooting/).
```
max-allan-cgr: > the message reads `Request image for access`

The message might say image, but should it say Container!? That's a whole different PR!

The Catalog view in the console just says "Request image" 
The tooltip says "Submit request for access to this image" which is maybe where your text came from... After a bit of mangling!

```suggestion
* **Pull URL**: the URL you can use to download each version of the Container. In the Console, Production Containers you don't already have access to will show a message reading `Add to organization` if you're logged in under an organization with access to Production Containers; if you're logged in under an organization without entitlement, the message reads `Request image` or `Request chart`. For what to do about either message, refer to [Troubleshoot container and version availability](/chainguard/containers/about/container-version-troubleshooting/).
```

"Containers" includes charts and images. The sentence starts referring generically to either.Only when it gets to the "Request" button does it need to specify chart/image .

I don't think if your org is unverified you get anything!! Pretty sure that was the wrong word.

max-allan-cgr (changes requested):

content/platform/console/images-directory/index.md
matthewhelmke: I appreciate the closer read! Thank you. Sometimes the mind gets cluttered when a writer gets deep into a topic and we don't see what becomes obvious from two steps back.

max-allan-cgr (approved):

PR #3927, opened 2026-09-04
fips: simplify legacy approved algorithm access

xnox:
Simplify legacy approved algorithm access


matthewhelmke (approved):
LGTM, thank you!

PR #3926, opened 2026-09-04
Reorganize the Chainguard Containers section

SharpRake:
resolves https://linear.app/chainguard/issue/DOCS-110/reorganize-containers-docs

The proposal behind this PR, with the traffic data and the reasoning for each
move: https://claude.ai/code/artifact/b609dddd-21c1-4455-b598-32c07dc383a2

## Type of change

**Documentation Update**
- Reorganizes `content/chainguard/containers` into 13 task-named top-level entries, with every page moved to a final home and a Hugo alias behind it
- Renames the files and directories still carrying pre-rename "images" vocabulary, and strips 32 dead `menu: docs: parent:` blocks
- Removes seven video transcripts whose topic an existing page already covers
- Rebuilds the landing-page banners and tiles, chosen from readership and written against final paths

### What should this PR do?

Restructures the Chainguard Containers section so every nav label names something a reader is doing, no first-level list runs past nine entries, and nothing sits deeper than three levels.

### Why are we making this change?

About, How to use, and Getting started guides held 50 pages between them as flat, unordered lists, and "About", "Features", and "How to use" sorted content by document genre rather than by reader task — so "how do I add a certificate" meant guessing between three sections. Naming and metadata still carried the images→containers rename, and the landing-page tiles had not been revisited in a long time. The structure here follows the reader's path: get access, start, run, customize, migrate, operate.

### What are the acceptance criteria? 

- No URL that resolved before this PR stops resolving. Verified by building the site before and after and diffing every generated path: zero lost, and the only two paths that go are tag paginator pages that shrank because seven tagged pages left.
- No broken internal link is introduced. Verified by crawling every `href="/…"` in the built output against the generated path set — the same 29 pre-existing broken targets before and after, all `/chainguard/chainctl/…` in another section.
- All 49 landing-page banner and tile links land on a real page rather than an alias stub or a 404.
- The sidebar shows the 13 entries in reader order, and `npm run lint:markdown` passes.

### How should this PR be tested?

Any documentation published to Chainguard Academy is reviewed carefully for accuracy. GUI procedures, API commands, and CLI code snippets in a draft are run and tested thoroughly — by both the author and the reviewer — to confirm they work exactly as written. This helps ensure that readers can follow along and get the same results. See the [`edu` repo's README](https://github.com/chainguard-dev/edu#testing).

1. Check the preview link and ensure the changes look right
2. Walk the four commits in order — cleanup, moves, deletions, landing pages — rather than reading one flat diff
[...]

PR #3953, opened 2026-09-09
Add a registry authentication error reference for Chainguard Containers

matthewhelmke:
Adds an error reference for registry authentication and authorization failures, and cross-links it from the pages where those failures happen.

Users arrive holding an exact `cgr.dev` error string and want it decoded, and we had no page that mapped one to the other. In the July 2026 Kapa export (DOCS-157) this cluster ran 11 threads from 9 unique users; Kapa hedged or deflected to support in 7 of 11 and cited no edu page in 4, leaning on a Zendesk registry authentication article 28 times within the cluster instead. This gives those threads a page to land on.

## The reference page

New page at `content/chainguard/containers/troubleshooting/registry-errors.md`, serving `/chainguard/containers/troubleshooting/registry-errors/`. It parallels [Chainguard Libraries error messages](https://edu.chainguard.dev/chainguard/libraries/troubleshooting/errors/) in structure and naming.

It's organized around the two requests a pull actually makes — the credential exchange at `cgr.dev/token`, then the pull itself against `cgr.dev/v2` — because those two layers reuse the same status codes for unrelated problems. Each section is keyed to the verbatim strings users report, so searching the page for a pasted error lands in the right place.

## Please read this part: the issue specified the wrong diagnosis

DOCS-157 asked us to distinguish "a credential failure (401, login rejected) versus an entitlement failure (403, login succeeded but the resource is not in your organization)." I probed `cgr.dev` before drafting, and **neither half holds**. Built as specced, the page would have misdiagnosed the single most-cited error string in the cluster.

At `/token`, unauthenticated:

| Scope | Status | Code |
| -- | -- | -- |
| `chainguard/python` (in the free catalog) | `200` | token issued |
| `chainguard/keycloak` (real product, not free) | `403` | `FORBIDDEN` "Forbidden" |
| `chainguard/nope-not-real-xyz` (nonexistent) | `403` | `FORBIDDEN` "Forbidden" |
| `<private-org>/python` (org has it) | `401` | `UNAUTHORIZED` "Authentication required" |
| `<private-org>/nope-not-real-xyz` | `403` | `FORBIDDEN` "Forbidden" |
| `<nonexistent-org>/python` | `400` | `BAD_REQUEST` `InvalidArgument` |

The real rule is that **`403` means the endpoint won't grant you that scope**, and **`401` means the scope is grantable once you authenticate**. So `401` is the closer thing to an entitlement signal and `403` is not one at all — the reverse of the spec.

The genuine authorization failure lives one layer up, at `/v2/<org>/<repo>/tags/list` with a valid token, and it announces itself in the message text: `403 FORBIDDEN "caller does not have the required capabilities at ..."`, distinct from layer 1's bare `"Forbidden"`. That string was already documented in `private-apk-repos`, and the probe matches it. Absence at that layer surfaces as `404 NAME_UNKNOWN`, not `403`.

So the page's organizing claim is that **the message text, not the status number, is the discriminator.** Two different `403`s mean unrelated things.

Two smaller corrections that follow:

- The issue attributed the `helm registry login` `403` to using an email address as the username. The scopeless `/token?service=cgr.dev` returns that same `403` with no credentials at all, and again with garbage credentials, so it can't support that diagnosis. The remedy is still right — the username must be the identity ID from a pull token — so the page lists it first among things to check rather than presenting it as the cause.
- `400 BAD_REQUEST` on an org name that doesn't resolve is documented nowhere. It's a clean signal for a mistyped or omitted organization, which users don't distinguish from a `403`.

Full measurements are in a comment on DOCS-157 so this doesn't have to be re-derived.

## Cross-links

The existing explanation of what an organization can pull was unreachable from where the error occurs, which was the reachability problem the issue identified:

- **Registry overview** — new sections for what your organization can pull and for troubleshooting authentication errors.
- **Registry authentication** — the error page after the quickstart, plus a pointer from the Console pull-token section to `registry.pull_token_creator`, so "I can't create a pull token" reaches the capability that grants it.
- **Both Helm chart guides** — not in the original plan. The `iamguarded-charts` chart-install `403` in the source data came from exactly that path, and neither Troubleshooting section mentioned authentication at all.

The genuine entitlement case routes to the availability guide from #3890 rather than restating it.

## Scope notes

- DOCS-157 proposed `content/chainguard/containers/errors.md`. #3926 reorganized the section after the issue was filed, so this goes under `troubleshooting/` where the Libraries page now lives and where the sibling troubleshooting pages are.
- The four `chainctl` custom-IdP threads folded into the issue are deferred. They fail at a different service with an unrelated fix, and would read as an intrusion on a page built around the two `cgr.dev` layers. `content/platform/administration/custom-idps/` has no troubleshooting content at all, which is worth its own issue.

## Second and third commits

Separate from the docs change, and reviewable on their own:

- **`chore: ignore the Hugo-generated assets/jsconfig.json`** — Hugo writes this during a build for editor import resolution. It showed up untracked after every build.
- **`chore: update stale pre-rename chainctl link paths`** — 61 links across 30 files still used `/chainguard/chainctl/`. Not broken for readers; production 301s the old prefix, so this removes the redirect hop and stops the stale prefix spreading by copy-paste. All 61 are markdown link targets, so the rewrite is scoped to the `](` form and touches no prose.

## Verification

- `npm run build` clean, 1683 pages.
- `markdownlint-cli2` clean on every changed file.
- Every internal link and anchor on all changed pages resolved against the built site, fragments included. Anchors were checked post-build against `public/`, since heading IDs are generated and a source-only check can't see them.
- Each of the 29 distinct rewritten `chainctl` targets verified against the local build, which has no redirect layer and so fails on a path that is wrong rather than merely stale.
- Error semantics verified by direct probe against `cgr.dev` rather than taken from the reported strings.

Closes DOCS-157

---
Created in collaboration with Claude Code running Opus 5 on 2026-09-09.

### DIFF
--- a/content/chainguard/containers/troubleshooting/registry-errors.md
+++ b/content/chainguard/containers/troubleshooting/registry-errors.md
@@ -4,7 +4,7 @@ linktitle: "Registry errors"
 description: "Map the errors cgr.dev returns during login and pull to their causes, including why the same HTTP status code means different things at the token endpoint and the registry API."
 type: "article"
 date: 2026-09-09T00:00:00+00:00
-lastmod: 2026-09-09T00:00:00+00:00
+lastmod: 2026-09-09T17:21:21+00:00
 draft: false
 tags: ["Chainguard Containers", "Registry"]
 images: []
@@ -33,16 +33,14 @@ The two steps reuse the same HTTP status codes for unrelated problems. A `403` f
 | `FORBIDDEN`, "caller does not have the required capabilities" | Registry API | You're authenticated, but you aren't authorized for this repository. |
 | `NAME_UNKNOWN`, "repository does not exist" | Registry API | The repository name doesn't exist in that organization. |
 
-A bare "Forbidden" body doesn't tell you which of those two it is. Chainguard's registry returns the same `403` to a caller that sent no credentials, a caller whose credentials it can't use, and a caller naming a repository it won't grant a token for. Treat it as a prompt to work through the possible causes in order rather than as a single diagnosis, and in particular don't read it as confirmation that your credentials were accepted.
+A bare `Forbidden` message doesn't tell you which of those two it is. Chainguard's registry returns the same `403` to a caller that sent no credentials, a caller whose credentials it can't use, and a caller naming a repository it won't grant a token for. Treat it as a prompt to work through the possible causes in order rather than as a single diagnosis, and in particular don't read it as confirmation that your credentials were accepted.
 
 ## Authentication required from the token endpoint
 
 You'll see a `401` like the following when a build or pull runs with no credentials configured:
 
 ```output
-failed to fetch anonymous token: unexpected status from GET request to
-https://cgr.dev/token?scope=repository%3A$ORGANIZATION%2Fpython%3Apull&service=cgr.dev:
-401 Unauthorized
+failed to fetch anonymous token: unexpected status from GET request to https://cgr.dev/token?scope=repository%3A$ORGANIZATION%2Fpython%3Apull&service=cgr.dev: 401 Unauthorized
 ```
 
 The word "anonymous" is the signal. Your tool found no credentials for `cgr.dev`, so it asked for a public token, and the repository you named isn't public. Containers in the `cgr.dev/chainguard/` namespace are public and need no authentication. Everything in your organization's own namespace at `cgr.dev/$ORGANIZATION/` requires credentials.
@@ -62,8 +60,7 @@ If the failure happens inside `docker build` rather than `docker pull`, check th
 A `403` from the token endpoint looks like this, often from `helm registry login` or another tool that logs in without naming a repository:
 
 ```output
-Error: authenticating to "cgr.dev": GET "https://cgr.dev/token?service=cgr.dev":
-response status code 403: forbidden
+Error: authenticating to "cgr.dev": GET "https://cgr.dev/token?service=cgr.dev": response status code 403: forbidden
 ```
 
 This response means the token endpoint rejected the request outright. It doesn't tell you which part of the request it objected to, so check these in order:
@@ -77,8 +74,7 @@ This response means the token endpoint rejected the request outright. It doesn't
 A `400` means the organization portion of the image reference didn't match any Chainguard organization:
 
 ```output
-{"errors":[{"code":"BAD_REQUEST","message":"rpc error: code = InvalidArgument
-desc = unable to resolve ..."}]}
+{"errors":[{"code":"BAD_REQUEST","message":"rpc error: code = InvalidArgument desc = unable to resolve ..."}]}
 ```
 
 Unlike the other errors on this page, this one has nothing to do with your credentials. The organization name is wrong. List the organizations you belong to and compare:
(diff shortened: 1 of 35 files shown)


content/chainguard/containers/troubleshooting/registry-errors.md
```diff
@@ -0,0 +1,191 @@
...
+
+| Response body | Returned by | Meaning |
+| -- | -- | -- |
+| `UNAUTHORIZED`, "Authentication required" | Token endpoint | The repository requires credentials and you presented none that worked. |
+| `FORBIDDEN`, "Forbidden" | Token endpoint | The endpoint won't issue a token for the repository you named, or it couldn't use the credentials you sent. |
+| `BAD_REQUEST`, "InvalidArgument" | Token endpoint | The organization in the image reference didn't resolve. |
+| `FORBIDDEN`, "caller does not have the required capabilities" | Registry API | You're authenticated, but you aren't authorized for this repository. |
+| `NAME_UNKNOWN`, "repository does not exist" | Registry API | The repository name doesn't exist in that organization. |
+
+A bare "Forbidden" body doesn't tell you which of those two it is. Chainguard's registry returns the same `403` to a caller that sent no credentials, a caller whose credentials it can't use, and a caller naming a repository it won't grant a token for. Treat it as a prompt to work through the possible causes in order rather than as a single diagnosis, and in particular don't read it as confirmation that your credentials were accepted.
```
SharpRake:
```suggestion
A bare `FORBIDDEN` body doesn't tell you which of those two it is. Chainguard's registry returns the same `403` to a caller that sent no credentials, a caller whose credentials it can't use, and a caller naming a repository it won't grant a token for. Treat it as a prompt to work through the possible causes in order rather than as a single diagnosis, and in particular don't read it as confirmation that your credentials were accepted.
```

content/chainguard/containers/troubleshooting/registry-errors.md
```diff
@@ -0,0 +1,191 @@
...
+desc = unable to resolve ..."}]}
+```
+
+Unlike the other errors on this page, this one has nothing to do with your credentials. The organization name is wrong. List the organizations you belong to and compare:
+
+```shell
+chainctl iam organizations list
+```
+
+Organization names are usually domain names, such as `example.com`, and the full image reference is `cgr.dev/$ORGANIZATION/$IMAGE:$TAG`. Omitting the organization, or using your organization's display name instead of its registry name, produces this error.
```
SharpRake:
```suggestion
Organization names are usually domain names, such as `example.com`, and the full image reference is `cgr.dev/$ORGANIZATION/$IMAGE:$TAG`. Omitting the organization, or using your organization's display name instead of its registry name, can produce this error.
```

content/chainguard/containers/troubleshooting/registry-errors.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Missing capabilities from the registry API
+
+This `403` is the one that means your login worked and your authorization didn't:
+
+```output
+{"errors":[{"code":"FORBIDDEN","message":"caller does not have the required
+capabilities at \"...\""}]}
```
SharpRake: is this supposed to be on 2 lines?

content/chainguard/containers/troubleshooting/registry-errors.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Repository does not exist
+
+Once you hold a valid token, a name that isn't in the organization returns a `404` rather than a `403`:
+
+```output
+{"errors":[{"code":"NAME_UNKNOWN","message":"repository does not exist \"...\""}]}
+```
+
+Check the spelling against `chainctl images repos list`, then against the [Chainguard Containers Directory](https://images.chainguard.dev/). If the Directory doesn't list the container either, Chainguard doesn't build it yet and you can ask for it. Refer to [Troubleshoot container and version availability](/chainguard/containers/troubleshooting/container-version-troubleshooting/).
```
SharpRake: you could also point folks here: https://edu.chainguard.dev/chainguard/containers/reference/request-resources/

SharpRake (approved):
This is a great resource! Excited that we have it

Just a few small comments

content/chainguard/containers/troubleshooting/registry-errors.md
matthewhelmke: Good catch, and no — it should be one line. The registry returns that on one line, and wrapping it broke two things: it made the JSON invalid (copying it into `jq` fails), and the break landed inside the `message` value, so it reads as though the message itself contains a newline.

Fixed here and in five other blocks with the same problem. All three JSON bodies now parse. The `NAME_UNKNOWN` block below was already on one line, which I think is what made this one look off.

Long lines scroll rather than wrap — the theme sets `overflow-x: auto` on `pre code` — so the unwrapped versions do not break the layout.

content/chainguard/containers/troubleshooting/registry-errors.md
matthewhelmke: Taken, with one tweak: I used `Forbidden` (the message) rather than `FORBIDDEN` (the code), because `FORBIDDEN` is the code on *both* 403s — the token endpoint one and the registry API one. Naming the code here would blur the exact distinction this section is drawing, since the discriminator is the message text. Kept your code font and changed "body" to "message" to make that explicit.

content/chainguard/containers/troubleshooting/registry-errors.md
matthewhelmke: Taken, thanks — "can produce" is more accurate, since omitting the organization entirely can fail differently.

content/chainguard/containers/troubleshooting/registry-errors.md
matthewhelmke: Added, thanks. It now links [Requesting new Chainguard resources](/chainguard/containers/reference/request-resources/) directly for the "Chainguard does not build it" case, and keeps the availability guide as the pointer for working out which situation you are in.

PR #3954, opened 2026-09-09
Add a task page for adding a package to a Chainguard Container

matthewhelmke:
Adds a task page for adding a package to a Chainguard Container, and fixes the pages it routes to.

Closes [DOCS-161](https://linear.app/chainguard/issue/DOCS-161/clarify-adding-package-to-a-chainguard-image).

While writing it I found that we tell customers a supported CI/CD capability doesn't exist: the `chainctl` guide states you cannot use `--save-as` with the `apply` subcommand, so anyone automating Custom Assembly reads that as "you can't create a new image from a pipeline." You can. Only batch mode excludes it. That correction is in this PR.

## Why

DOCS-161 came out of the Kapa mining pass (DOCS-145): a lot of questions about how to install additional packages in a Chainguard Container, where Kapa's own answer was easier to follow than ours because it plainly listed the ways to do it. Five of our pages answer some part of this question and none of them is the page a reader lands on.

Two steps existed nowhere in the docs: how to find out what a package is called before you add it, and how to confirm afterward that it actually reached the image.

## The new page

`content/chainguard/containers/building-and-modifying/adding-packages.md`, serving `/chainguard/containers/building-and-modifying/adding-packages/`. It routes rather than duplicates — the deep procedures stay where they are.

- A comparison table of all six routes: Custom Assembly through the Console, `chainctl` interactively, `chainctl` non-interactively, and the API; plus `apk add` in a Dockerfile and the chroot method for distroless.
- Custom Assembly prerequisites in one place: entitlement, `repo.update` versus `repo.create`, and `chainctl` login.
- **Find the package name** — new. Console filter, and `apk search` for both free and entitled images.
- Short procedures for the three Custom Assembly paths, each linking to the full guide.
- **Confirm the package is in the image** — new. Build status, then `apk info -e` for images with a package manager, and an SBOM query for distroless.
- What to do when a build fails.

## Corrections to existing pages

- **`custom-assembly-chainctl.md`**: `--save-as` does work with `apply`, verified against `chainctl` 0.2.347 and the generated reference, which carries an explicit example. The page said it doesn't. Corrected, and the non-interactive `--save-as` form is now shown.
- **`custom-assembly-gitops.md`**: "interactively using `chainctl`" linked to the API demo page.
- **`containers/overview.md`**: the "Install and manage additional packages" link went to the package-repository architecture page rather than to anything procedural. It also described Custom Assembly as `chainctl`-only, omitting the Console and the API.
- A `cgr.dev/example.come/node` typo.

## Structural changes

- Split the merged section in `custom-assembly-chainctl.md` into `Editing packages interactively` and `Applying packages non-interactively`. The `apply` path previously started mid-section under an `edit` heading, so it never appeared in the table of contents — which is the specific thing DOCS-161 asks us to fix.
- Documented `--dry-run` and `edit --file`.
- Normalized `chainctl image repo build` to `chainctl images repos build`. Both forms work, but the page used each in different places, which reads like two different commands.
- Turned the Custom Assembly overview's bulleted list of interfaces into a table that names both `chainctl` modes and GitOps, and says when to pick each.

## Testing

- `npm run build` clean, 1687 pages.
- Every internal link and `#fragment` on all ten changed pages resolves against the built site.
- Commands in the new page were run, and the sample outputs are captured rather than written by hand: `apk search -e` and `apk info -e` against `wolfi-base`, the SBOM `pkg:apk/` filter against `cgr.dev/chainguard/python:latest`, and `chainctl images repos build list` against a real customized repository.
- Not tested end to end: applying a Custom Assembly change, since that mutates a repository and starts a build. Everything run against the platform was read-only.

## Note for reviewers

The new page uses **Private APK Repository** per the brand style guide. The Word Usage Guide in this repo says `apk - never capitalized`, and the rest of `content/` uses the lowercase form roughly 66 times to 4, so the two guides conflict and the corpus follows neither consistently. I've kept this PR to the brand-guide form on new text only, rather than sweeping 70 occurrences into an unrelated diff. Happy to file that separately if we want a single answer.

---
Created in collaboration with Claude Code running Claude Opus 5 on 2026-09-09.

### DIFF
--- a/content/chainguard/containers/building-and-modifying/adding-packages.md
+++ b/content/chainguard/containers/building-and-modifying/adding-packages.md
@@ -1,0 +1,253 @@
+---
+title: "Adding a package to a Chainguard Container"
+linktitle: "Add a package"
+description: "Choose how to add a package to a Chainguard Container, find the package name, apply the change, and confirm the package reached the finished image."
+type: "article"
+date: 2026-09-09T00:00:00+00:00
+lastmod: 2026-09-09T00:00:00+00:00
+draft: false
+tags: ["Chainguard Containers", "Custom Assembly", "Procedural"]
+images: []
+weight: 005
+toc: true
+---
+
+Chainguard Containers ship with only the packages their application needs, so sooner or later you'll want one that isn't there. [Custom Assembly](/chainguard/containers/custom-assembly/overview/) is the supported way to add it. You declare the package you want, Chainguard builds the image on its own infrastructure, and Chainguard rebuilds that image whenever the package is updated. You can drive Custom Assembly from the Chainguard Console, interactively with `chainctl`, or non-interactively with `chainctl` from a pipeline.
+
+This page helps you pick an approach, then covers the three steps that apply whichever one you pick: finding the package name, adding the package, and confirming that it reached the finished image.
+
+## Choose an approach
+
+The following table compares the available approaches:
+
+| Approach | Use it when | What you need |
+| --- | --- | --- |
+| [Custom Assembly in the Console](#add-the-package-in-the-console) | You want to browse the packages your organization can add and apply the change in a few clicks. | A Console account with a role that has the `repo.update` capability. |
+| [Custom Assembly with `chainctl`, interactively](#add-the-package-with-chainctl-interactively) | You work from a terminal and want to review a diff before it applies. | `chainctl`, installed and authenticated. |
+| [Custom Assembly with `chainctl`, non-interactively](#add-the-package-with-chainctl-non-interactively) | You keep image configuration in version control or apply it from CI/CD. | `chainctl` and a YAML build configuration file. |
+| [Custom Assembly with the Chainguard API](/chainguard/containers/custom-assembly/custom-assembly-api-demo/) | You're building your own tooling around Custom Assembly. | An API client and a Chainguard token. |
+| [`apk add` in a Dockerfile](/chainguard/containers/using-and-deploying/using-containers/#extending-chainguard-base-containers) | You build on a `-dev` variant or on `wolfi-base`, and you're prepared to pin package versions and image digests yourself. | A Dockerfile and a container image that includes `apk`. |
+| [`apk` with `chroot` in a multi-stage build](/chainguard/containers/building-and-modifying/install-apks-in-distroless-variants/) | You need a package in a distroless image and Custom Assembly doesn't fit your workflow. | A multi-stage Dockerfile. |
+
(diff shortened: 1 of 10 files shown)


content/chainguard/containers/building-and-modifying/adding-packages.md
```diff
@@ -0,0 +1,253 @@
...
+
+Custom Assembly adds packages to an image; it can't remove the packages the source image already contains. You can, however, remove packages you added in an earlier build.
+
+## Find the package name
+
+You can add only the packages your organization is entitled to, which are the packages that appear in the Chainguard Containers you already have access to. Package names often carry a version stream — `python-3.14` rather than `python` — so confirm the exact name before you add it.
+
+### Find a package in the Console
+
+Open the image in the [Chainguard Console](https://console.chainguard.dev), click **Customize image**, then use the **Filter packages** box. The list holds every package your organization can add. If the package you want isn't listed, open a Chainguard support ticket.
```
s-stumbo:
```suggestion
Open the image in the [Chainguard Console](https://console.chainguard.dev), click **Customize image**, then use the **Filter packages** box. The list holds every package your organization can add. If the package you want isn't listed, [open a Chainguard support ticket](/get-started/get-support/).
```

content/chainguard/containers/building-and-modifying/adding-packages.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Find a package with apk
+
+Container images that include `apk` — a `-dev` variant does — can search the repository from inside a running container. For a free container image, no authentication is needed:
```
s-stumbo: This is a little confusing at first. Maybe we could rephrase to something like "Container images that include apk - such as a -dev variant..."

content/chainguard/containers/building-and-modifying/adding-packages.md
```diff
@@ -0,0 +1,253 @@
...
+docker run -it --rm --entrypoint /bin/sh --user root \
+  -e "HTTP_AUTH=basic:apk.cgr.dev:user:$(chainctl auth token --audience apk.cgr.dev)" \
+  cgr.dev/$ORGANIZATION/$CONTAINER:latest-dev
+```
+
+From the container's shell, run `apk update` and then `apk search`.
+
+## Add the package
+
+The three procedures that follow all produce the same result. Pick the one that matches how you work.
```
s-stumbo: Have you considered putting these options into tabs?

matthewhelmke: It hadn't even crossed my mind, but I'm considering it now. I think I may start off my Thursday morning doing exactly this.

s-stumbo (approved):

content/chainguard/containers/building-and-modifying/adding-packages.md
matthewhelmke: This turned out to be a great idea! Thank you!! I would love your opinion one last time on this change before I merge. The latest commit adds tabs to this and the next section: https://deploy-preview-3954--ornate-narwhal-088216.netlify.app/chainguard/containers/building-and-modifying/adding-packages/#find-the-package-name

PR #3955, opened 2026-09-09
Document Sigstore bundle signatures for Chainguard Containers

codysoyland:
Chainguard is moving container image signatures and attestations from the legacy `sha256-<digest>.sig`/`.att` tags on Rekor v1 to Sigstore bundles attached as OCI referrers and logged to Rekor v2 (CON-1109). Customers need the Cosign version floor, a list of what is affected, and a heads-up on the one output-format change before the breaking-change notice goes out, and the notice should be able to link to tested guidance instead of restating it.

This PR adds a migration hub page and threads the version floor through the verification and install pages. The `cosign verify` and `verify-attestation` commands are unchanged; Cosign 3.1.1 and newer detects the format automatically. Every command was run against `cgr.dev/cosign-v3-testing/request-cosign-v3-canary` with cosign 3.1.3, 2.6.5 and 2.6.2, and the compatibility table on the hub page is the observed result.

Key changes:
- **New hub page** `security-and-compliance/migrating-to-sigstore-bundles/`: what changes, an "am I affected" checklist, the Cosign/policy-controller/Kyverno compatibility summary, and what to do; no commands live here so there is one canonical home per snippet.
- **Verify guide**: Cosign 3.1.1 floor with a note on 2.6.3–2.6.5 (`--use-signed-timestamps`), a short "how signatures are stored" section (`cosign tree` / `oras discover`), and a `--trusted-root` pointer for restricted networks since Rekor is not contacted during verification.
- **`download attestation` pipe**: `jq -r '.dsseEnvelope.payload // .payload'` so the same command works for bundle output (whole bundle per line) and the legacy envelope.
- **Install page**: 3.1.1 floor note; replaces the stale "newest release is v3.0.2" example, which also used the wrong `/v2` Go module path.
- **SBOMs vs attestations concept page**: notes that the `.att`/`.sig`/`.sbom` tag layout is Cosign 2.x behaviour and that 3.x stores bundles as referrers.
- **Section index**: lists the hub page under Security and compliance.

Assisted by Claude Fable 5.1

### DIFF
--- a/content/chainguard/containers/security-and-compliance/_index.md
+++ b/content/chainguard/containers/security-and-compliance/_index.md
@@ -9,7 +9,7 @@ aliases:
 description: "Scanning, SBOMs, verification, security advisories, policy enforcement, and the compliance evidence Chainguard Containers carry."
 type: "article"
 date: 2024-12-19T08:49:15+00:00
-lastmod: 2026-08-03T12:43:21+00:00
+lastmod: 2026-09-09T18:49:52+00:00
 draft: false
 images: []
 weight: 090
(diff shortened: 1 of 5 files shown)


codysoyland:

PR #3956, opened 2026-09-09
Update attestation snippets for Sigstore bundle output

codysoyland:
Companion to #3955 (CON-1109). Several pages pipe `cosign download attestation` through `jq -r .payload`, which assumes the legacy DSSE envelope at the top level. For attestations published as Sigstore bundles, Cosign prints the whole bundle and the envelope sits under `.dsseEnvelope`, so those pipelines produce nothing once images stop carrying legacy tags. This PR makes every such pipeline shape-agnostic and states the Cosign 3.1.1 floor wherever a page installs or requires Cosign. Two pages link to the migration guide added in #3955; those links resolve once that PR merges.

Key changes:
- **Download pipes**: `jq -r '.dsseEnvelope.payload // .payload'` on the retrieve-SBOMs, quickstart, onboarding, SLSA, and inspecting-containers pages; tested against `cgr.dev/chainguard/php` (legacy) and the bundle-only canary image.
- **Version floor**: "Cosign 3.1.1 or newer" in the prerequisites of retrieve-SBOMs, quickstart, and repro; the `--platform` note no longer cites 2.2.1.
- **Custom Assembly GitOps example**: pins `cosign-installer` to `cosign-release: v3.1.3`, since the installer's default is still a 2.x cosign.
- **Trivy tutorial**: the Rekor SBOM lookup only finds Rekor v1 entries, which is the path being retired; replaced with downloading the per-platform SPDX attestation and `trivy sbom`, verified to yield the same 19 packages for `nginx` as the Rekor lookup did.

Assisted by Claude Fable 5.1

### DIFF
--- a/content/chainguard/containers/concepts/how-we-build-and-test/repro.md
+++ b/content/chainguard/containers/concepts/how-we-build-and-test/repro.md
@@ -11,7 +11,7 @@ description: "This video explains the importance of reproducibility and how to r
 Chainguard image from an attestation."
 type: "article"
 date: 2024-05-20T12:21:01+00:00
-lastmod: 2024-05-20T12:21:01+00:00
+lastmod: 2026-09-09T19:47:59+00:00
 draft: false
 images: []
 weight: 030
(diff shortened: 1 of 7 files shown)


codysoyland:

PR #3957, opened 2026-09-09
Add mirroring guidance for bundle-signed Chainguard Containers

codysoyland:
Part of the Sigstore bundle migration docs (CON-1109; see #3955 for the hub page). Chainguard is moving image signatures and attestations to Sigstore bundles attached as OCI referrers. Most copy tools follow the image tag and drop referrers, so a mirrored image arrives unsigned and nothing fails until `cosign verify` runs on the other side. The docs had no mirroring guidance at all, and the air-gapped verification guide merged last week asserts that Chainguard keeps signatures under `.sig`/`.att` tags rather than in the referrers graph, which inverts after the cutover.

Every command was run against the bundle-signed canary image, copying into a local `registry:3` with one repository per tool so referrers could not leak between tests; the tools table on the new page is the observed result. Links to the migration guide resolve once #3955 merges.

Key changes:
- **New page** `registry/mirroring-signed-images/`: `oras cp -r` registry-to-registry, the OCI-layout procedure for air gaps, the `--to-distribution-spec v1.1-referrers-tag` option for registries without the Referrers API, a tested tools comparison, and how to check a pull-through cache.
- **Air-gapped guide**: replaces `cosign copy`/`cosign save` with the oras equivalents (`cosign save`/`load` 3.1.3 fails to load bundle referrers; `cosign copy` is deprecated and carried only some bundles), inverts the storage-model note, updates the `cosign tree` example to show referrers, scopes `COSIGN_REPOSITORY` to the legacy layout, and raises the floor to 3.1.1.
- **Pull-through guides** (ECR, Harbor, Artifactory, Nexus, Cloudsmith, Artifact Registry): a "Signatures and attestations" section with the `oras discover`/`cosign verify` check through the cache; ECR additionally notes AWS's referrer sync and that ECR rejects referrer pushes.
- **Network requirements**: the TUF host row now covers Cosign, Kyverno and policy-controller, with a note that verification does not contact Rekor, Fulcio, or the timestamp authority.

Assisted by Claude Fable 5.1

### DIFF
--- a/content/chainguard/containers/registry/mirroring-signed-images/index.md
+++ b/content/chainguard/containers/registry/mirroring-signed-images/index.md
@@ -1,0 +1,108 @@
+---
+title: "Mirroring Chainguard Containers with their signatures"
+linktitle: "Mirroring signed containers"
+type: "article"
+description: "Copy Chainguard Containers into your own registry or across an air gap without losing their Sigstore signatures and attestations"
+lead: "Signatures and attestations are separate artifacts. Copy tools that only follow the image tag leave them behind."
+date: 2026-09-09T00:00:00+00:00
+lastmod: 2026-09-09T00:00:00+00:00
+draft: false
+tags: ["Chainguard Containers", "Procedural"]
+images: []
+toc: true
+weight: 045
+---
+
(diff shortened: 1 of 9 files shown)


codysoyland:

PR #3980, opened 2026-09-11
Remove unnecessary --parent flags from chainctl examples (DOCS-25)

SharpRake:

## Type of change

**Documentation Update**
- Removes 109 unnecessary `--parent` flags from `chainctl` examples across 35 pages
- Trims surrounding prose to a short note for readers in more than one organization
- Keeps the flag where it is genuinely required, and in CI/automation examples
- Cleans up variables and placeholder prose the removals orphaned

### What should this PR do?

resolves https://linear.app/chainguard/issue/DOCS-25/audit-remove-parent-flag-from-chainctl-commands

Removes the `--parent` flag from `chainctl` examples outside the chainctl reference section, where it is optional for the single-organization case that describes most customers.

Most customers belong to exactly one organization, so `--parent` is noise in nearly every example — it lengthens commands and implies a required argument. `chainctl`'s interactive group resolver short-circuits when exactly one candidate is available, printing `One organization available, selecting <org>` and continuing without prompting. That path needs no TTY, so the shortened examples still work in scripts.

- `chainctl iam external-group-role-mappings list` keeps `--parent`; it is the only command that marks the flag required, and omitting it errors with `required flag(s) "parent" not set`
- CI and automation examples keep the flag, since a prompt would break a pipeline: `automating-chainctl.md`, the Custom Assembly GitOps workflow, and the GitHub identity workflow
- Commands that resolve against groups rather than organizations (`iam roles create`, `iam role-bindings create`, `images repos build *`) carry a note that they prompt when the org contains folders
- No organization variable is left set but unused, and no prose refers to a placeholder that no longer appears in its command

1. Check the preview link and ensure the changes look right
2. Run a shortened command as a single-org user and confirm it auto-selects, for example `chainctl policies binding list` or `chainctl auth pull-token list`
3. Confirm `chainctl iam external-group-role-mappings list` still fails without `--parent` in `platform/administration/custom-idps/grant-roles-from-groups/`
4. Review the Custom Assembly pages (`containers/custom-assembly/`) and the new `containers/building-and-modifying/adding-packages.md`, which carry the largest share of the removals

### DIFF
--- a/content/chainguard/actions/overview.md
+++ b/content/chainguard/actions/overview.md
@@ -37,12 +37,6 @@ To follow this guide, you need:
 - An active Chainguard organization.
 - Owner access on the organization.
 
-The examples in this guide use an `$ORGANIZATION` environment variable to refer to your organization. Set it to the name of your organization before you begin:
-
-```shell
-export ORGANIZATION=<your-organization>
-```
-
 ## Preliminary steps
 
 Before using Chainguard Actions, log in to Chainguard and enable the Chainguard Actions entitlement for your organization.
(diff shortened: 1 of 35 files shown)


matthewhelmke (approved):
LGTM

PR #3984, opened 2026-09-11
Document the Available in organization label on the Tags tab

matthewhelmke:
Three users reported the same thing in August: the Console showed **Available in organization** beside a container version, `docker pull` returned not found, and one opened a support ticket titled "A version appears at images.chainguard.dev but is missing in our org." Kapa fielded six threads from five unique users on this and cited no edu page in four of them, because no page described what they were looking at. This adds the missing entry point.

## What the label actually means

`PullUrlCell` renders the real pull URL whenever a row has one. A status label is only the fallback for a row with **no** URL, and `sourceAccessType` is a single repository-level value handed identically to every row. So **Available in organization** in the Pull URL column means *this version has no pull URL for you, and the repository is in your organization* — phrased as reassurance, in a column whose heading implies a per-row statement.

The page now states that once, near the decision table, which disambiguates all six labels the column can show rather than only this one.

## Changes

**`container-version-troubleshooting.md`** — adds a decision-table row for the badge plus a failing pull, and a section covering the two causes: a sync still in flight, and a tag that isn't in the organization's registry. Quotes both sync-state tooltips verbatim from source. Corrects one stale "Versions tab" reference from DOCS-155; the tab routes to `/versions` but its visible label is **Tags**.

**`platform/console/images-directory/index.md`** — the Pull URL bullet listed two of the six labels, both truncated, and attributed `Request chart` to the container table. That string belongs to `helm/HelmTable.tsx`, and this page already covers Helm charts in its own section.

**Three more pages naming the same tab** — `cve-visualizations` (twice) and `eol-grace-period` sent readers to a **Versions** tab or page. Corrected to **Tags**, including one image alt text, so the screen reader description matches what is on screen.

**The Console's real names for the images views** — three pages sent readers to an **Organization Images** tab or sidebar item, and to a **Browse Containers** section. Neither string exists in `console-ui`'s translation catalog. The left navigation item is **Images**, and its two tabs are **Organization** and **Chainguard catalog** (`Navigation.tsx`, and the images route definitions in `ConsoleRouter.tsx`).

Rather than fix only the reported labels, I checked every bolded UI label on the edited pages against `app/src/locales/en.po`, which holds every translatable string the Console ships. That turned up more staleness in `cve-visualizations`' description of the **Reports** page — the **Compare Containers** and **Historical CVEs** tab names are absent from the catalog entirely, and two chart headings differ in case. **Those are left untouched here.** Correcting them means describing a UI as it is now, which needs eyes on the Console rather than a string lookup, and guessing would trade one wrong label for another. Tracked in [DOCS-195](https://linear.app/chainguard/issue/DOCS-195/audit-console-ui-labels-across-the-docs-against-console-uis), which carries the full scan: 103 label candidates across 21 Console pages, ranked, with the triage caveats.

## A finding that changed the recommendation

The draft told readers to check `chainctl images tags list --repo=$IMAGE`. Testing against a live organization killed that:

```
$ docker pull cgr.dev/$ORGANIZATION/ingress-nginx-controller:1.10.0
Error response from daemon: unknown: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"Unknown manifest"}]}
```

For that tag, `chainctl` lists it with a digest, the Console renders a real pull URL on its row, and the registry refuses to serve the manifest. Tag records outlive the images they name, so the full listing overstates what an organization can pull — the check would have sent some readers away believing a tag was available when it is not. The page now recommends `--active-only` and says plainly that a tag's presence in a listing is a weaker signal than whether the tag is still active.

Tested across six superseded tags spanning 175 days to nearly three years old (`1.8.2`, `1.9.0`, `1.10.0`, `1.11.0`, `1.12.0`, `1.13.8`). All six are listed by `chainctl`; none pull. The page still says a superseded tag *can* fail rather than *will*, because six tags in one organization is a pattern and not a guarantee.

Worth flagging beyond this PR: any availability or freshness tooling that treats "the tag is listed" as "the tag is pullable" has the same hole. Why the manifest is gone while the record remains is not established, and the page does not speculate about it.

## Verification

* Console behavior read from `chainguard-dev/console-ui` at `main` on 2026-09-11: `getAccessType` in `self-service.ts`, `PullUrlCell` in `Versions.tsx`, tab titles in `ImageDetails.tsx`.
* `chainctl images tags list` exercised against a live organization. `--active-only` confirmed as a strict subset, keeping the current patch of each maintained stream plus the floating tags.
* `--public` returns `Error: no repositories found` for a paid-catalog image, so no public-versus-organization tag comparison is offered.
* `hugo --gc --minify` exit 0, 1689 pages. All 14 in-page anchors resolve against the built HTML.

Two claims in DOCS-193's original description did not survive source review and are corrected in comments on the issue: `IN_ORG_ADD_IMAGE` is not a distinct access state from `IN_ORG` (both come from `inOrg && hasActive`; the only discriminator is whether the viewing user holds the `ADD_IMAGE` capability), and `showInactive: true` suppresses the label on inactive rows rather than showing it.

Resolves DOCS-193.

---
Created in collaboration with Claude Code running Opus 5 on 2026-09-11.

### DIFF
--- a/content/chainguard/containers/troubleshooting/container-version-troubleshooting.md
+++ b/content/chainguard/containers/troubleshooting/container-version-troubleshooting.md
@@ -6,7 +6,7 @@ aliases:
 type: "article"
 description: "When a container or version isn't available to you: how to identify which situation you're in, what to do about each, and when to open a support request."
 date: 2026-09-02T00:00:00+00:00
-lastmod: 2026-09-11T12:56:16+00:00
+lastmod: 2026-09-11T13:20:58+00:00
 draft: false
 tags: ["Chainguard Containers"]
 images: []
@@ -28,7 +28,7 @@ Browsing an image in the Directory doesn't mean your organization can pull it. F
 
 ## Find your situation
 
-Several of these situations turn on a label in the Console. On a container's **Tags** tab, the **Pull URL** column shows the pull URL for a version when your organization can pull that version. When it can't, the column shows a status label instead, and that label describes the repository rather than the version on that row. A label where you expected a URL means the version isn't available to you.
+You identify several of these situations from a label in the Console. On a container's **Tags** tab, the **Pull URL** column shows the pull URL for a version when your organization can pull that version. When it can't, the column shows a status label instead, and that label describes the repository rather than the version on that row. A label where you expected a URL means the version isn't available to you.
 
 Look up the image in the Directory, then match what you see to the following table:
 
@@ -83,13 +83,13 @@ When a container is added to an organization, only its actively supported tags c
 
 Those records outlive the images they name. A superseded tag can still appear in the Console with a pull URL beside it, and still be listed by `chainctl`, while pulling it returns `MANIFEST_UNKNOWN`. A tag's presence in a listing is a weaker signal than whether that tag is still active.
 
-So check against the active tags, which uses your default organization:
+So check against the active tags. This command reads your default organization:
 
 ```shell
 chainctl images tags list --repo=$IMAGE --active-only
 ```
 
-If the tag you want isn't in that output, treat it as unavailable, whatever the Directory or the Console shows for it. Drop `--active-only` to see everything your organization holds, including tags that are no longer maintained and may no longer pull.
+If the tag you want isn't in that output, treat it as unavailable, whatever the Directory or the Console shows for it. Drop `--active-only` to see everything your organization holds, including tags that are no longer maintained. Some of those will fail to pull.
 
 What you do next depends on which kind of tag it is:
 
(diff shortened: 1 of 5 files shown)


content/chainguard/containers/troubleshooting/container-version-troubleshooting.md
```diff
@@ -17,7 +17,7 @@ weight: 010
 toc: true
 ---
 
-You need a container image, or a particular version of one, and you can't pull it. The right next step depends on why it's missing, and there are four distinct reasons. This guide helps you tell them apart and resolve each one.
+You need a container image, or a particular version of one, and you can't pull it. The right next step depends on why it's missing, and there are five distinct reasons. This guide helps you tell them apart and resolve each one.
```
s-stumbo, line 20: Just to make this evergreen in case other scenarios are listed, we could rephrase to avoid stating the exact number of reasons

s-stumbo (approved):

PR #3978, opened 2026-09-10
Add the Chainguard OpenSSL 4.0 configuration page

xnox:
Adds `content/chainguard/chainguard-os/openssl-4.0.md`, a sibling of the OpenSSL 3.6 page, with the same four tables (TLS cipher suites, supported groups, signature schemes, elliptic curves) populated from loopback handshake tests rather than from `openssl list` output.

Also add PQC security category strengths as well as update 3.6 page to match.

PR #3990, opened 2026-09-11
Head the OpenSSL table columns with the OpenSSL and FIPS provider versions

xnox:
On the OpenSSL configuration pages, head the two availability columns with the versions they describe instead of Non-FIPS / FIPS:

- OpenSSL 3.6 Configuration: **v3.6** and **FIPS v3.4**
- OpenSSL 4.0 Configuration: **v4.0** and **FIPS v3.6**

The legend sentence on each page is updated to match, and the 3.6 page picks up the "enabled by default" wording fix that #3978 applied to the 4.0 page.

Tables are regenerated by the generator in sandbox-of-power `xnox/tls-docs` (c11cc922), which now carries the header names per profile; no table content changes.


matthewhelmke (approved):
LGTM

PR #3999, opened 2026-09-14
docs(container-policies): updates policies docs to include latest changes

Rodrigo-Munoz-Salas:
Updating the policies docs based on all recent changes that have landed.

1. Each system policy now displays reasoning of denial decisions.
2. Policy decisions table output has changed (digest --> artifact id) and added a new reasons column

Refs: CON-2964


matthewhelmke (approved):
LGTM

PR #4007, opened 2026-09-15
docs(guardener): document opt-in smallest-major-bump migration strategy

priyawadhwa:
Documents the opt-in `migrate.version-strategy: smallest-major-bump` migration strategy on the Hardened Actions page, per [ACID-617](https://linear.app/chainguard/issue/ACID-617/xometry-guardener-opt-in-mode-for-smallest-major-version-bump).

## What changed

- New **"Migrate across major versions"** section explaining the default exact-match behavior, why older pins are left untouched, and how to opt into the smallest-major-bump strategy.
- Includes the Xometry `actions/checkout` v3.6.0 → v4.3.1 example and the strategy's conservative guarantees (exact matches win, never downgrades, skips references it can't reason about).
- Added `migrate.version-strategy` to the Configuration reference table.

Wording matches the feature implementation in chainguard-dev/mono#57249.


SharpRake (approved):
lgtm!

PR #4016, opened 2026-09-17
feat(nexus-packages-pull-through): add new docs page for configuring apk repos in Nexus

ribbybibby:

## Type of change

Add a page that describes how to configure Nexus as a proxy for our APK repos.

### What should this PR do?

Provide instructions on how to configure Nexus as a proxy for our APK repos.

We recently fixed our private APK repositories so that its possible to do this now.

We're using a raw proxy here, so we need to introduce some complexity in the setup in order to get the desired caching behaviour. This makes documenting it even more important.

This should describe a working setup accurately and in a way that is easy to follow. I'd appreciate this being tested end to end to make sure it works as expected.

You should be able to run a local Nexus, follow these steps and then execute the 'Testing' steps successfully and install APK packages from the proxy.

```
docker run -d -p 8081:8081 --name nexus sonatype/nexus3
docker exec nexus cat /nexus-data/admin.password
```

Then log in at http://localhost:8081 as 'admin' with that password.


mosabua:
Fyi @tazinprogga @angela-zhang 

And also .. From what I know we are working with Sonatype to get proper apk support added. We should figure out timing for these docs vs the actual feature.

content/chainguard/containers/registry/pull-through-guides/nexus-packages-pull-through/index.md
```diff
@@ -0,0 +1,213 @@
...
+Specific package versions should be immutable and can be cached indefinitely, whereas the `APKINDEX.tar.gz` is updated as new packages are added upstream. To give each artifact type its own caching properties, you'll create two raw proxy repositories (one for the packages, one for the `APKINDEX.tar.gz`) and a pair of routing rules that direct traffic to the correct proxy. Both proxies sit behind a raw group repository. Clients point `apk` at the group's URL.
+
+You can name the repositories and rules whatever suits your organization; the examples below use `chainguard-apk-*` throughout.
+
+### Creating a pull token
+
+The Nexus proxies need to authenticate to Chainguard to fetch packages from a private APK repository. Generate a pull token with `chainctl`:
+
+```shell
+chainctl auth pull-token --repository=apk -o env
```
justinprince:
```suggestion
chainctl auth pull-token --repository=apk
```

justinprince: Adding `-o env` here results in the output not matching what's below.

content/chainguard/containers/registry/pull-through-guides/nexus-packages-pull-through/index.md
```diff
@@ -0,0 +1,213 @@
...
+The repository configuration, package install, and cleanup all happen in a single `RUN` so that the Nexus credentials never end up in an image layer. `--mount=type=secret` mounts the credentials only for the duration of the `RUN`, and the final `rm` removes the `/etc/apk/repositories` file that held them in plain text.
+
+Not every package is available in every private APK repository. For example, your organization may not have access to the `sed` package. See our [private APK repository documentation](/chainguard/containers/building-and-modifying/packages/private-apk-repos/#about-private-apk-repositories) for more details.
+
+Before building, export your Nexus credentials and hostname. If Nexus is running locally, `NEXUS_URL` will be something like `localhost:8081`:
+
+```shell
+export NEXUS_USER=my-nexus-user
+export NEXUS_PASSWORD=my-nexus-password
+export NEXUS_URL=my-nexus-hostname:8081
```
justinprince:
```suggestion
export NEXUS_URL=my-nexus-hostname:8081
export HTTP_AUTH="basic:*:${NEXUS_USER}:${NEXUS_PASSWORD}"
```

content/chainguard/containers/registry/pull-through-guides/nexus-packages-pull-through/index.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+> **Note**: If your Nexus username is an email address, you must percent-encode the `@` sign, as in `export NEXUS_USER=linky%40example.com`.
+
+Then build the image, passing the credentials as Docker build secrets:
+
+```shell
+docker build \
+  --secret id=nexus_user,env=NEXUS_USER \
+  --secret id=nexus_password,env=NEXUS_PASSWORD \
```
justinprince:
```suggestion
  --secret id=http_auth,env=HTTP_AUTH \
```

content/chainguard/containers/registry/pull-through-guides/nexus-packages-pull-through/index.md
```diff
@@ -0,0 +1,213 @@
...
+ARG NEXUS_URL
+RUN --mount=type=secret,id=nexus_user \
+    --mount=type=secret,id=nexus_password \
+    cp /etc/apk/repositories /etc/apk/repositories.disabled && \
+    echo "https://$(cat /run/secrets/nexus_user):$(cat /run/secrets/nexus_password)@${NEXUS_URL}/repository/chainguard-apk" > /etc/apk/repositories && \
+    apk update && \
+    apk add sed && \
+    rm /etc/apk/repositories && \
+    mv /etc/apk/repositories.disabled /etc/apk/repositories
+USER nonroot
```
justinprince:
```suggestion
FROM cgr.dev/chainguard/wolfi-base:latest
USER root
ARG NEXUS_URL
RUN --mount=type=secret,id=http_auth,env=HTTP_AUTH,required=true \
    cp /etc/apk/repositories /etc/apk/repositories.disabled && \
    echo "https://${NEXUS_URL}/repository/chainguard-apk" > /etc/apk/repositories && \
    apk update && \
    apk add sed
USER nonroot
```

justinprince (changes requested):

content/chainguard/containers/registry/pull-through-guides/nexus-packages-pull-through/index.md
```diff
@@ -0,0 +1,210 @@
...
+    echo "https://${NEXUS_URL}/repository/chainguard-apk" > /etc/apk/repositories && \
+    apk update && \
+    apk add sed
+USER nonroot
+EOF
+```
+
+This Dockerfile uses the `python:latest-dev` image. You don't have to use this particular image, but because we're using `apk` to install a package from Nexus, you should use a Chainguard container image that has this package manager available.
+
+The repository configuration, package install, and cleanup all happen in a single `RUN` so that the Nexus credentials never end up in an image layer. `--mount=type=secret` mounts the credentials only for the duration of the `RUN`, and the final `rm` removes the `/etc/apk/repositories` file that held them in plain text.
```
justinprince:
```suggestion
```

justinprince: Removing this section because we're not writing the credentials to the /etc/apk/repositories file due to using the HTTP_AUTH env var.

justinprince (changes requested):

justinprince (approved):

SharpRake:
Thanks for this — I ran the guide end to end against a local Nexus 3.96.3, following it as written for both the private and public repositories. Most of it worked exactly as documented.

I've pushed a single commit (`5e99c6ebf`) with the corrections rather than leaving a pile of inline comments. Happy to split any of it out, or back it out entirely, if you'd rather take them yourself.

## One missing step blocks the guide

`apk update` succeeded, then `apk add` failed:

```
ERROR: sed-4.10-r3: package mentioned in index not found (try 'apk update')
```

The Nexus log gave the reason:

```
Invalid content in repository 'chainguard-apk-packages': GET /x86_64/sed-4.10-r3.apk:
InvalidContentException: Detected content type [application/gzip, ...],
but expected [application/vnd.android.package-archive]
```

Nexus maps the `.apk` extension to Android packages, but an Alpine apk is a gzip archive, so **Strict Content Type Validation** rejects every package it fetches. The box is checked by default on a fresh **Create Repository: raw (proxy)** page, so every reader hits this. Clearing it on the packages proxy fixed the build immediately.

Only that one proxy needs it. The index proxy is unaffected, since `.tar.gz` implies the type the file actually is. I also checked whether the group mattered by re-enabling validation there and rebuilding — it still passed, so the group setting is irrelevant.

## The first Debugging bullet points at the wrong setting

It attributes `package mentioned in index not found` to **Preserve encoded characters in URLs** being off. That setting was enabled on both proxies for the entire failing run, so it wasn't the cause here.

I've split the bullet in two, ordered by likelihood: Strict Content Type Validation first, with the log line so a reader can confirm the diagnosis rather than guess, then your R2 signature explanation second, gated on "if that setting is already off." I didn't verify the SigV4 mechanism myself, so I left your wording for it intact.

## The apk-tools auth claim isn't correct

> `apk-tools` doesn't send HTTP basic auth against a repository URL without embedded credentials.

apk-tools 2.14.10 supports `HTTP_AUTH`, and the guide's own test step depends on that — every build I ran authenticated with `basic:*:user:pass` and no credentials in any URL. The bullet also pointed at embedding credentials "as described in the testing section," which the testing section doesn't describe. Rewritten to lead with `HTTP_AUTH`, and to note that Nexus still wants its own credentials for the public repositories even though their upstreams don't.

## The test step can't work against a local Nexus

The Dockerfile hardcoded `https://${NEXUS_URL}` while the text below it said `NEXUS_URL` would be something like `localhost:8081`. A stock `docker run sonatype/nexus3` serves plain HTTP only — so the exact setup your PR description gives reviewers fails here. (Separately, `localhost` inside a `docker build` is the build container, not the host; I used the docker0 gateway.)

I moved the scheme into `NEXUS_URL`, which also fixes the public snippet for free. Those lines were already written as `"${NEXUS_URL}/repository/..."` with no scheme, so apk warned and skipped them:

```
WARNING: opening 172.17.0.1:8081/repository/chainguard-apk-public: No such file or directory
```

Because the guide suggests adding the public repos *alongside* the private one, the build still succeeded from the private repo — so the misconfiguration was invisible unless you read the warnings.

## Smaller changes

- **Group member order doesn't matter.** The routing rules make the two proxies mutually exclusive, so precedence never decides anything. Verified by reversing the members and rebuilding. The old wording ("add `chainguard-apk-packages` first, then...") implied it did matter.
- Four internal links pointed at aliases rather than canonical paths.
- The second note now uses the `{{< note >}}` shortcode, matching the first.
- Editorial pass: passive voice, future tense, `e.g.`, placeholder style, punctuation.

## Confirmed working — no changes needed

- `HTTP_AUTH="basic:*:..."` wildcard-host syntax
- Reusing one BLOCK and one ALLOW routing rule across all three repository sets
- `chainctl iam organizations list -o table` for the org UID, and both `virtualapk.cgr.dev` URLs
- The public setup end to end — installed `cuda-stub-13.3`, which exists only in `extra-packages`, so it proves that group specifically

## Two for you to decide

- I replaced "At the time of writing (2026-09-17)" with "As of Nexus 3.96" so the note ages against a version rather than a date. 3.96.3 is what I tested, but I'm inferring that's the version you wrote against — correct it if not.
- "Specific package versions should be immutable" — the `-1` cache age rests on this. If they *are* immutable, it's worth saying so outright.

One caveat on method: I created the six public repositories through the REST API rather than the GUI, so that section's settings and outcomes are verified but its click path and field labels aren't.

SharpRake (approved):

matthewhelmke:
> @ribbybibby is attempting to deploy a commit to the **Chainguard** Team on [Vercel](https://vercel.com).
> 
> A member of the Team first needs to [authorize it]...

This is new. I'm creating a new preview workflow for a number of reasons. It will operate in parallel with the Netlify previews for a short time while I get it configured how we want. Yours is the first PR to trigger it! (Now I have to go look and see if it worked...)

matthewhelmke (approved):
LGTM, thank you! This has been on the wanted list for a long time!

SharpRake (approved):

